PR #8, opened 2020-09-14
Read log frames over itm

steveklabnik:
This patch changes the itm command to intepret log frames coming from
hubris. The frames look like this:

* u8, u8: 0x01, 0xde: magic number
* u16: task id, in big endian format
* u16: frame length, in big endian format
* defmt frame

The information needed to interpret the defmt frame is stored in each
task's ELF, so we first read those all in, and save them in a hashmap.
Then, when a message comes in, we read the above, look up the information
using the task id, and then interpret the frame accordingly.

------------------

I haven't opened a PR on the hubris side yet, and I'm not sure if we want to put this in the ITM command, or create a new specific one. I am open to any and all feedback on this PR.

- [x] make a separate subtask


steveklabnik:
Closing this one out until I come back to logging in the future.

PR #41, opened 2021-11-29
Implement the `vsc7448` subcommand

mkeeter:
This PR implements the `humility vsc7448` subcommand, which lets a Hubris system talk over SPI to a VSC7448 ethernet switch IC.

It also includes a handful of drive-by changes:
- Adds `Archive::Ignored` and refactors some code to avoid loading a Hubris archive if it isn't needed, saving 1-2 seconds of startup time.
- Moves get-a-function-from-HIF into a standalone function in a new `struct HiffyFunctions`, to get rid of some duplicated code.
- Adds `HiffyContext::execute_blocking`, which is just `execute` + wait until the core is done + return the result.
- Move which-task-is-SPI into a standalone helper function, since it's now used in both `humility spi` and `humility vsc7448`.

### DIFF
--- a/src/cmd.rs
+++ b/src/cmd.rs
@@ -2,13 +2,11 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-mod vsc7448;
-
+use anyhow::{bail, Context, Result};
 use humility::hubris::*;
 use humility_cmd::Args;
 use humility_cmd::{attach_dump, attach_live};
 use humility_cmd::{Archive, Attach, Command, Validate};
-use anyhow::{bail, Context, Result};
 use std::collections::HashMap;
 use structopt::clap::App;
 
@@ -46,7 +44,7 @@ pub fn init<'a, 'b>(
         cmd_test::init,
         cmd_trace::init,
         cmd_stmsecure::init,
-        vsc7448::init,
+        cmd_vsc7448::init,
     ];
 
     for dcmd in &dcmds {
--- a/src/cmd.rs
+++ b/src/cmd.rs
@@ -4,11 +4,11 @@
 
 mod vsc7448;
 
-use anyhow::{bail, Result};
 use humility::hubris::*;
 use humility_cmd::Args;
 use humility_cmd::{attach_dump, attach_live};
 use humility_cmd::{Archive, Attach, Command, Validate};
+use anyhow::{bail, Context, Result};
 use std::collections::HashMap;
 use structopt::clap::App;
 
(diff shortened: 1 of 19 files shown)


src/cmd.rs
```diff
@@ -28,22 +28,24 @@ mod stmsecure;
...
 use structopt::clap::App;
 
 #[allow(dead_code)]
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub enum Archive {
     Required,
     Optional,
     Prohibited,
+    Ignored,
```
bcantrill, line 48: It feels like `Ignored` doesn't have materially different semantics as `Optional`?

src/hiffy.rs
```diff
@@ -347,6 +368,20 @@ impl<'a> HiffyContext<'a> {
         Ok(())
     }
 
+    /// Blocking execution of a program, returning the results
+    pub fn execute_blocking(
```
bcantrill, line 372: This looks really helpful to refactor it in this way -- in the spirit of less typing (and the really important stuff: fewer line breaks), what do you think about calling it `run` instead?

src/spi.rs
```diff
@@ -0,0 +1,63 @@
+/*
```
bcantrill, line 1: Note that you'll want to make this the MPL header before merging

bcantrill (approved):
This looks good!  One question and one nomenclature suggestion (`execute_blocking` to `run`) and one reminder to delete the copyright header and replace with the MPL header.  The `vsc7448` subcommand itself looks really useful!

src/cmd.rs
mkeeter: It's subtly different: `Optional` will load a Hubris archive if one is available (which is a 1-2 second startup penalty), vs `Ignored` which will not.

This means I can keep `HUMILITY_ARCHIVE` defined but still have the `info` subcommand be faster.

(It looks like `Prohibited` isn't actually used _anywhere_ – want me to take it out while I'm in here?)

src/hiffy.rs
mkeeter: Done!  How do you feel about renaming `execute` (which kicks off the program but doesn't block) to `start`, to avoid ambiguity with both `execute` and `run` functions?

src/cmd.rs
bcantrill: Ah, you use the environment variable! Yes, that makes sense.  It feels safe to ditch `Prohibited` -- and maybe worth adding a comment on each member as to the difference in semantics?

src/hiffy.rs
bcantrill: Yes, that sounds good!

src/spi.rs
mkeeter: Done in the Great Rebasing!

src/cmd.rs
mkeeter: Okay, done!

src/hiffy.rs
mkeeter: Done!

PR #43, opened 2021-11-30
fix typo in README.md

ahrens:
fix typo: prented -> printed

Incidentally, I also had trouble understanding this sentence in `humility map`, but I wasn't sure how to fix it, perhaps due to my unfamiliarity with embedded systems:

> This can happen, for example, because the task attempted to access device memory that was not allocated to it in the build description, because it exceeded the device memory allocated to it, or because it allocated the memory allocated to it, e.g., because of a stack overflow.

Specifically, "because it allocated the memory allocated to it" doesn't sound like a problem.  Should this be something like, "because it allocated *past* the memory allocated to it"?

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,7 +517,7 @@ humility: reading at 0x11d00 for 100 bytes
 ```
 
 Both arguments can be in either hex, decimal, octal or binary (addresses
-and contents will always be prented in hex):
+and contents will always be printed in hex):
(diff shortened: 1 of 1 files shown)


kc8apf:
Filed #45 for the sentence regarding `humility map` so it doesn't get lost when this PR is merged.

PR #51, opened 2021-12-07
Add basic GitHub Action

steveklabnik:
This is a very basic Actions setup. I figure we can start here and iterate.

Latest questions here: https://github.com/oxidecomputer/humility/pull/51#issuecomment-989212627

~~@bcantrill some comments/thoughts:~~

~~* I left in a `cargo test` because even though we don't have any tests, nice to have it there for when we do. If you don't plan on any testing any time soon, could remove it.~~
~~* rustfmt uses nightly options here, I don't know if you want to pin to a nightly or just use "nightly" generally.~~
~~* speaking of which it seems like humility is currently not rustfmt clean, happy to push a commit here that fixes that, but i wonder if it's a nightly divergence~~
~~* this version thing also goes for the main build; i just left it on "stable" but if you want something other than that, happy to tweak.~~
~~* I left a clippy action in here too just for kicks, if you don't want this, happy to remove~~

~~This does *not* currently include the license check from Hubris. Extracting that from hubris's build system will take some more work. Happy to do that as part of this PR or as a follow-up.~~

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,60 @@
+on: [push, pull_request]
+
+name: Continuous integration
+
+jobs:
+  check:
+    name: Check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Dependencies
+        run: sudo apt-get update && sudo apt-get install libusb-1.0-0-dev libftdi1-dev
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: 1.57.0
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: check
+
+  test:
+    name: Test Suite
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Dependencies
+        run: sudo apt-get update && sudo apt-get install libusb-1.0-0-dev libftdi1-dev
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: 1.57.0
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+
+  clippy:
+    name: Clippy
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Dependencies
+        run: sudo apt-get update && sudo apt-get install libusb-1.0-0-dev libftdi1-dev
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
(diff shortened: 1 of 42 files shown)


steveklabnik:
* rustfmt is failing on pmbus and spd
* I want to add some more sophisticated build caching but until we answer the above questions, wanted to keep it simple and easy to review.

steveklabnik:
A toolchain file is being added in https://github.com/oxidecomputer/humility/pull/54

steveklabnik:
https://github.com/oxidecomputer/humility/pull/56 moves the submodules to git deps, which means we *should* be rustfmt clean after that is merged.

steveklabnik:
I have pushed a commit that gets clippy clean. I am not saying that you *should* enforce clippy in CI, or take that patch, but I figured that having access to that would make said decision easier or harder. I do think there's at least one bit of this that you should take regardless, and that's https://github.com/oxidecomputer/humility/pull/51/commits/1075ae73274973c493d65fc22c55351112cbb764#diff-6c2cd2ddc6195c23d24ae83f43a64f5822500f7fcac1c17cef6f82e4b3c7dd12L473-R473, this is a real bug. The rest is, I think overall good but also neutral and it's really up to you. I also made some small judgement calls in how to handle some of them; I bet that the string format literal stuff was on purpose and so I tried to suppress the warning rather than "fix" it the way clippy may have wanted to.

steveklabnik:
Okay, this is green! Some final decisions before I merge:

- [ ] Do we want to bother with `cargo test`?
- [ ] Formatting is using the `nightlly` toolchain, is that okay or do you want a specific nightly?
- [ ] Do you want clippy and the associated commit, or not?
- [ ] Do you want license checking in here or not? I think it should exist but it's more work and I'd like to land this first, but if you feel differently I am open to it.

I want to add some more sophisticated build caching but until we answer the above questions, wanted to keep it simple and easy to review.

bcantrill:
> Okay, this is green! Some final decisions before I merge:
> 
> * [ ]  Do we want to bother with `cargo test`?
> * [ ]  Formatting is using the `nightly` toolchain, is that okay or do you want a specific nightly?
> * [ ]  Do you want clippy and the associated commit, or not?
> * [ ]  Do you want license checking in here or not? I think it should exist but it's more work and I'd like to land this first, but if you feel differently I am open to it.
> 
> I want to add some more sophisticated build caching but until we answer the above questions, wanted to keep it simple and easy to review.

Yes on `cargo test`; `nightly` is good for rustfmt; yes on `cliippy`; license checking up to you, but agreed about landing this first! Thanks for doing this!

steveklabnik:
Filed https://github.com/oxidecomputer/humility/issues/58 to track the licensing thing. This is now good to go! Merging!

PR #54, opened 2021-12-07
Add rust-toolchain file.

### LINKED ISSUE #53: humility needs a rust-toolchain.toml
Depending on the date of your installation, building humility will fail with something like:

```
error: failed to load manifest for workspace member `/home/rm/oxide/src/humility/humility-core`

Caused by:
  failed to parse manifest at `/home/rm/oxide/src/humility/humility-core/Cargo.toml`

Caused by:
  feature `edition2021` is required

  consider adding `cargo-features = ["edition2021"]` to the manifest
```

While the solution for folks is to utter a `rustup update`, it seems that like hubris, we should have a `rust-toolchain.toml` that describes what toolchain is actually needed for the build.

cbiffle:
Because we've started using the 2021 Edition in some crates, peoples'
build will mysteriously fail with an older toolchain.

Fixes #53.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "1.57"


bcantrill (approved):
Thank you!

PR #56, opened 2021-12-08
Remove submodules

steveklabnik:
We had these as submodules when we were a private repository. Now that
we're all open, this is not needed, and causes way more headaches than a
git dependency would.


steveklabnik:
I am gonna make an executive decision to merge this because Bryan agreed to it in chat and I'm sure has better things to do with his time, and it's blocking the rest of my work here. Bryan if you change your mind happy to revert this in the future.

PR #61, opened 2021-12-13
redo formatting job

steveklabnik:
Let's try this.


steveklabnik:
Okay, this is at *least* not a regression:

image

steveklabnik:
There is a small amount of weirdness here: because it pushed a commit, and that commit doesn't trigger another workflow run (good), it shows there being no actions because the HEAD didn't run any. I cannot remember if this happened with our older code too or not.

steveklabnik:
I'm going to merge this and see if it works with a new PR, if it doesn't, we can always revert this stuff.

PR #62, opened 2021-12-13
formatting error, deliberately introduced

steveklabnik:
Testing https://github.com/oxidecomputer/humility/pull/61 by seeing if it Does The Right Thing from a fork.


steveklabnik:
Sadly, it does not: https://github.com/oxidecomputer/humility/runs/4510133672?check_suite_focus=true#step:5:17

Issue #14, opened 2021-10-08
should allow probe to be specified as VID:PID


bcantrill:

bcantrill:
Merged in #65.

PR #65, opened 2021-12-15
should allow probe to be specified as VID:PID (#14)

bcantrill:


README.md
```diff
@@ -108,6 +108,16 @@ the `-p` option (long form `--probe`), which can have the following values:
...
   examine the serial number in the output of `humility probe`.
 
+- `vid:pid[:serial]`: In some cases, the automatic algorithm may either find
+  the wrong thing, or timeout attempting to search for non-existent probes.
+  In these cases, it can help to explicitly set the vendor ID (VID) and
+  product ID (PID) of the debug probe, which should be colon-delimited, e.g.,
+  `0483:374e`.  (Determining the VID and PID of an attached probe is
+  platform-specific; on Linux one can use the `lsusb` command.)  In cases
+  where there are multiple probes with the same VID and PID, the serial number
+  of the probe (as reported via `humility probe` can be postpended, also
```
rmustacc, line 118: I'd like to somehow indicate that this serial number is also the USB device serial number. Maybe something like 'the USB Device Serial Number (as reported by `humilty probe` or platform-specific tools like `libusb`)'. I dunno. Maybe you can phrase that in a way that sounds better. There's also a missing closing paren ')' after humility probe in this line.

rmustacc (commented):
Thanks for doing this! With this in place a lot more is working for me with humility much more reliably.

humility-core/src/core.rs
```diff
@@ -1009,12 +1009,7 @@ pub fn attach(mut probe: &str, chip: &str) -> Result<Box<dyn Core>> {
...
                 let product_id = selector.product_id;
-
-                #[allow(clippy::manual_map)]
-                let serial_number = match selector.serial_number {
-                    Some(ref serial_number) => Some(serial_number.clone()),
-                    None => None,
-                };
+                let serial_number = selector.serial_number.clone();
 
                 let res = probe_rs::Probe::open(selector);
```
cbiffle, line 1014: `let probe = probe_rs::Probe::open(selector)?;`

bcantrill: Of course, this has now left me with the question of why I didn't write it that way the first time... :|  In any case, fixed -- thanks!

README.md
bcantrill: Fixed -- thank you!

PR #66, opened 2021-12-16
Search harder for TICKS, CLOCK_FREQ_KHZ

theotherjimmy:
In my builds of hubris, the `TICKS` and `CLOCK_FREQ_KHZ` variables were both not included in `HubrisArchive`'s `variables` member. Instead, they showed up in the `esyms` and `esyms_byname` members. So I made humility search a bit harder for them by also looking in the `esyms_byname` member. There may be more of these that I have yet to find, and this makes humulity's trace and tasks subcommands work for me.


theotherjimmy:
Turns out this breaks the jefe subcommand. In particular the first commit is incorrect.

theotherjimmy:
Huh, there was something wrong with my build. It all works now.

PR #73, opened 2022-01-06
Check license headers in a GitHub Action

steveklabnik:
This is in contrast to https://github.com/oxidecomputer/hubris/pull/270 and https://github.com/oxidecomputer/hubris/pull/279, where I implemented a script in Rust to check for it.

This GitHub Action is used by the Apache Software Foundation to ensure compliance in all of their projects, so it seems much better to rely on that rather than port over my little thing. Will probably also move Hubris over if everyone prefers.

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -1,0 +1,12 @@
+header:
+  license:
+    spdx-id: MPL-2.0
+
+    content: |
+      This Source Code Form is subject to the terms of the Mozilla Public
+      License, v. 2.0. If a copy of the MPL was not distributed with this
+      file, You can obtain one at https://mozilla.org/MPL/2.0/.
+  paths:
+    - '**/*.rs'
+
+  comment: on-failure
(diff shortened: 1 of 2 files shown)


steveklabnik:
Nice: https://github.com/oxidecomputer/humility/runs/4730455799?check_suite_focus=true#step:4:515

.licenserc.yaml
```diff
@@ -0,0 +1,12 @@
+header:
+  license:
+    spdx-id: MPL-2.0
+
+    content: |
+      This Source Code Form is subject to the terms of the Mozilla Public
+      License, v. 2.0. If a copy of the MPL was not distributed with this
+      file, You can obtain one at https://mozilla.org/MPL/2.0/.
```
steveklabnik, line 8: This step is necessary because the upstream workflow only knows the header text of the Apache license. I *could* eliminate this by sending them a PR, but haven't done so yet (they claim they welcome such things, so I suspect it would be well received, but four lines of YAML seems like the right call for this moment, rather than waiting on all of that)

steveklabnik:
I am going to merge this one, and if anyone has any issues, please let me know and I'm happy to change things.

PR #75, opened 2022-01-09
Make an error message include the likely solution to the problem in it.

jeffallen:


bcantrill:
This looks good; I think I might back the message off a little bit to ask the question (e.g. "have you run \"cargo install cargo-readme\"?").  Thanks for fixing this!

PR #78, opened 2022-01-14
Fix git deps to work on latest

steveklabnik:
Both the spd and idt8a3xxxx crates have some incompatible changes that
have been made upstream. This normally wouldn't matter, as we have a
Cargo.lock, but the default behavior of `cargo install` doesn't respect
the lockfile, causing confusion if you forget `--locked`.

For spd, this was changed here:
https://github.com/oxidecomputer/spd/commit/61d4e5dd9c4a0c77c17140bbcbe2c94b8dfe00b0#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L107

For idt8a3xxxx, this was changed here:
https://github.com/oxidecomputer/idt8a3xxxx/commit/24284e14a2357b763dbaf90c7573ea0046f34f6f#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L37

I am confident in the spd change, but less confident in the idt8a3xxxx change. Looking at the diff, it's the same value, so I'm confident that it doesn't *break*, but I am less confident it's the correct choice. I *think* it is but am not 100% sure, so would ask for a careful review there.

Oh: also to reduce this confusion, we *could* select the specific hash and put it in `Cargo.toml`. This means that the operation that becomes confusing in the future is `cargo update` rather than `cargo install`, which may be a good tradeoff. Right now, `cargo update` will update to `HEAD` of these repos, whereas in that world, nothing would change. This would make `cargo install` Just Work but may lead to confusion when trying to update these deps later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.16.0"
+version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e61f2b7f93d2c7d2b08263acaa4a363b3e276806c68af6134c44f523bf1aacd"
+checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
 dependencies = [
- "gimli 0.25.0",
+ "gimli 0.26.1",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


bcantrill (approved):
Yes, thank you for doing this -- and my apologies!

PR #79, opened 2022-01-14
buildomat: add basic build/test job

jclulow:


jclulow:
This buildomat job is modelled on what I interpreted the GitHub actions configuration as doing, with respect to running tests via Cargo.

PR #84, opened 2022-01-26
Move from structopt to clap v3

steveklabnik:
Clap v3 is finally here! This basically makes structopt obsolete, as
clap has now absorbed its main feature.

Most of this is very mechanical. Some names were changed, a lifetime was
removed, repeat a few dozen times.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,15 +33,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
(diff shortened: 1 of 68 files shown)


steveklabnik:
One small change: I think this adds color by default? 
image

i like it, but I could understand if you want that turned off by default, happy to figure it out.

bcantrill (approved):
Looks great -- thank you!

PR #89, opened 2022-02-01
Modernize logging

steveklabnik:
This is pretty standard, expected, and @cbiffle asked nicely :)

### DIFF
--- a/cmd/ringbuf/src/lib.rs
+++ b/cmd/ringbuf/src/lib.rs
@@ -40,9 +40,6 @@ use humility_cmd::doppel::{Ringbuf, StaticCell};
 use humility_cmd::reflect::{self, Format, Load, Value};
 use humility_cmd::{Archive, Args, Attach, Command, Validate};
 
-#[macro_use]
-extern crate log;
-
 #[derive(Parser, Debug)]
 #[clap(name = "ringbuf", about = env!("CARGO_PKG_DESCRIPTION"))]
 struct RingbufArgs {
@@ -119,6 +116,9 @@ fn taskname<'a>(
     Ok(&hubris.lookup_module(HubrisTask::from(variable.goff))?.name)
 }
 
+// this allow is meant for the header println! in the body but you cannot apply
+// an attribute to a macro invoction, so we have to put it here instead.
+#[allow(clippy::print_literal)]
 fn ringbuf(
     hubris: &HubrisArchive,
     core: &mut dyn Core,
@@ -150,11 +150,12 @@ fn ringbuf(
     ringbufs.sort();
 
     if subargs.list {
-        info!("{:18} {:<30} {:<10} {}", "MODULE", "BUFFER", "ADDR", "SIZE");
+        println!("{:18} {:<30} {:<10} {}", "MODULE", "BUFFER", "ADDR", "SIZE");
 
         for v in ringbufs {
             let t = taskname(hubris, v.1)?;
-            info!("{:18} {:<30} 0x{:08x} {:<}", t, v.0, v.1.addr, v.1.size);
+
+            println!("{:18} {:<30} 0x{:08x} {:<}", t, v.0, v.1.addr, v.1.size);
         }
 
         return Ok(());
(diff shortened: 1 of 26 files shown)


steveklabnik:
> What's the behavior of this with respect to the -v option (which enables logging)? 

I completely forgot that you've implemented your own logging backend, hilariously to validate this works I added a log statement, tested, but that was *before* the code that initializes your logger, so this actually will not work as is.

I will work up a better patch.

steveklabnik:
Okay, I have updated this. It now:

1. removes the de novo log backend in favor of env_logger
2. modernizes invocation of logging macros; #[macro_use] is no longer
   neccesary and prefixing them with log:: makes things a bit more clear
3. removes the fatal! macro from the main binary in favor of just doing
   its job at the few callsites it was used in

I am happy to write up documentation here but it's not 100% sure to me where exactly it should go. Any thoughts?

cmd/ringbuf/src/lib.rs
```diff
@@ -150,11 +147,14 @@ fn ringbuf(
     ringbufs.sort();
 
     if subargs.list {
-        info!("{:18} {:<30} {:<10} {}", "MODULE", "BUFFER", "ADDR", "SIZE");
+        #[rustfmt::skip]
```
steveklabnik, line 150: this line and 156 both annoy me, however, i could not get `#[rustfmt::skip::macros]` to work like we do with `println` elsewhere, so resorted to this. These are *just* over 80 chars and so get splatted into five lines without it, which I figured you would not like, but if you don't mind, could remove these two.

bcantrill (commented):
So,  this results in much noisier output in normal Humility operation:

```console
[2022-02-03T18:02:12Z INFO  humility_core::core] attached via ST-Link
system time = 2752649
ID TASK                 GEN PRI STATE    
 0 jefe                   0   0 recv, notif: bit0 bit1(T+51)
 1 rcc_driver             0   1 recv
 2 gpio_driver            0   2 recv
 3 spi4_driver            0   2 recv
 4 spi2_driver            0   2 recv
 5 i2c_driver             0   2 recv
 6 spd                    0   2 notif: bit0
 7 thermal                0   3 recv, notif: bit0(T+464)
 8 power                  0   3 notif: bit0(T+328)
 9 hiffy                  0   3 notif: bit0(T+101)
10 gimlet_seq             0   3 recv
11 hf                     0   3 recv
12 sensor                 0   3 recv, notif: bit0(T+351)
13 idle                   0   5 RUNNING
```

The use of logging for these kind of messages is historical, and I think we want to change Humility's use of `info!` to be a macro that emits to stderr with the command name prefixed (i.e., preserving the existing output).  Perhaps a `msg!` or similar?

steveklabnik:
@bcantrill updated as per our convo; give this a quick try if you have a moment. I want to do one or two little things (like documenting the msg macro) before merge, but let's see if this makes the output better for you

steveklabnik:
```
warning: unused attribute `allow`
   --> cmd\ringbuf\src\lib.rs:150:9
    |
150 |         #[allow(clippy::print_literal)]
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_attributes)]` on by default
note: the built-in attribute `allow` will be ignored, since it's applied to the macro invocation `println`
   --> cmd\ringbuf\src\lib.rs:151:9
    |
151 |         println!("{:18} {:<30} {:<10} {}", "MODULE", "BUFFER", "ADDR", "SIZE");
    |         ^^^^^^^

```

sigh...

cmd/ringbuf/src/lib.rs
```diff
@@ -119,6 +116,8 @@ fn taskname<'a>(
     Ok(&hubris.lookup_module(HubrisTask::from(variable.goff))?.name)
 }
 
+// this allow is meant for line 151 but it does not work, so we do it here
```
bcantrill, line 119: Maybe change this to indicate that it's a `println!` in the body rather than the specific line?  This comment will become *really* confusing if the file gets reordered. ;)

bcantrill (approved):
One small nit, otherwise looks great -- thank you!

Issue #90, opened 2022-02-03
Conflict with `jefe -h` and `-h`


steveklabnik:
A change from clap v2 vs v3 will cause a panic when doing `jefe -h`:

```
     Running `target\debug\humility.exe -a ..\hubris\target\demo-stm32f4-discovery\dist\build-demo-stm32f4-discovery.zip jefe -h`
thread 'main' panicked at 'App jefe: Short option names must be unique for each argument, but '-h' is in use by both 'hold' and 'help'', C:\Users\steve\.cargo\registry\src\github.com-1ecc6299db9ec823\clap-3.0.12\src\build\app\debug_asserts.rs:100:17
```

Reported by @lzrd 

As a short fix, @bcantrill is okay migrating the non-help arg to `-H`.

PR #92, opened 2022-02-03
Validate clap

### LINKED ISSUE #91: Can clap's configuration be asserted statically?
In #90 we didn't discover that there was an issue until runtime. Given that this always panics, it would be nice to figure out if there is a way to either:

1. get a compile-time failure
2. assert the failure after construction so that every command fails if something is wrong

steveklabnik:
~~This will eventually fix #91, but it's going to be failing until I work through all of the current failures.~~

This introduces a test that will assert that the state built up within
clap is coherent.

Fixes https://github.com/oxidecomputer/humility/issues/91

Most of these removals or changes are simply fixes, but we also change
`jefe -h` to `jefe -H`

Fixes https://github.com/oxidecomputer/humility/issues/90

### DIFF
--- a/cmd/i2c/src/lib.rs
+++ b/cmd/i2c/src/lib.rs
@@ -190,7 +190,7 @@ pub struct I2cArgs {
     /// flash the specified file, assuming two byte addressing
     #[clap(long, short,
         conflicts_with_all = &[
-            "write", "raw", "nbytes", "read", "writeall", "register", "scan",
+            "write", "raw", "nbytes", "register", "scan",
             "writeraw"
         ],
         value_name = "filename",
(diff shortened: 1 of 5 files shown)


steveklabnik:
I don't understand this latest failure because it's right there: https://github.com/oxidecomputer/humility/blob/master/cmd/spi/src/lib.rs#L57

steveklabnik:
Thanks to clap's `debug` feature:

```
[clap::build::arg::debug_asserts]       Arg::_debug_asserts:bigendian-address
```

you give the name of the flag (`bigendian-address`) not the name of the struct member (`bigendian_address`). nice.

bcantrill (approved):
This looks great -- sorry about the delay!

steveklabnik:
To be honest I thought it was already merged, that's my bad!

PR #95, opened 2022-02-08
add nightly build

steveklabnik:
We want to check that hubris doesn't regress certain things in humility, and so the best way to do that is to start producing a nightly build.

Every day, this will run this job. This job will:

* build humility in release
* upload the binary to https://github.com/oxidecomputer/humility/releases/tag/nightly

It will overwrite the one that was previously there, so we don't keep every single nightly around.

This workflow *may* break if the dependencies needed to build changes, in theory I will get a notification if this fails, but it's worth remembering.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,31 @@
+name: 'nightly'
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


bcantrill (approved):
Looks great -- THANK YOU!

PR #97, opened 2022-02-11
Add spdm command

lzrd:


andrewjstone (approved):

andrewjstone:
We aren't using SPDM any longer

PR #101, opened 2022-02-23
Add support for idol Reply::Simple

labbott:
Commit 7818dbdc73f6e52d2d29297113bb8d29d23479e9 in idolitry added
Reply::Simple to indicate functions that do not return error codes.
Add support into humility for this.

This also includes clippy deprecation warnings because apparently this triggered a version bump(????)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,18 +34,18 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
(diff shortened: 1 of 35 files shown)


bcantrill (approved):
Looks good -- thanks for doing this, and sorry that it ended up expanding so much!

PR #109, opened 2022-03-14
Add 'rng' command.

flihp:
This relies on a new function in the hiffy task. We invoke this function
causing the task to return --block-size bytes to us. The new command takes
parameters similar to 'dd':

-a, --ascii
        base64 encode / ascii armor output

-b, --block-size <BLOCK_SIZE>
        size of each block read from RNG [default: 32]

-c, --count <COUNT>
        number of blocks read from RNG, if omitted bytes will be read until
        interrupted by the user

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -850,6 +850,7 @@ dependencies = [
  "humility-cmd-renbb",
  "humility-cmd-rencm",
  "humility-cmd-ringbuf",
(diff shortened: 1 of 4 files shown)


flihp:
this command requires commits from https://github.com/oxidecomputer/hubris/pull/448 to do something meaningful

PR #115, opened 2022-03-21
Flash using probe-rs

labbott:
probe-rs has native flashing support now. Use that instead of
calling out to OpenOCD/pyOCD.


mkeeter (approved):
Tested and confirmed this works on my machine!

PR #119, opened 2022-04-03
prevents an overflowing subtraction when using JLink

benbrittain:
When using `JLink V762c` on Linux with a Cortex-M4, I got a 

```
thread 'main' panicked at 'attempt to subtract with overflow', humility-core/src/core.rs:597:56
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: humility failed
```

when issuing a `cargo xtask humility <app> tasks` command

This looks like it's caused by a spurious (or perhaps overlapping?) ack.

```
[2022-04-03T23:22:37Z TRACE humility_core::core] sending $m200022a8,20#ba
[2022-04-03T23:22:37Z TRACE humility_core::core] response: +
[2022-04-03T23:22:37Z TRACE humility_core::core] response: +$382300201000000001000000602300209cff0000a9830100f887010000000041#62
```

Checking that the responding message is larger or equal to the expected GDB packet fixes the problem.

```
humility: attached via JLink
system time = 352367
ID TASK                 GEN PRI STATE    
 0 idle                   0   5 RUNNING
 1 jefe                   0   0 recv, notif: bit0 bit1(T+33)
 2 hiffy                  0   3 notif: bit0(T+133)
 3 gpio                   0   2 recv
 4 blinky                 0   3 recv, notif: bit0(T+133)
 ```

### CODE REVIEW

File: humility-core/src/hubris.rs [hunk 1]
```diff
@@ -678,7 +678,8 @@ impl HubrisArchive {
             let file = match header.file(file) {
                 Some(header) => header,
                 None => {
-                    bail!("no header at {}", goff);
+                    //bail!("no header at {}", goff);
```
kc8apf, line 681: This seems unrelated to the PR description.

cmd/flash/src/lib.rs
```diff
@@ -305,6 +306,10 @@ fn flashcmd(
                             c = Some("STM32G030C6Tx".to_string());
                             break;
                         }
+                        if nr.is_match(s) {
```
kc8apf, line 309: Unrelated to the PR description.

benbrittain: Sorry, those snuck in because I committed from the master branch, classic github mistake.

benbrittain:
Removed the accidental code due to a makeing a pull request from master branch & rebased. Confirmed it still works as intended

PR #124, opened 2022-05-04
Add command for controling the SP via RoT

labbott:


cmd/spctrl/src/lib.rs
```diff
@@ -0,0 +1,132 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! ## `humility spctrl`
+//!
+//! `humility spctrl` runs commands from the RoT to control the SP
```
bcantrill, line 7: I haven't always been a model in this regard, but trying to make sure that new commands come in with some documentation. Might be worth spelling out the commands a bit here? See, e.g. [the docs for "humility i2c"](https://github.com/oxidecomputer/humility/blob/cf6c33cb5327386e8ed19dc9de5dcfb1b31ebcfa/cmd/i2c/src/lib.rs#L5-L92).

bcantrill (approved):
Looks great! Only comment is to flesh our the docs a tad -- and then run `cargo xtask readme` and include the resulting change to `README.md`.

PR #126, opened 2022-05-08
Stop relying on ihex/srec files in build archives.

cbiffle:
Previously we would use three different representations of the program
in a build archive:

- ELF for most things.
- SREC only for OpenOCD
- IHEX only for PyOCD

The build archive stored all three, adding a _lot_ of size to every
build archive. This commit continues using SREC and IHEX but generates
them on the fly from the ELF file in the archive, so that we can later
land a corresponding Hubris change to stop including SREC/IHEX in every
archive.

### DIFF
--- a/cmd/flash/src/lib.rs
+++ b/cmd/flash/src/lib.rs
@@ -155,7 +155,7 @@ fn flashcmd(
                 bail!("unexpected OpenOCD payload: {:?}", payload);
             }
 
-            std::fs::write(&srec, flash_config.srec)?;
+            std::fs::write(&srec, generate_srec_from_elf(&flash_config.elf)?)?;
 
             //
             // OpenOCD only deals with slash paths, not native paths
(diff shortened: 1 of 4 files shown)


cmd/flash/src/lib.rs
```diff
@@ -283,3 +283,83 @@ pub fn init() -> (Command, ClapCommand<'static>) {
...
 }
+
+/// While it may sound like the impetus for an OSHA investigation at the North
+/// Pole, this function is _actually_ designed to generate small (32-byte)
+/// chunks describing the data in the PHDRs of an ELF file. Unless the file is
+/// missing PHDRs, because objcopy sometimes does that for whatever reason, in
+/// which case we do the section headers.
+///
+/// This is an implementation factor of both SREC and IHEX generation.
+fn elf_chunks(elf_data: &[u8]) -> Result<Vec<(u32, &[u8])>> {
```
bcantrill, line 294: 🤣

bcantrill (approved):
Looks great -- presumably you have tested this by verifying that the contents correctly match the old contents?  And is it worth adding a test for that that can operate on our non-slimmed core files? (It may well not be...)

cbiffle:
I've inspected the files and confirmed that the contents match (and fixed a series of bugs where they didn't).

In your current framework I'm not sure how to test the flash command.

PR #145, opened 2022-05-17
Accept `chip.toml` as peripherals filename

mkeeter:
This is Step 0 in my grand plan to move the `gdb` subcommand from Hubris to Humility, so you can easily run GDB against an archive.

As part of this work, I'm reorganizing more files in Hubris to be in per-chip folders; this means the peripherals file is now named `chip.toml` instead of a chip-specific name.  This work is staged in [this PR](https://github.com/oxidecomputer/hubris/pull/539).

This PR loads either the old peripherals file or the new file, so it should be backwards-compatible.


bcantrill (approved):
Looks good!

PR #148, opened 2022-05-25
Add LPC55 ISP support

labbott:
The LPC55 has support for sending commands over UART. This is required
for certain functionality (e.g. key programming). Integrate this
support into humility.


cmd/isp/src/protocol.rs
```diff
@@ -0,0 +1,544 @@
...
+        } else if retval == 10101 {
+            Err(anyhow!(
+                "Incorrect signature. Is the SBKEK set correctly? (err 10101)"
+            ))
+        } else {
+            Err(anyhow!("ISP error returned: {}", retval))
+        }
+    } else {
+        Ok(params)
+    }
```
mkeeter, line 481: This could be a little cleaner as a match statement:
```suggestion
    match params[0] {
        0 => Ok(params),
        10203 => Err(anyhow!("Did you forget to erase the flash? (err 10203)")),
        10101 => Err(anyhow!(
            "Incorrect signature. Is the SBKEK set correctly? (err 10101)"
        )),
        retval => Err(anyhow!("ISP error returned: {}", retval)),
    }
```

cmd/isp/src/protocol.rs
```diff
@@ -0,0 +1,544 @@
...
+        let data_packet = DataPacket::new_data(data[cnt..end].to_vec())?;
+
+        let data_bytes = data_packet.to_bytes()?;
+
+        port.write_all(&data_bytes)?;
+        port.flush()?;
+
+        read_ack(port)?;
+        cnt += 512;
+    }
```
mkeeter, line 522: You can use [`chunks`](https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.chunks) to avoid doing this by hand:
```suggestion
    // Target doesn't like it when we send an entire binary in one pass
    // so break it down into 512 byte chunks which is what the existing
    // tools seem to use
    for chunk in data.chunks(512) {
        let data_packet = DataPacket::new_data(chunk.to_vec())?;

        let data_bytes = data_packet.to_bytes()?;

        port.write_all(&data_bytes)?;
        port.flush()?;

        read_ack(port)?;
    }
```

cmd/isp/src/protocol.rs
```diff
@@ -0,0 +1,544 @@
...
+}
+
+pub fn recv_data(
+    port: &mut dyn serialport::SerialPort,
+    cnt: u32,
+) -> Result<Vec<u8>> {
+    let mut data = Vec::new();
+    let mut received: usize = 0;
+
+    while received < (cnt as usize) {
```
mkeeter, line 534: Nit: you could do `while data.len() < (cnt as usize)` instead of tracking the size separately.

cmd/isp/src/cmd.rs
```diff
@@ -0,0 +1,218 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use crate::protocol::*;
+use anyhow::Result;
+
+enum DataPhase<'a> {
+    NoData,
+    Send { code: ResponseCode, data: &'a Vec<u8> },
```
mkeeter, line 10: Nit: Storing a reference to a `Vec` is a little weird – could you make it a `&'a [u8]` instead?

cmd/isp/src/cmd.rs
```diff
@@ -0,0 +1,218 @@
...
+pub fn do_isp_get_property(
+    port: &mut dyn serialport::SerialPort,
+    prop: BootloaderProperty,
+) -> Result<Vec<u32>> {
+    let args = vec![
+        // Arg 0 = property
+        prop as u32,
+    ];
+
+    send_command(port, CommandTag::GetProperty, args)?;
```
mkeeter, line 192: If you split `do_command` into multiple arguments like I suggested above, maybe `do_command_nodata` could return the result of `read_response`; then it could be used here and in `do_isp_last_error`

cmd/isp/src/lib.rs
```diff
@@ -0,0 +1,396 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! ## `humility isp`
```
mkeeter, line 5: You should run `cargo xtask readme` to put these docs in the README!

cmd/isp/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+//!
+//! The LPC55 has a built-in In-System Programming (ISP) mode. ISP mode
+//! does _not_ work over the regular debug probe but over UART. You must
+//! specificy the UART path specifcially
```
mkeeter, line 9: Nit: typos (`specificy` and `specifcially`)

cmd/isp/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+use crate::protocol::BootloaderProperty;
+
+#[derive(Parser, Debug)]
+#[clap(name = "subcmd")]
+enum IspCmd {
+    /// Runs a single ping to verify communication with the target
+    Ping,
+    /// Reads memory from the specified address
+    ReadMemory {
+        #[structopt(parse(try_from_str = parse_int::parse))]
```
mkeeter, line 39: Nit: now that we're on Clap 3, you can use `clap(parse(...))` instead of `structopt(parse(...))` for all of these.

cmd/isp/src/protocol.rs
```diff
@@ -0,0 +1,544 @@
...
+
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes)]
+pub struct PingResponse {
+    header: PacketHeader,
+    protocol_bugfix: u8,
+    protocol_minor: u8,
+    protocol_major: u8,
+    protocol_name: u8,
+    options_low: u8,
```
mkeeter, line 117: You could use `zerocopy::byteorder::U16<BigEndian>` (or is it `LittleEndian`?) for all of the two-byte values in the various `struct`s, which would save manually building them in constructors!

cmd/isp/src/protocol.rs
```diff
@@ -0,0 +1,544 @@
...
+    length_high: u8,
+    crc16_low: u8,
+    crc16_high: u8,
+}
+
+impl FramingPacket {
+    fn new(ptype: PacketType) -> FramingPacket {
+        FramingPacket {
+            header: PacketHeader::new(ptype),
+            length_low: 0,
```
mkeeter, line 137: Nit: passing in `length` here would make the API harder to misuse (I'd be worried about forgetting to set length values later on).

cmd/isp/src/protocol.rs
```diff
@@ -0,0 +1,544 @@
...
+    }
+
+    fn to_bytes(&self) -> Result<Vec<u8>> {
+        let mut v = Vec::new();
+
+        v.extend_from_slice(&self.packet.as_bytes());
+
+        for e in self.params.iter() {
+            v.extend_from_slice(&e.to_le_bytes());
+        }
```
mkeeter, line 224: If you want to get fancy
```suggestion
        v.extend(self.params.iter().flat_map(|e| e.to_le_bytes()));
```

cmd/isp/src/protocol.rs
```diff
@@ -0,0 +1,544 @@
...
+                if p.is_empty() {
+                    return Err(anyhow!(
+                        "Response returned an unknown error code?"
+                    ));
+                }
+                // The return value is always the first parameter
+                let retval = p[0];
+
+                // Some more specific error messages.
+                if retval == 10203 {
```
mkeeter, line 323: Same as my [suggestion below](https://github.com/oxidecomputer/humility/pull/148/files#r881810637), consider a `match` statement here to make it easier to extend in the future.

cmd/isp/src/protocol.rs
```diff
@@ -0,0 +1,544 @@
...
+        v.packet.crc16_low = (digest & 0xff) as u8;
+        v.packet.crc16_high = ((digest >> 8) & 0xff) as u8;
+
+        Ok(CommandPacket { packet: v, params: args })
+    }
+
+    fn to_bytes(&self) -> Result<Vec<u8>> {
+        let mut v = Vec::new();
+
+        v.extend_from_slice(&self.packet.as_bytes());
```
mkeeter, line 220: Optional: since `as_bytes` returns `&[u8]`, which implements `IntoIter`, you can do
```suggestion
        v.extend(self.packet.as_bytes());
```
(not sure if this is _better_, but it's got one fewer `&`, which is always nice)

cmd/isp/src/protocol.rs
```diff
@@ -0,0 +1,544 @@
...
+
+    Ok(())
+}
+
+pub fn read_data(port: &mut dyn serialport::SerialPort) -> Result<Vec<u8>> {
+    let mut frame_bytes = vec![0; core::mem::size_of::<FramingPacket>()];
+    let mut cnt = 0;
+
+    while cnt != core::mem::size_of::<FramingPacket>() {
+        let r = port.read(&mut frame_bytes[cnt..])?;
```
mkeeter, line 377: Could you use `read_exact` here?

cmd/isp/src/protocol.rs
```diff
@@ -0,0 +1,544 @@
...
+            frame.header.packet_type
+        ));
+    }
+
+    cnt = 0;
+    let length: usize =
+        (frame.length_low as usize) | ((frame.length_high as usize) << 8);
+    let mut response = vec![0; length];
+
+    while cnt != length {
```
mkeeter, line 396: Same here (using `read_exact` instead of `read` in a loop)

cmd/isp/src/protocol.rs
```diff
@@ -0,0 +1,544 @@
...
+}
+
+pub fn read_response(
+    port: &mut dyn serialport::SerialPort,
+    response_type: ResponseCode,
+) -> Result<Vec<u32>> {
+    let mut frame_bytes = vec![0; core::mem::size_of::<FramingPacket>()];
+    let mut cnt = 0;
+
+    while cnt != core::mem::size_of::<FramingPacket>() {
```
mkeeter, line 413: Same here

cmd/isp/src/protocol.rs
```diff
@@ -0,0 +1,544 @@
...
+            response_type as u8,
+            command.tag
+        ));
+    }
+
+    let mut params: Vec<u32> = Vec::new();
+    let mut cnt = 0;
+    let mut index = core::mem::size_of::<RawCommand>();
+
+    while cnt < command.parameter_count {
```
mkeeter, line 457: Nit: You could use `params.len()` instead of `cnt`, to keep track of one fewer thing

PR #157, opened 2022-06-14
Bump Idolatry and add handling for new error return kind

mkeeter:


humility-core/src/hubris.rs
```diff
@@ -16,7 +16,7 @@ use std::fs::{self, OpenOptions};
 use std::io::Cursor;
 use std::mem::size_of;
 use std::path::Path;
-use std::str;
+use std::str::{self, FromStr};
```
mkeeter, line 19: No idea why this needed to change!

bcantrill: That... is weird?

bcantrill (approved):
Looks great -- thank you!

PR #159, opened 2022-06-15
Use new halt API for reading HIFFY_SCRATCH

andrewjstone:


bcantrill (approved):
This looks fine -- though I think in general we want want to keep as many operations within a single `op_start`/`op_done` pair because on some platforms (specifically, ARMv6-M), this *will* induce a halt and run.  Still, this should integrate as is for now -- thank you for spotting this!

PR #161, opened 2022-06-16
Upgrade clap to 3.2.5

steveklabnik:
Clap 3.2 is the last release before 4.0:
https://epage.github.io/blog/2022/06/clap-32-last-call-before-40/

These changes upgrade our usage of features that are being deprecated,
insuring that we shall have an easy upgrade path to 4.0.

The two biggest changes:

1. removing NoAutoVersion.
2. changing our usage of parse_int::parse to the new value_parser API

Big thanks to @epage for helping out with 2:
https://github.com/clap-rs/clap/discussions/3839

And for working through some error messages:
https://github.com/clap-rs/clap/issues/3822

As for NoAutoVersion, opting into the new behavior via `action` does
what we want, *except* that the output for doing it with another
incorrect flag doesn't give the error about how that's incorrect. I
suspect that this is an acceptable change in behavior, but worth calling
out. You can see the difference in the chip.trycmd tests.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,6 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use clap::ValueSource;
 use humility_cmd::{Args, Subcommand};
 
 use clap::CommandFactory;
@@ -57,17 +58,18 @@ fn main() {
     // line) to win the conflict.
     //
     if args.dump.is_some() && args.archive.is_some() {
-        let dump_count =
-            m.get_many::<usize>("dump").map(|m| m.len()).unwrap_or(0);
-        let archive_count =
-            m.get_many::<usize>("archive").map(|m| m.len()).unwrap_or(0);
-        match (dump_count == 1, archive_count == 1) {
-            (true, true) => {
+        use ValueSource::*;
+
+        let dump = m.value_source("dump").unwrap();
+        let archive = m.value_source("archive").unwrap();
+
+        match (dump, archive) {
+            (CommandLine, CommandLine) => {
                 log::error!("cannot specify both a dump and an archive");
                 std::process::exit(1);
             }
 
-            (false, false) => {
+            (EnvVariable, EnvVariable) => {
                 log::error!(
                     "both dump and archive have been set via environment \
                     variables; unset one of them, or use a command-line option \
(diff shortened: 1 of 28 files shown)


src/main.rs
```diff
@@ -57,8 +57,11 @@ fn main() {
     // line) to win the conflict.
     //
     if args.dump.is_some() && args.archive.is_some() {
-        match (m.occurrences_of("dump") == 1, m.occurrences_of("archive") == 1)
-        {
+        let dump_count =
+            m.get_many::<usize>("dump").map(|m| m.len()).unwrap_or(0);
+        let archive_count =
+            m.get_many::<usize>("archive").map(|m| m.len()).unwrap_or(0);
+        match (dump_count == 1, archive_count == 1) {
```
steveklabnik, line 64: @epage feel absolutely zero obligation to respond, but this is one change where this API seems to be slightly more verbose. I don't know if that's expected or if I'm maybe doing something wrong here, if there's a more modern UI to make this happen, I'd love to upgrade this code!

epage: > we want to allow an explicitly set value (that is, via the command line) to win the conflict.

`occurrences_of` was an indirect way of checking if something was set on the command-line.  We now have `ArgMatches::value_source(id) -> Option<ValueSource>` which let's you explicitly check if the value came from the command-line or from the environment (#3844 updates the deprecation message).  `ValueSource` is also `PartialEq + Eq + PartialOrd + Ord`.  Unsure if a `match` is preferred or seeing which one has a higher priority source.

On the other hand, the approach this was replaced with just returns the number of values, regardless of where it came from.

btw I wonder if there is an ergonomic enough / not-too-terrible-document way of exposing `value_source` in the derive API.  Feel free to create an issue about it.  The only other piece of information I can think of that is in `ArgMatches` but not in the derive is arg group values.  We push all values that belong to arguments in groups to the group's id.  I do want to change that as it seems wasteful when I bet most people don't use it or don't even know it exists.  I think I'll make an `Action` to control it.

steveklabnik: Nice, this worked really well. Thank you!

> Unsure if a match is preferred or seeing which one has a higher priority source.

In this specific case, you can only have one of either of these, the more accurate check would be "not zero" rather than "is one" in some sense. The trick is mostly that folks will often want to provide these via environment variables for normal use, and then use the command line to override that env var when doing something special. So we don't want to make them actually exclusionary, because then users would have to go remove their normal environment to get things to work, which would be awkward.

I'm not sure how to expose it via the derive API either, to be honest. If I can think of anything, let me know.

steveklabnik:
This is so out of date that if I pursue this again in the future I'll have to re-do it from scratch anyway, closing.

PR #162, opened 2022-06-17
Add checks for Hubris archive version

mkeeter:
This will give us better error messages when making breaking changes on the Hubris side!


bcantrill (approved):
Love it! Our future selves will thank you for planting this tree today.

PR #165, opened 2022-06-17
Decode Hiffy return values with `ssmarshal`

mkeeter:
Right now, many of our Hiffy commands use `ssmarshal` to encode parameters or return values.  Consider the following fancy `struct` and a function which generates it:

```rust
#[derive(Copy, Clone, Serialize, Deserialize)]
pub enum Bla {
    Boo,
    Far(u16),
    Baz,
}

#[derive(Copy, Clone, Serialize, Deserialize)]
pub struct FancyTestType {
    pub u: u32,
    pub b: bool,
    pub f: f32,
    pub e: Bla,
    pub g: Bla,
}
```
```ron
        "return_struct_ss": (
            args: {
                "a": "u32",
            },
            reply: Result(
                ok: "FancyTestType",
                err: CLike("NetError"),
            ),
            encoding: Ssmarshal,
        ),
```
```rust
    fn return_struct_ss(
        &mut self,
        _: &userlib::RecvMessage,
        u: u32,
    ) -> Result<FancyTestType, RequestError<NetError>> {
        Ok(FancyTestType {
            u,
            b: true,
            f: 1.23,
            e: Bla::Far(91),
            g: Bla::Boo,
        })
    }
```

If you invoke this function in Humility, it will attempt to decode the buffer based on the DWARF data, and get it _terribly wrong_:
```
Net.return_struct_ss() =  { u: 0xc, b: false, f: -0.0000000000000000000031848523554306865, e: <? (0x13f)>, g: Boo }
```

This PR adds special decoding for Hiffy calls that using the `ssmarshal` encoding, so that Humility prints the output correctly.

```
Net.return_struct_ss() = FancyTestType { u: 0xc, b: true, f: 1.23, e: Far(0x5b), g: Boo }
```

I don't _love_ hard-coding `ssmarshal` details into the deserialization, but `ssmarshal` doesn't expose its `Deserializer` class, and `serde` doesn't have a great way of doing deserialization from an out of band format...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -908,6 +908,7 @@ dependencies = [
  "log",
  "parse_int",
  "postcard",
+ "ssmarshal",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


bcantrill (approved):
Looks great -- thank you!

PR #166, opened 2022-06-17
Consolidate struct formatting code

### LINKED ISSUE #164: Reimplement printfmt in terms of reflect
Right now, both of these snippets do the same-ish thing through completely different code paths:

```rust
println!("{}", hubris.printfmt(val, ty, &fmt)?);
```

```rust
humility_cmd::reflect::load_value(
    hubris,
    val,
    hubris.lookup_type(ty).unwrap(),
    0,
)?
.format(hubris, fmt, &mut std::io::stdout())?;
```

They read a value out of memory, then print it to the screen.

We should implement the former in terms of the latter `reflect` API, to reduce duplicated code.

mkeeter:
This PR removes a duplicate implementation of walking + printing arbitrary objects!  We now reusing the implementation from `reflect.rs` everywhere.

It requires moving `reflect.rs` from `humility-cmd` into `humility-core`, since `HubrisArchive` now uses it.

Fixes #164


mkeeter:
Interesting, the test suite caught a formatting change!

## Before
```
TICKS (0x2000104c) = [
     {
        v: UnsafeCell<u32> {
            value: 0x0
        }
    },
     {
        v: UnsafeCell<u32> {
            value: 0x0
        }
    }
]
```

## After
```
TICKS (0x2000104c) = [
    AtomicU32 {
        v: UnsafeCell<u32> {
            value: 0x0
        }
    },
    AtomicU32 {
        v: UnsafeCell<u32> {
            value: 0x0
        }
    }
]
```

------------------------------------------------------------------------

I think **After** is better, so I'm going to edit the test case to match.

bcantrill (approved):
Looks great -- thanks for doing this!

PR #169, opened 2022-06-22
Add ssmarshal encoding for Hiffy calls

mkeeter:
Previously, calling a Hiffy function that used `ssmarshal` encoding would mess up the arguments.


bcantrill (approved):
Looks great. I would love to have a way to test this better (it all more or less has to be tested manually), but that's definitely not a new problem!

PR #179, opened 2022-07-05
"humility ringbuf" filter by type instead of name

jgallagher:
Instead of filtering out any variables whose name does not end with `RINGBUF`, `humility ringbuf` now attempts to check whether the stringified type name contains `ringbuf::Ringbuf`, allowing it to list and match on ringbufs with custom names. I left the `.ends_with("RINGBUF")` check in place for variables for which we fail to look up the type definition; I'm unclear on whether that's necessary, but it seemed easy enough to leave in place.

### DIFF
--- a/cmd/ringbuf/src/lib.rs
+++ b/cmd/ringbuf/src/lib.rs
@@ -156,7 +156,7 @@ fn ringbuf(context: &mut humility::ExecutionContext) -> Result<()> {
             Ok(s) => {
                 // Skip variables whose type does not indicate they contain a
                 // ringbuf; this check is imprecise but probably good enough
-                if s.name.contains("ringbuf::Ringbuf") {
+                if s.name.contains("Ringbuf") {
                     Some(s)
                 } else {
                     continue;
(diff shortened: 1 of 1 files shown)


cmd/ringbuf/src/lib.rs
```diff
@@ -151,16 +151,32 @@ fn ringbuf(
     let mut ringbufs = vec![];
 
     for v in hubris.qualified_variables() {
+        let def = match hubris.lookup_struct(v.1.goff) {
+            Ok(s) => {
+                // Skip variables whose type does not indicate they contain a
+                // ringbuf; this check is imprecise but probably good enough
+                if s.name.contains("Ringbuf") {
```
jgallagher, line 158: FWIW my first shot here was `"ringbuf::Ringbuf"`, which worked on the latest archive I'm using from hubris but failed for 99 of the committed tests. My second try was `Ringbuf<`, which failed for only two tests (which I find somewhat surprising). `Ringbuf` passes all tests and is probably fine, but if you have insight into why the other two failed I'd appreciate hearing it.

mkeeter (approved):

PR #183, opened 2022-07-06
Add humility command for the LPC55 debug mailbox

labbott:
The LPC55 has a dedicated access port for doing useful thing such as
forcing the device into SWD and ISP mode. Add a humility command to
handle this.

### DIFF
--- a/humility-arch-cortex/src/scs.rs
+++ b/humility-arch-cortex/src/scs.rs
@@ -470,7 +470,7 @@ impl CoreInfo {
             let ctrl = LPC55_SYSCON_AHBCLKCTRL0::read(core)?;
 
             if !ctrl.iocon() {
-                bail!("IOCON is not clocked");
+                warn!("IOCON is not clocked");
             }
         }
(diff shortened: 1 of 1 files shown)


cmd/debugmailbox/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! ## `humility debugmailbox`
+//!
+//! The LPC55 includes an extra access port referred to as the Debug Mailbox.
+//! This allows for running a fixed set of commands to do useful things such
+//! as forcing SWD enablement and putting the chip into ISP mode without
+//! needing to touch an external pin
+//!
+//! $ humility debugmailbox debug
```
bcantrill, line 12: These can be wrapped in '```console' to get them to format properly

bcantrill (approved):
Looks great -- small doc comment to get the docs to format correctly

PR #186, opened 2022-07-08
cmd/flash: better reporting for command not found

### LINKED ISSUE #185: humility flash with pyocd doesn't say what is missing
If you have some file missing for some reason while trying to use `humility flash` on the RoT, it's hard to know what it is that's actually missing because the error message doesn't say:

```
$ sudo ./target/release/humility -a ../hubris/target/gimlet-rot/dist/build-gimlet-rot.zip  flash
humility: specifying serial XAEPYT3E3JX0D
humility flash failed: failed to flash ("pyocd" "flash" "-t" "lpc55s28" "--format" "hex" "/tmp/.tmpmN8PJ1" "-u" "XAEPYT3E3JX0D")

Caused by:
    No such file or directory (os error 2)
```

It'd really help if it could tell us what actually caused an ENOENT.

cbiffle:
Fixes #185.

### DIFF
--- a/cmd/flash/src/lib.rs
+++ b/cmd/flash/src/lib.rs
@@ -23,6 +23,7 @@ use humility::hubris::*;
 use humility_cmd::{Archive, Args, Command, RunUnattached};
 use path_slash::PathExt;
 use std::io::Write;
+use std::process::ExitStatus;
 
 use serde::Deserialize;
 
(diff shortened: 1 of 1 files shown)


bcantrill (approved):

rmustacc:
This does work as advertised in the case that came up:

```
$ sudo -E ./target/release/humility -a ../hubris/target/gimlet-rot/dist/build-gimlet-rot.zip  flash -F
humility: archive appears to be already flashed; forcing re-flash
humility: specifying serial XAEPYT3E3JX0D
humility flash failed: unable to execute "pyocd", is it in your PATH and executable?

Caused by:
    No such file or directory (os error 2)
```

PR #197, opened 2022-07-28
Fix Humility compatibility with rustc 1.64

cbiffle:
Humility has historically required all symbols to be described the same way in both DWARF and ELF. Nothing actually requires the compiler to behave this way, however, and somewhere between the old version we've been stuck on, and 1.64, this behavior changed.

This caused Humility to refuse to load new images, failing in creative and inscrutable ways.

To address the loading, this PR teaches Humility how to parse and evaluate the DWARF `AT_location` expressions for static variables, and use those if there is no matching information in the ELF. This fixes image loading.

To address the "creative and inscrutable" part, this PR also improves error message reporting for type resolution failures and value loading.

Finally, it turns out that rustc is now generating unnamed pointer types in certain circumstances; this PR hackishly treats them as `"<UNNAMED>"` since Humility in general requires things to have names.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1472,6 +1472,7 @@ dependencies = [
  "clap",
  "humility-cmd",
  "humility-core",
+ "log",
  "num-traits",
 ]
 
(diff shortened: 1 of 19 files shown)


cbiffle:
@mkeeter and I feel that this is urgent enough (Hubris-Humility interaction is currently broken for everyone on master) that I'm setting this to automerge without review.

PR #199, opened 2022-07-29
Bump Humility reported version to 0.8

cbiffle:
People were unable to tell if they had the fixes from yesterday installed, so, time to bump.


rmustacc (commented):
From looking at 8567d0306f42689e1f67958039e059035594ca15 there may be some tests that need to be updated.

PR #200, opened 2022-08-01
Add support for applying an update

labbott:


cmd/update/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+fn update(context: &mut humility::ExecutionContext) -> Result<()> {
+    let core = &mut **context.core.as_mut().unwrap();
+    let hubris = context.archive.as_ref().unwrap();
+    let Subcommand::Other(subargs) = context.cli.cmd.as_ref().unwrap();
+
+    let subargs = UpdateArgs::try_parse_from(subargs)?;
+    let mut context = HiffyContext::new(hubris, core, subargs.timeout)?;
+    let funcs = context.functions()?;
+
+    let start_update_fn = funcs.get("StartUpdate", 0)?;
```
mkeeter, line 63: Vague architecture question: why are these specially-defined Hiffy functions (e.g. [here](https://github.com/oxidecomputer/hubris/blob/92397da6b5a7a003e63d0bc1e343b1a437002002/task/hiffy/src/stm32h7.rs#L135-L142)), instead of calls through the `Update.*` Idol API ([here](https://github.com/oxidecomputer/hubris/blob/92397da6b5a7a003e63d0bc1e343b1a437002002/idl/update.idol))?  It seems like this is an unnecessary duplication of API surface area.

cmd/update/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    bar.finish_and_clear();
+    if subargs.skip_commit {
+        humility::msg!("Not committing update");
+    } else {
+        humility::msg!("Comitting update");
+        let ops = vec![Op::Call(finish_update_fn.id), Op::Done];
+        let _ = context.run(core, ops.as_slice(), None)?;
+    }
```
mkeeter, line 136: For systems with auxiliary flash data (just Sidecar at the moment), that flash _also_ needs to be programmed for the system to be functional.

This is a bit awkward: programming the auxflash uses Idol APIs, so we have to reboot into the new image to access them (since we have no knowledge of the previous image).

[Here's](https://github.com/oxidecomputer/humility/blob/master/cmd/flash/src/lib.rs#L429-L506) where that's done in `humility flash`.

It looks like this `update` subcommand doesn't reboot into the new image at all (with or without `--skip-commit`), and also doesn't program the auxiliary flash, meaning you could end up with broken systems next time it's rebooted.

For now, I'd vote to either

(a) Refuse to run `humility update` if there is auxiliary flash data in the Hubris archive **at all**, or
(b) Refuse to run `humility update` if there's auxflash data **and** `--skip-commit` is set, and reboot into the new image + program the auxflash if `--skip-commit` is not set

(more generally, refuse to run a `humility update` command that could leave the system broken)

cmd/update/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+        _ => bail!("Couldn't find image target {}", subargs.target),
+    };
+
+    let ops = vec![Op::Call(block_size_fn.id), Op::Done];
+
+    let results = context.run(core, ops.as_slice(), None)?;
+
+    let blk_size = match &results[0] {
+        Ok(s) => {
+            if s.len() < 4 {
```
mkeeter, line 84: Should this be `s.len() == 4` instead, or do we expect extra data to be returned?

cmd/update/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+                bail!("Bad length for block size!");
+            }
+            (s[0] as u32)
```
mkeeter, line 87: If we know that `s` is exactly 4 `u8`s long, then
```rust
u32::from_le_bytes(s.try_into().unwrap())
```
otherwise
```rust
u32::from_le_bytes(s[..4].try_into().unwrap())
```

cmd/update/src/lib.rs
labbott: Can you give me a pointer to how you would call these through idol? This was the technique I know and love for adding humility cmmands.

cmd/update/src/lib.rs
labbott: I think this is expecting `humility update` to work at a much higher level than it currently does.  It has no knowledge of hubris archives and the path that it takes is a raw binary. The existing behavior is closer to `humility gpio` than `humility flash` in that it can in fact put the system in a weird state which is actually very useful for testing. I agree that what you're describing is what we will eventually want but I'm also not sure if this is the right time to start adding it since we're still working out other parts of the update sequence.

mkeeter: Okay, that makes sense – if this is taking raw `.bin` files (rather than Hubris archives), then I agree that it shouldn't care about auxflash.

cmd/update/src/lib.rs
mkeeter: Sure, [here's how it's done in `humility auxflash`](https://github.com/oxidecomputer/humility/blob/master/cmd/auxflash/src/lib.rs#L328-L340)

This is somewhat new: it relies on `humility hiffy` support for leases, which is a recent addition!

The advantage is that – instead of needing special-case Hiffy functions and manually building HIF programs – you can call _any_ Idol interface that takes a lease, using a nice high-level API.

mkeeter (approved):

PR #205, opened 2022-08-08
Add `humility reset`

labbott:
This will reset the target using the `nRESET` pin


cmd/reset/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+    let subargs = ResetArgs::try_parse_from(subargs)?;
+
+    // Get a list of all available debug probes.
+    let probes = Probe::list_all();
+
+    if probes.is_empty() {
+        bail!("No probes found!");
+    }
+
+    let num = subargs.probe_num.unwrap_or(0);
```
bcantrill, line 39: It feels like it would be more consistent to allow the probe to be specified on the command line (e..g, as we do for `humility probe`), but I think that would necessitate adding a `reset` function to the `Core` trait (one that would only be implemented for `ProbeCore`).  But this would allow one to (e.g.) reset a probe by specifying VID/PID/Serial.

bcantrill (changes requested):
I think it's worth making this consistent with the other Humliity commands, which I don't *think* will be too bad -- really just changing where the reset logic is implemented.

cmd/reset/src/lib.rs
labbott: I don't think `ProbeCore` is the right place. `ProbeCore` implies an attachment to a chip which I wanted to avoid with `humility reset` for cases where something has truly wedged the chip. I agree about making it consistent with other humility commands so I'll give this some more thought.

humility-core/src/core.rs
```diff
@@ -1085,55 +1182,121 @@ impl Core for DumpCore {
...
+fn parse_probe(probe: &str) -> (&str, Option<usize>) {
     if probe.contains('-') {
         let str = probe.to_owned();
         let pieces: Vec<&str> = str.split('-').collect();
 
         if pieces[0] == "usb" && pieces.len() == 2 {
             if let Ok(val) = pieces[1].parse::<usize>() {
-                index = Some(val);
-                probe = "usb";
+                //index = Some(val);
```
bcantrill, line 1192: Looks like we can ditch these lines. ;)

cmd/reset/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+        None => "auto",
+    };
+
+    let mut c = if subargs.soft_reset {
+        // The LPC55 based chips have a special reset sequence. Using this
+        // requires attaching with the specific chip name. Right now there
+        // isn't an easy way to get the chip name from the Hubris archive
+        // so we attach to a generic chip. The result is that a soft reset
+        // on LPC55 targets will work but it will produce some error messages.
+        println!("Some errors may be expected when doing a soft reset");
```
bcantrill, line 42: I think you want to make these `humility::msg!()`.

bcantrill (approved):
Thanks for this -- much cleaner, and will (or should?) Just Work with environments.

PR #209, opened 2022-08-10
Fix `humility openocd` when using environments

### LINKED ISSUE #208: `humility openocd` doesn't work with environments
```console
matt@niles ~ () $ pfexec humility -t gimlet openocd
humility openocd failed: Cannot specify --probe with `openocd` subcommand
```

I believe this is because the environment [unconditionally sets `args.probe` here](https://github.com/oxidecomputer/humility/blob/d36039eb9e030cfb67e5610273803b68b4bdc9a5/src/main.rs#L65), which is then checked [here](https://github.com/oxidecomputer/humility/blob/d36039eb9e030cfb67e5610273803b68b4bdc9a5/cmd/openocd/src/lib.rs#L41).

mkeeter:
This kills two birds with one stone:

- It allows us to use `humility openocd` when using an environment + target
- It selects the _correct_ probe (based on serial name), for systems with multiple probes attached

I rolled the same fixes into `humility gdb`, which would otherwise fail in the same way.

Fixes #208

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -854,7 +854,7 @@ dependencies = [
 
 [[package]]
 name = "humility"
-version = "0.8.2"
+version = "0.8.3"
 dependencies = [
  "anyhow",
  "bitfield",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1093,6 +1093,7 @@ dependencies = [
  "clap",
  "ctrlc",
  "humility-cmd",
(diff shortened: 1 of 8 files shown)


bcantrill (approved):
This looks great -- my only comment is that I think it's worth bumping the micro/patch of the Humility version so folks can easily know if they have this fix or not!

PR #210, opened 2022-08-10
Add `monorail mac` subcommand

mkeeter:
Depends on [hubris#704](https://github.com/oxidecomputer/hubris/pull/704)

Sample output:

```console
matt@niles ~ (sidecar) $ pfexec ./humility monorail mac
humility: attached to 0483:374f:002A001C4D46500F20373033 via ST-Link V3
Reading 3 MAC addresses...
 PORT |        MAC
------|-------------------
   18 | 0e:1d:23:88:1a:3b
   41 | 0e:1d:7f:c3:07:31
   48 | 0e:1d:15:70:3d:bb
```

The only downside is that you need to make two HIF calls (one to get the MAC address count, and a second call to actually loop over them).


bcantrill (approved):
Looks great -- at some point (maybe now, maybe later?) it might be worth a little more documentation on `humliity monorail`? (But I say this as one who has commands that are still entirely undocumented!)

mkeeter:
Yeah, probably time for some more docs – I added explanations of the main subcommands.

PR #211, opened 2022-08-11
Allow specifying multiple archives in environment

labbott:
Recent changes to Hubris require multiple archive files to fully
describe what is flashed on some targets. Update the environment
files to account for this.

### DIFF
--- a/humility-cmd/src/lib.rs
+++ b/humility-cmd/src/lib.rs
@@ -55,6 +55,11 @@ pub struct Args {
         conflicts_with_all = &["dump", "probe"])]
     pub target: Option<String>,
 
+    /// If multiple archives are specified in an environment, name of
+    /// the archive to use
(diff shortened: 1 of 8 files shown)


kc8apf (approved):

humility-cmd/src/lib.rs
```diff
@@ -55,6 +55,9 @@ pub struct Args {
         conflicts_with_all = &["dump", "probe"])]
     pub target: Option<String>,
 
+    #[clap(long, requires = "environment")]
+    pub archive_name: Option<String>,
```
bcantrill, line 59: This needs a documentation string.

bcantrill (approved):
This looks good. I think it would be that it's worth updating the Environment section of README.md to describe this capability and its intent. (I think it's also worth rolling the micro version of Humility for this.)

PR #213, opened 2022-08-12
Add `net` subcommand

mkeeter:
The `net` subcommand allows for management and debugging of the _client side_ of the management network, i.e. the KSZ8463 + VSC85x2 which is common across all boards.

Depends on [hubris#709](https://github.com/oxidecomputer/hubris/pull/709)

# Todo
- [x] Bump version
- [x] Write some docs

### DIFF
--- a/cmd/net/src/lib.rs
+++ b/cmd/net/src/lib.rs
@@ -2,6 +2,40 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+//! ## `humility net`
+//! `humility net` exposes commands to interact with the management network from
+//! the client's perspective.
+//!
+//! (It is the counterpart to `humility monorail`, which interacts with the
+//!  management network from the _switch_'s perspective.)
+//!
+//! It is fully functional on
+//! - Gimlet
+//! - Sidecar
+//! - PSC
+//! - `gimletlet-mgmt`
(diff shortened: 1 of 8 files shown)


cmd/net/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+
+    // Due to HIF limitations (lack of Expand16 / Collect16), we're going to
+    // limit ourselves to 255 MAC table entries.
+    let (mac_count, clipped) = if mac_count > u8::MAX as u32 {
+        (u8::MAX, true)
+    } else {
+        (mac_count as u8, false)
+    };
+
+    println!("Reading {} MAC addresses...", mac_count);
```
labbott, line 135: these `println` should be `humility::msg`

cmd/net/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+                println!("Skipping empty MAC address");
+            } else {
+                mac_table.entry(port).or_default().push(mac);
+            }
+        } else {
+            // Log the error but keep going for other entries in the table
+            println!("Got error result: {:?}", r);
+        }
+    }
+    println!(" {} |        {}", "PORT".bold(), "MAC".bold());
```
labbott, line 193: I wonder if it would be worth it to put our table printing into a library? (Not needed, more thinking long term)

mkeeter: Yeah, or maybe switching to something like [`tabled`](https://crates.io/crates/tabled)...

labbott (commented):
LGTM!

PR #219, opened 2022-08-22
Fix the build on Windows and add GitHub Action

steveklabnik:
Let's see if I did the action right. 🤞

### DIFF
--- a/cmd/rpc/Cargo.toml
+++ b/cmd/rpc/Cargo.toml
@@ -17,7 +17,12 @@ anyhow = { version = "1.0.44", features = ["backtrace"] }
 clap = { version = "3.0.12", features = ["derive", "env"] }
 colored = "2.0.0"
 indexmap = "1.7"
-libc = "0.2"
 log = {version = "0.4.8", features = ["std"]}
 parse_int = "0.4.0"
 zerocopy = "0.6.1"
+
+[target.'cfg(not(windows))'.dependencies]
+libc = "0.2"
+
+[target.'cfg(windows)'.dependencies]
+winapi = { version = "0.3.9", features = ["netioapi"] }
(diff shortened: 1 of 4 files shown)


steveklabnik:
I do not understand why the "check", "clippy" and "test suite" jobs aren't running. It's late here, and I'm off tomorrow, so I'm going to leave this as-is; if anyone wants to fix the Windows build in the meantime, pulling in the first commit should work just fine.

cmd/rpc/Cargo.toml
```diff
@@ -21,3 +21,6 @@ libc = "0.2"
 log = {version = "0.4.8", features = ["std"]}
 parse_int = "0.4.0"
 zerocopy = "0.6.1"
+
+[target.'cfg(windows)'.dependencies]
```
mkeeter, line 25: Nit: do we want to make the `libc` dependency only included on `not(windows)`?

steveklabnik: Seems fine to me, I was being lazy; it's probably transitively included anyway.

steveklabnik:
Looking again, they still say the status haven't been reported, but if I dig way down:

image

So, an error that doesn't show up as an error. Sigh. At least this points to what needs fixed!

steveklabnik:
Okay this is looking good! Turns out an array of `runs-on` is not the behavior I thought it was, haha. I've enabled auto-merge so that we can get to testing asap; if anyone has any other tweaks they'd like me to make here, please let me know and I'd be happy to make a follow-up PR.

steveklabnik:
I am *very* confused about these top-level "Check" and "Test Suite" jobs that aren't running. The two jobs named these things are both under "Continuous integration" and are shown above, and working. When I go to the "checks" tab, these two things don't show up. wtf.

steveklabnik:
Thank you @alilleybrinker and others; this was in fact a branch permissions thing. Fixed now!

PR #221, opened 2022-08-24
What the heck gha

steveklabnik:
This is https://github.com/oxidecomputer/humility/pull/219 but with the intention of seeing if the weird thing I'm seeing in Actions is some messed up state.


steveklabnik:
image

It's still there. This isn't the problem.

PR #223, opened 2022-09-02
Adds handling of hiffy functions that take a single lease

mkeeter:
Depends on [hubris#748](https://github.com/oxidecomputer/hubris/pull/748)

### DIFF
--- a/cmd/hiffy/src/lib.rs
+++ b/cmd/hiffy/src/lib.rs
@@ -92,6 +92,10 @@ struct HiffyArgs {
     #[clap(long, short, requires = "call", conflicts_with = "input")]
     output: Option<String>,
 
+    /// print returned data in hex
+    #[clap(short = 'x', requires = "num")]
+    hex: bool,
+
     /// arguments
     #[clap(long, short, requires = "call")]
     task: Option<String>,
@@ -514,6 +518,8 @@ fn hiffy(
                     std::fs::write(out, &data)
                         .context(format!("Could not write to {}", out))?;
                     println!("Wrote {} bytes to '{}'", data.len(), out);
+                } else if subargs.hex {
+                    println!("Data: {:x?}", data);
                 } else {
                     println!("Data: {:?}", data);
                 }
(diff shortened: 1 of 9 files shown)


cmd/hiffy/src/lib.rs
```diff
@@ -314,7 +466,61 @@ fn hiffy(
...
+
+        if let Some(input) = input {
+            let r = hiffy_call_write(
+                hubris,
+                core,
+                &mut context,
+                &op,
+                &args,
+                &input,
+            )?;
```
kc8apf, line 499: I understand wanting to put the precondition checks inside hiffy_call_write but the error messages that are returned assume you know the internal details that --input invokes a method called hiffy_call_write.

humility-cmd/src/hiffy.rs
```diff
@@ -399,6 +399,89 @@ impl<'a> HiffyContext<'a> {
...
+    }
+
+    /// Convenience routine to translate an Idol call into HIF operations
+    pub fn idol_call_ops_rw(
+        &self,
+        funcs: &HiffyFunctions,
+        op: &idol::IdolOperation,
+        payload: &[u8],
+        ops: &mut Vec<Op>,
+        lease_size: u32,
```
kc8apf, line 446: Splitting off the no-lease/read/write paths this low in the stack appears to create a lot of new parallel APIs upwards.  I suspect this PR would be a lot smaller if you:
* added a `lease_size: Option<u32>` argument to idol_call_ops()
* added `read_size: Option<usize` and `write_data: Option<&[u8]>` to hiffy_call()
* changed hiffy_call() return type to `Result<std::result::Result<(humility::reflect::Value, Option<Vec<u8>>), String>>`
* use an enum for hiffy_call()'s Result error type to indicate lease mismatches to the caller

mkeeter: Good suggestion – I didn't do _exactly_ this, but consolidated a bunch in 836c1f65710561ed604f62e411140404b6be4dc9 and c78ff03235f46e9592b04ad73114f99de106de4f

kc8apf (approved):

bcantrill (approved):
Looks great -- thanks for doing this!

PR #225, opened 2022-09-06
Port to support RV32 hubris

beezow:
This modifies the core of humility to support riscv32, along with most of the commands.  It pairs with this [hubris port](https://github.com/rivosinc/hubris).

Since changes to the humility-core have a cascading effect to each command, each commit may not build.  This pull request is intended to be taken as a whole patch series.


FawazTirmizi:
oops

PR #227, opened 2022-09-09
Add support for auxiliary flash

mkeeter:
Proposed commit message:

```
Adds support for auxiliary flash!

This PR adds the `humility auxflash` subcommand, which exposes tools for
reading, writing, and checking the status of an auxiliary flash chip (when
present).

In addition, it modifies `humility flash` to program an auxiliary flash if a
blob is found in the Hubris archive.

This depends on [hubris#754](https://github.com/oxidecomputer/hubris/pull/754/),
which adds the `auxflash` task and modifies our archive packing to include these
blobs (based on the app TOML file).
```

(this should be merged with https://github.com/oxidecomputer/hubris/pull/754 , but is not strictly dependent on it)

### DIFF
--- a/cmd/auxflash/src/lib.rs
+++ b/cmd/auxflash/src/lib.rs
@@ -1,0 +1,147 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! ## `humility auxflash`
+//!
+//! Tools to interact with the auxiliary flash.
+use std::collections::BTreeMap;
+
+use anyhow::{anyhow, bail, Context, Result};
+use clap::{Command as ClapCommand, CommandFactory, Parser};
+use colored::Colorize;
+
+use humility::core::Core;
+use humility::hubris::*;
+use humility::reflect::*;
+use humility_cmd::hiffy::HiffyContext;
+use humility_cmd::idol::IdolArgument;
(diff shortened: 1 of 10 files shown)


cmd/auxflash/src/lib.rs
```diff
@@ -0,0 +1,410 @@
...
+        let context = HiffyContext::new(hubris, core, hiffy_timeout)?;
+        Ok(Self { hubris, core, context })
+    }
+
+    fn get_idol_command(&self, name: &str) -> Result<IdolOperation<'a>> {
+        IdolOperation::new(self.hubris, "AuxFlash", name, None).with_context(
+            || {
+                format!(
+                    "Could not find `AuxFlash.{}`, \
+                     is your Hubris archive new enough?",
```
bcantrill, line 89: An error message in the Humility style! ;)

cmd/flash/src/lib.rs
```diff
@@ -363,28 +373,102 @@ fn flashcmd(
...
+    //
+    let delay = subargs.reset_delay;
+
+    if delay != 0 {
+        std::thread::sleep(std::time::Duration::from_millis(delay));
+    }
+
+    core.reset()?;
+
+    try_program_auxflash(hubris, core)?;
```
bcantrill, line 397: Worth adding a comment here explaining that it is deliberate to do this *after* resetting?

bcantrill (approved):
Looks great!

cmd/flash/src/lib.rs
mkeeter: Yup, done

PR #230, opened 2022-09-14
Use tlvc-text for packing and dumping of TLV-C

### LINKED ISSUE #229: Use `tlvc-text` for packing and dumping of TLV-C
Right now, we've got (yet another) independent TLV-C implementation [here](https://github.com/oxidecomputer/humility/blob/2f60b05c467a8fd4b32856674515468798190065/cmd/vpd/src/lib.rs#L163-L258).

Once https://github.com/oxidecomputer/tlvc/pull/2 is merged, we should use those canonical functions instead.

bcantrill:
Fixes #229.

### DIFF
--- a/cmd/vpd/src/lib.rs
+++ b/cmd/vpd/src/lib.rs
@@ -214,9 +214,7 @@ fn vpd_write(
 
         tlvc_text::pack(&p)
     } else {
-        let mut bytes = vec![];
-        bytes.resize_with(1024, || 0xffu8);
-        bytes
+        vec![0xffu8; 1024]
     };
 
     let mut all_ops = vec![];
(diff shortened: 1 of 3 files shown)


cmd/vpd/src/lib.rs
```diff
@@ -304,21 +205,19 @@ fn vpd_write(
...
 
-        for piece in p {
-            bytes.extend(pack(&piece));
-        }
+        tlvc_text::pack(&p)
     } else {
+        let mut bytes = vec![];
         bytes.resize_with(1024, || 0xffu8);
-    }
+        bytes
```
mkeeter, line 219: Perhaps replace the whole `else` block with
```rust
    } else {
        vec![0xffu8; 1024]
    }
```

mkeeter (approved):

cmd/vpd/src/lib.rs
bcantrill: Duh -- thanks!

Issue #233, opened 2022-09-15
Duplicate type definitions cause hiffy failures


jgallagher:
I created a hubris build that includes two idol clients of the `gimlet-hf` task (`gimlet-seq` and `mgmt-gateway`). When attempting to call host flash functions via hiffy, I ran into a duplicate type definition error:

```
$ pfexec humility hiffy -c HostFlash.hash -a address=0,len=$((16<<20)) -T 60000
humility hiffy failed: no type for HostFlash.hash: Result { ok: AttributedTy { ty: Ty("[u8; crate::SHA256_SZ]"), recv: FromBytes }, err: CLike(Ty("HfError")) } (as basetype: "expected [u8; crate::SHA256_SZ] to be a basetype"; as enum: "expected enum not found"; as struct: "expected structure not found"; as REPLY: "HostFlash_hash_REPLY matches more than one structure")
```

We have some [basic deduping](https://github.com/oxidecomputer/humility/blob/d336c21c7b6da7f8203a9331c7657581de2bc6ad/humility-core/src/hubris.rs#L4971-L4984) logic in one place, but I'm hitting the same error [in a different place](https://github.com/oxidecomputer/humility/blob/d336c21c7b6da7f8203a9331c7657581de2bc6ad/humility-core/src/hubris.rs#L2746-L2748). A preliminary attempt to extract the deduping logic and call it from both places bumped into an issue with it comparing struct members by equality, because the members of the duplicate structs themselves have different `HubrisGoff`s:

```
HubrisStruct {
    name: "HostFlash_hash_REPLY",
    goff: HubrisGoff {
        object: 10,
        goff: 62343,
    },
    size: 32,
    members: [
        HubrisStructMember {
            offset: 0,
            name: "value",
            goff: HubrisGoff {
                object: 10,
                goff: 63359,
            },
        },
    ],
}
```

```
HubrisStruct {
    name: "HostFlash_hash_REPLY",
    goff: HubrisGoff {
        object: 13,
        goff: 689,
    },
    size: 32,
    members: [
        HubrisStructMember {
            offset: 0,
            name: "value",
            goff: HubrisGoff {
                object: 13,
                goff: 21219,
            },
        },
    ],
}
```

Humility dump from the above: [duplicate_HostFlash_hash_REPLY.dump.tar.gz](https://github.com/oxidecomputer/humility/files/9578233/duplicate_HostFlash_hash_REPLY.dump.tar.gz)

PR #234, opened 2022-09-15
do not merge: hacky workaround for #233

jgallagher:
Instead of doing the right thing and deep comparing the types of struct members, we only compare their offset and name (yikes!). Do not merge, but in limited cases this branch may be useful until #233 is fixed for real.


bcantrill:
Given that we have indeed fixed #233 for real, closing this PR

PR #243, opened 2022-10-07
Widen `humility tasks` TASK column to 20 chars

jgallagher:
I want to name a task `control-plane-agent`, which at 19 characters was previously truncated. The full `humility tasks` output will still fit into 80 columns if the `STATE` is <= 42 chars wide.


jgallagher:
All changes to the tests stdout files in 7e0d699 are whitespace only: https://github.com/oxidecomputer/humility/pull/243/commits/7e0d699f8f11b67ae10bab137285fe02ed5b95e5?diff=unified&w=1

bcantrill (approved):

PR #247, opened 2022-10-14
Handle --list-targets properly

steveklabnik:
It turns out that this is handled in the creation of the ExecutionContext, and so we have to do that earlier than we did before. This also means cloning the Cli.

I'd love to make this cleaner, but this at least fixes the regression, and adds a test.

### DIFF
--- a/tests/cmd/list-targets/env.json
+++ b/tests/cmd/list-targets/env.json
@@ -1,0 +1,51 @@
+{
+    "sn4": {
+        "description": "Gimlet Rev B, serial number 4",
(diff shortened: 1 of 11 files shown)


tests/cmd/list-targets/env.json
```diff
@@ -0,0 +1,51 @@
+{
```
steveklabnik, line 1: i don't know if this is the env file we want in our test suite, but it's the one @rmustacc gave me, and it works, so...

bcantrill (approved):
Looks great -- thank you!

PR #248, opened 2022-10-18
Fix spurious re-builds.

### LINKED ISSUE #245: `cargo run` rebuilds `humility-cmd-doc` and therefore humility every time
Not sure what the cause is, but

```console
C:\Users\steve\Documents\Oxide\humility〉cargo run -- -V                                                                 10/14/2022 02:09:12 PM   Compiling humility-cmd-doc v0.1.0 (C:\Users\steve\Documents\Oxide\humility\cmd\doc)
   Compiling humility v0.9.0 (C:\Users\steve\Documents\Oxide\humility)
    Finished dev [unoptimized + debuginfo] target(s) in 6.76s
     Running `target\debug\humility.exe -V`
humility 0.9.0
C:\Users\steve\Documents\Oxide\humility〉cargo run -- -V                                                                 10/14/2022 02:09:28 PM
   Compiling humility-cmd-doc v0.1.0 (C:\Users\steve\Documents\Oxide\humility\cmd\doc)
   Compiling humility v0.9.0 (C:\Users\steve\Documents\Oxide\humility)
    Finished dev [unoptimized + debuginfo] target(s) in 6.59s
     Running `target\debug\humility.exe -V`
humility 0.9.0
```

steveklabnik:
First of all, this patch removes a line in build.rs; I believe that this is vestigial in #118 from when a different strategy was used to build docs. In any case, I don't remember why I added it exactly, and with no other context... let's just remove that.

The main change here is to move the repl's package from main.rs to lib.rs. Why, you ask? Well, the documentation command's build.rs emits

    println!(
        "cargo:rerun-if-changed={}",
        path.parent().unwrap().join("src/lib.rs").display()
    );

for every package. This makes sense, as every single package is a library.

... except, of course, for the repl package. It is entirely a dummy package, purely to get those docs from. I happened to make it as a binary instead of a library, probably beacuse that's what `cargo new` defaults to.

By simply moving it into lib.rs and removing the un-used main function, everything now builds properly.

Fixes #245.

Thanks to @mkeeter and @cbiffle for poking at this. And special thanks to @jgallagher, who found the smoking gun.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -13,6 +13,9 @@ fn main() -> Result<()> {
     use cargo_metadata::MetadataCommand;
     let mut cmds = BTreeSet::new();
 
+    // MetadataCommand doesn't emit this, so we should
+    println!("cargo:rerun-if-changed=Cargo.toml");
+
     let metadata =
         MetadataCommand::new().manifest_path("./Cargo.toml").exec().unwrap();
 
@@ -59,8 +62,6 @@ fn dcmds() -> Vec<CommandDescription> {{
     }
 
     write!(output, "    ]\n}}")?;
-
-    println!("cargo:rerun-if-changed=Cargo.toml");
 
     Ok(())
 }
--- a/cmd/repl/src/lib.rs
+++ b/cmd/repl/src/lib.rs
@@ -12,7 +12,7 @@
 //! `humility repl` takes the same top level arguments as any other subcommand, and will remember them
 //! inside of the prompt. For example:
 //!
-//! ```
+//! ```console
 //! $ humility -a ../path/to/hubris/archive.zip repl
 //! humility: attached via ST-Link V2-1
 //! Welcome to the humility REPL! Try out some subcommands, or 'quit' to quit!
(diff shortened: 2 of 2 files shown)


mkeeter (approved):

build.rs
```diff
@@ -60,7 +60,5 @@ fn dcmds() -> Vec<CommandDescription> {{
 
     write!(output, "    ]\n}}")?;
 
-    println!("cargo:rerun-if-changed=Cargo.toml");
```
mkeeter, line 63: Is it legit to remove this?  It doesn't look like `MetadataCommand::new().manifest_path("./Cargo.toml").exec().unwrap()` prints `rerun-if-changed`, so this line might still be needed.

In fact, we might want to move this line above the `MetadataCommand::new()` call, to ensure that `build.rs` reruns on `Cargo.toml` changes even if it errors out sometime _after_ reading `Cargo.toml` (avoiding stuff like https://github.com/oxidecomputer/hubris/pull/836)

steveklabnik: I'm open to that! I couldn't remember why I added it, and so I thought I should remove it. Happy to move it up instead; I do think that's the right call, and I'll add a comment so I remember why next time.

mkeeter (approved):

steveklabnik:
Ahaha, because it was a binary, the "doc test" didn't run. Now it does, and fails because I forgot to add the proper language annotation to the fenced code block.

PR #252, opened 2022-10-26
Improve PMBus handling for PSC

mkeeter:
When bringing up the PSC firmware, I ran into a few issues.

Both Hubris and Humility assume that for an I2C device with a `power = { rails = [ ... ] }` field, all sensors (voltage / current / temperature / etc) correspond one-to-one with power rails.

This is problematic for the power shelves, which have two rails (54V and 12V), two fans (uncorrelated to voltage rails), and _three_ temperature sensors (also unrelated to rails!).

This PR adds support for an optional `sensors = [ ... ]` field within power, e.g.
```toml
[[config.i2c.devices]]
bus = "backplane"
name = "psu0mcu"
address = 0b1011_000
device = "mwocp68"
description = "PSU 0 MCU"
power = { rails = [ "V54_PSU0", "V12_PSU0" ], sensors = ["voltage", "current"] }
sensors = { voltage = 2, current = 2, temperature = 3, speed = 2 }
```

The `sensors` field indicates that only `voltage` and `current` correspond one-to-one with rails; other sensor types are treated as independent.

If the `sensors` field is blank (i.e. all of our existing TOML files), then every sensor corresponds one-to-one with rails, matching the current behavior.

In addition, `humility pmbus --summarize` is modified to not `bail!` on an error; instead, it prints the error and continues.  This makes working with the PSC nicer, because not all 6 PSUs will be populated at all times.

Screen Shot 2022-10-26 at 3 13 52 PM

### DIFF
--- a/humility-core/src/hubris.rs
+++ b/humility-core/src/hubris.rs
@@ -116,6 +116,14 @@ struct HubrisConfigI2cPower {
     rails: Option<Vec<String>>,
     #[serde(default = "HubrisConfigI2cPower::default_pmbus")]
     pmbus: bool,
+
+    /// Lists which sensor types have a one-to-one association with power rails
+    ///
+    /// When `None`, we assume that all sensor types are mapped one-to-one with
+    /// rails.  Otherwise, *only* the listed sensor types are associated with
+    /// rails (which is the case in systems with independent temperature sensor
+    /// and power rails).
+    sensors: Option<Vec<HubrisSensorKind>>,
 }
 
 impl HubrisConfigI2cPower {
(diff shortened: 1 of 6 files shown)


humility-core/src/hubris.rs
```diff
@@ -2228,10 +2238,15 @@ impl HubrisArchive {
                         bail!("sensor count exceeds rails for {:?}", d);
                     }
                 }
-            }
-
-            if let Some(power) = &d.power {
-                if let Some(rails) = &power.rails {
+            } else if d.power.is_some()
+                && d.power
+                    .as_ref()
```
bcantrill, line 2243: /me shakes fist at https://github.com/rust-lang/rustfmt/issues/4306

bcantrill (approved):
Looks great!

PR #253, opened 2022-11-02
Add `--halt` option to reset

labbott:
It's useful to be able to halt immediately after reset. Doing so requires that we know are attached to a specific chip. Bring some of the logic that we have for chip detection in flashing into core hubris so it can be used elsewhere.

### DIFF
--- a/humility-core/src/hubris.rs
+++ b/humility-core/src/hubris.rs
@@ -275,11 +275,42 @@ impl HubrisSensorKind {
 }
 
 //
+// This is the Hubris definition
+//
+#[derive(Debug, Deserialize)]
+pub enum FlashProgram {
+    PyOcd(Vec<FlashArgument>),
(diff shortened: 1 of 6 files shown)


humility-core/src/hubris.rs
```diff
@@ -274,12 +274,43 @@ impl HubrisSensorKind {
...
+#[derive(Debug, Deserialize)]
+pub enum FlashArgument {
+    Direct(String),
+    Payload,
+    FormattedPayload(String, String),
+    Config,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct HubrisRawFlash {
```
mkeeter, line 301: Changing the name here (from `FlashConfig`) is a little confusing – maybe `HubrisFlashConfig` instead?

`HubrisRawFlash` makes me think of the _contents_ of flash, rather than a configuration.

PR #254, opened 2022-11-03
Handle new RegionDesc (reference vs index)

mkeeter:
https://github.com/oxidecomputer/hubris/pull/897 broke lots of Humility subcommands, because the `regions` table changed from `[u8]` to `[&'static RegionDesc]`.

This PR fixes stuff, while maintaining backwards compatibility.


bcantrill (approved):
I think my only request would be that we bump the version number so we at least know where this change has been rolled out and not. I also think it would be helpful to add at least one new dump to the test suite that has the new region descriptors -- which should be pretty straightforward.

mkeeter:
Done, queued up for automerge!

PR #255, opened 2022-11-07
Parse auxflash size from TOML file

mkeeter:
This fixes an issue where Humility would assume the wrong auxflash size.

Note that there was a brief time period (September 13-15) where the
auxflash size both *was not* in the TOML file **and** was 1 MiB (rather
than the 2 MiB default now in Humility); archives from that narrow window will
probably emit errors upon use.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -992,7 +992,7 @@ dependencies = [
 
 [[package]]
 name = "humility"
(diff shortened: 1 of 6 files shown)


arjenroodselaar (approved):
Nice, thanks!

PR #257, opened 2022-11-09
Don't die on invalid input

### LINKED ISSUE #256: `humility repl` exits on bad subcommands
`humility repl` doesn't handle the case where the humility command is valid but the clap subcommand is not

```
Welcome to the humility REPL! Try out some subcommands, or 'quit' to quit!
humility> isp asdf
error: Found argument 'asdf' which wasn't expected, or isn't valid in this context

USAGE:
    humility isp [OPTIONS] --port <port> <SUBCOMMAND>

[labbott@labbott-the-desktop humility]$ 
```

steveklabnik:
I previously had attempted to special case the help error, and died on the rest. Of course, this means when we're in the repl, we'd exit out, which is not intended behavior.

Instead we now just print and return no matter what.

FIxes #256

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,7 +21,7 @@ mod cmd_repl;
 fn main() -> Result<()> {
     let (commands, m, args) = match parse_args(&mut std::env::args_os()) {
         Some(s) => s,
-        None => std::process::exit(0),
+        None => std::process::exit(1),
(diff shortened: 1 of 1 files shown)


labbott (approved):
LGTM, thanks for fixing!

PR #267, opened 2022-11-28
Add support for TOML inheritance

mkeeter:
This reads and applies `patches.toml`, which will be shipped in archives after https://github.com/oxidecomputer/hubris/pull/957 (for builds using TOML inheritance).

In addition, it adds `name => ...` to `humility manifest`, which seems helpful to distinguish between multiple builds targeting the same board.  This required updating many test `stdout`'s.


bcantrill (approved):
Looks terrific -- and thanks for adding the new tests!

PR #269, opened 2022-11-29
bump patch version

bcantrill:
#267 is required for `humility manifest` to correctly print task features on archives using TOML inheritance; it's worth bumping the version so folks know to get on the correct one.


jgallagher:
#266 (now merged, but wasn't when this PR was opened) will be required to report psc READ_VIN / READ_IIN measurements, which can also sneak into this version bump.

PR #277, opened 2022-12-02
Add `humility dashboard` support for Sidecar

mkeeter:
The Sidecar sequencer has a slightly different API!

This PR has us try both APIs, picking the first one that's available, and handles translation from Gimlet states -> Tofino policy.

### DIFF
--- a/cmd/dashboard/src/lib.rs
+++ b/cmd/dashboard/src/lib.rs
@@ -12,7 +12,7 @@
 //! graphed by the dashboard.
 //!
 
-use anyhow::{bail, Result};
+use anyhow::{anyhow, bail, Result};
 use clap::Command as ClapCommand;
 use clap::{CommandFactory, Parser};
 use crossterm::{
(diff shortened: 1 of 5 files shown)


cmd/dashboard/src/lib.rs
```diff
@@ -794,7 +794,22 @@ fn sequencer_state_ops<'a>(
     ops: &mut Vec<Op>,
 ) -> Result<idol::IdolOperation<'a>> {
     let funcs = context.functions()?;
-    let op = idol::IdolOperation::new(hubris, "Sequencer", "get_state", None)?;
+    let op = idol::IdolOperation::new(hubris, "Sequencer", "get_state", None)
```
jgallagher, line 797: Double-checking I understand: this is keying off the existence of a `Sequencer.get_state` operation (if present, assume a gimlet), then a `Sequencer.tofino_seq_state` operation (if present, assume a sidecar), and it would break if the sidecar sequencer added a `get_state` operation?

If that's right, this seems okay, but a couple thoughts:

* Maybe swap the order of the checks? It's hard to imagine the gimlet sequencer gaining a `tofino_seq_state` operation, but sidecar gaining `get_state` seems less unreasonable.
* Should we change the idol interface names so they don't both use `Sequencer`?

mkeeter: Changing the order is a good idea; done!

Renaming stuff is... probably out of scope for this PR 😅

jgallagher (approved):

PR #279, opened 2022-12-05
isp: Get CMPA and CFPA from humility archive

flihp:
This is a quality of life improvement for device programming.


labbott (approved):
If it makes programming easier I'm all for it. We can always fall back to `write-memory` for any other hacks that are needed.

PR #280, opened 2022-12-05
Fix enum decoding

mkeeter:
The compiler has gotten smarter, and is better at reusing `enum` discriminator fields!

Considering the following DWARF data:
```
0x0002f740: DW_TAG_structure_type
              DW_AT_name	("Trace")
              DW_AT_byte_size	(0x10)
              DW_AT_alignment	(8)

0x0002f747:   DW_TAG_variant_part
                DW_AT_discr	(0x0002f74c)

0x0002f74c:     DW_TAG_member
                  DW_AT_type	(0x0000070d "u8")
                  DW_AT_alignment	(1)
                  DW_AT_data_member_location	(0x00)
                  DW_AT_artificial	(true)

0x0002f753:     DW_TAG_variant
                  DW_AT_discr_value	(0x2f)

0x0002f755:       DW_TAG_member
                    DW_AT_name	("None")
                    DW_AT_type	(0x0002f78b "task_monorail_server::bsp::Trace::None")
                    DW_AT_alignment	(8)
                    DW_AT_data_member_location	(0x00)

0x0002f760:       NULL

0x0002f761:     DW_TAG_variant
                  DW_AT_discr_value	(0x30)

0x0002f763:       DW_TAG_member
                    DW_AT_name	("FrontIoSpeedChange")
                    DW_AT_type	(0x0002f792 "task_monorail_server::bsp::Trace::FrontIoSpeedChange")
                    DW_AT_alignment	(8)
                    DW_AT_data_member_location	(0x00)

0x0002f76e:       NULL

0x0002f76f:     DW_TAG_variant

0x0002f770:       DW_TAG_member
                    DW_AT_name	("AnegCheckFailed")
                    DW_AT_type	(0x0002f7bb "task_monorail_server::bsp::Trace::AnegCheckFailed")
                    DW_AT_alignment	(8)
                    DW_AT_data_member_location	(0x00)

0x0002f77b:       NULL

0x0002f77c:     DW_TAG_variant
                  DW_AT_discr_value	(0x32)

0x0002f77e:       DW_TAG_member
                    DW_AT_name	("Restarted10GAneg")
                    DW_AT_type	(0x0002f7ce "task_monorail_server::bsp::Trace::Restarted10GAneg")
                    DW_AT_alignment	(8)
                    DW_AT_data_member_location	(0x00)

0x0002f789:       NULL

0x0002f78a:     NULL
```

Notice that the discriminants are in the range 47-50, and `Trace::AnegCheckFailed` **does not have** a discriminant; this is because `AnegCheckFailed` contains a `VscError`, which uses discriminants 0-46 in that same byte position.

How clever!

Our code assumed that only single-value `enum`s would have no discriminant, and selected them immediately when seen.  This is no longer a valid assumption!

(in fact, it turns out it wasn't a valid assumption even before the compiler bump; this is why `stderr` for some very old test archives no longer prints an error message!)


bcantrill (approved):
This looks great -- thank you!  Is it worth also adding code to die on ingest if we see that two variants in an enum are missing a tag? (Happy to do this as a separate PR.)

mkeeter:
That makes sense, but I'm not exactly sure where in the ingestion process that should go – I'll open a ticket for you to do as a follow-up (https://github.com/oxidecomputer/humility/issues/281)

PR #286, opened 2022-12-07
Add support for hubpack-encoded idol arguments

jgallagher:
hubpack is _mostly_ a drop-in replacement for ssmarshal: for types that ssmarshal supports, hubpack encodes them the same way. However, for types that ssmarshal doesn't support (e.g., enums that aren't `#[repr(C)]`), hubpack has a well-defined encoding based on the source code order of the variants. Without the `#[repr(C)]` requirement, we cannot assume a variant's hubpack encoding is the same as its rustc DWARF tag. We _do_ continue to assume that the order of variants in the DWARF matches the source code order.

All that said, this PR has 3 main changes:

1. `lookup_variant()` has been renamed `lookup_variant_by_tag()`, and a new function `lookup_variant_by_index()` has been added.
2. In the couple of places where we're dealing with ssmarshal or hubpack encoded variants, we no longer assume variant tagging matches, and use variant indices instead.
3. All of humility itself's uses of `ssmarshal` are replaced with `hubpack`.


andrewjstone (approved):
LGTM. Maybe add some comments relating to the fact that dwarf ordering only matches serialization for derived types, and also a link to section 5.7.10 of the DWARF 5 spec that says enum variants do actually match source ordering, so the assumptions should work fine.

mkeeter (approved):

jgallagher:
> LGTM. Maybe add some comments relating to the fact that dwarf ordering only matches serialization for derived types, and also a link to section 5.7.10 of the DWARF 5 spec that says enum variants do actually match source ordering, so the assumptions should work fine.

Thanks - added in 3972eacc0e1123ebc0e515b2f1ce2eaeea213293

PR #287, opened 2022-12-08
Add `readvar --as-c-string` to pretty-print arrays that contain strings

### LINKED ISSUE #285: humility readvar: Can we interpret arrays as strings?
If I have a static array of significant size (e.g., a `[u8; 4096]`) that is meant to hold a human-readable message, it would be nice if I could ask `humility readvar` to attempt to interpret its contents as a string. A couple thoughts on possible implementations:

* We could use `String::from_utf8_lossy()` to replace any non-utf8 sequences
* We could treat the contents as a C string: find the first `0x00` and then `str::from_utf8()` everything up to that point

jgallagher:
Now instead of this:

```
$ pfexec ./humility readvar LAST_HOST_BOOT_FAIL
LAST_HOST_BOOT_FAIL (0x2401c10d) = MaybeUninit<[u8; 4096]> {
    value: [
        0x43,
        0x6f,
        0x75,
        0x6c,
        0x64,
        0x20,
        0x6e,
        0x6f,
        0x74,
        0x20,
        0x66,
        0x69,
        0x6e,
        0x64,
        0x20,
        0x61,
        0x20,
        0x76,
        0x61,
        0x6c,
        0x69,
        0x64,
        0x20,
        0x70,
        0x68,
        0x61,
        0x73,
        0x65,
        0x32,
        0x20,
        0x69,
        0x6d,
        0x61,
        0x67,
        0x65,
        0x20,
        0x6f,
        0x6e,
        0x20,
        0x64,
        0x69,
        0x73,
        0x6b,
        0x3a,
        0x31,
        0x0,
        0x0,
        0x0,
        ... snip 4000+ more 0x00s ...
    ]
}
```

we can do this:

```
$ pfexec ./humility readvar LAST_HOST_BOOT_FAIL --as-c-string
LAST_HOST_BOOT_FAIL (0x2401c10d) = MaybeUninit<[u8; 4096]> {
    value: "Could not find a valid phase2 image on disk:1"
}
```

Fixes #285.

### DIFF
--- a/humility-core/src/reflect.rs
+++ b/humility-core/src/reflect.rs
@@ -599,6 +599,23 @@ impl std::ops::Deref for Array {
     }
 }
 
+impl Array {
+    fn format_as_c_string(&self, out: &mut dyn std::io::Write) -> Result<()> {
+        let mut bytes = Vec::new();
+        for v in &self.0 {
+            let b = v.as_base().ok().and_then(Base::as_u8).context(
+                "cannot interpret array of non-u8 values as a C string",
+            )?;
+            if b == 0 {
+                break;
+            }
+            bytes.push(b);
+        }
+        write!(out, "{:?}", String::from_utf8_lossy(&bytes))?;
+        Ok(())
+    }
+}
+
 impl Format for Array {
     fn format(
         &self,
@@ -606,6 +623,9 @@ impl Format for Array {
         mut fmt: HubrisPrintFormat,
         out: &mut dyn std::io::Write,
     ) -> Result<()> {
+        if fmt.interpret_as_c_string {
+            return self.format_as_c_string(out);
+        }
         fmt.indent += 4;
         fmt.no_name = false;
 
--- a/cmd/readvar/src/lib.rs
+++ b/cmd/readvar/src/lib.rs
@@ -50,6 +50,10 @@ struct ReadvarArgs {
     /// values in decimal instead of hex
     #[clap(long, short)]
     decimal: bool,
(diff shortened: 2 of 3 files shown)


humility-core/src/reflect.rs
```diff
@@ -599,13 +599,33 @@ impl std::ops::Deref for Array {
...
+        let mut bytes = Vec::new();
+        for v in &self.0 {
+            let b = v.as_base().ok().and_then(Base::as_u8).context(
+                "cannot interpret array of non-u8 values as a C string",
+            )?;
+            if b == 0 {
+                break;
+            }
+            bytes.push(b);
+        }
```
mkeeter, line 613: You _could_ accumulate a `Result<Vec<u8>, _>` and make this immutable, e.g.
```suggestion
        let bytes = self
            .0
            .iter()
            .map(|b| {
                b.as_base().ok().and_then(Base::as_u8).context(
                    "cannot interpret array of non-u8 values as a C string",
                )
            })
            .take_while(|b| b.as_ref().map(|b| *b != 0).unwrap_or(true))
            .collect::<Result<Vec<u8>, _>>()?;
```

Not sure if you want to, though!

mkeeter (approved):

humility-core/src/reflect.rs
jgallagher: I don't think I do, hah. The `.take_while()` line seems overly dense for the operation it's doing ("stop at the first 0").

PR #288, opened 2022-12-09
Add `console-proxy` command

jgallagher:
`humility console-proxy` uses the uart control functions added in https://github.com/oxidecomputer/hubris/pull/1005 to do its best (very slow) picocom impersonation.

~~Before merging:~~

- [x] ~~don't try to build this command (or make it report "this isn't supported") on Windows~~
- [x] ~~don't allow this command to be used within the repl on non-Windows~~


Cargo.lock
```diff
@@ -1043,7 +1043,7 @@ dependencies = [
 
 [[package]]
 name = "humility"
-version = "0.9.13"
+version = "0.9.14"
```
mkeeter, line 1046: This will conflict with https://github.com/oxidecomputer/humility/pull/289, so we'll see who gets merged first

Cargo.lock
```diff
@@ -1173,6 +1174,23 @@ dependencies = [
  "tlvc",
 ]
 
+[[package]]
+name = "humility-cmd-console-proxy"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "clap 3.2.21",
```
mkeeter, line 1182: Unrelated to this PR, but seeing a version number made me realize that we have an unused `structopt` dependency that's dragging in a _different_ `clap` version.  I opened #292 to fix that.

cmd/console-proxy/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+humility = { workspace = true }
+humility-cmd = { workspace = true }
+cmd-hiffy = { workspace = true }
+
+hif = { workspace = true }
+idol = { workspace = true }
+
+[target.'cfg(not(windows))'.dependencies]
+termios = "0.3"
```
mkeeter, line 22: Is there a way to make this a conditional workspace dependency?

jgallagher: Maaaaybe? I don't think we can have target-specific workspace dependencies (e.g., https://github.com/rust-lang/cargo/issues/5220), but maybe we can specify `termios = "0.3"` as a workspace dep but conditionally depend on it in Cargo.toml. Trying that in 5db1630 to see if it passes Windows CI.

cmd/console-proxy/src/posix/picocom_map.rs
```diff
@@ -0,0 +1,185 @@
...
+                        self.prev.extend(repl);
+                        continue;
+                    } else {
+                        return Some(raw::LF);
+                    }
+                }
+                raw::BS if self.rules.bsdel => return Some(raw::DEL),
+                raw::DEL if self.rules.delbs => return Some(raw::BS),
+                b => return Some(b),
+            }
```
mkeeter, line 118: Nebulous vibes-based refactoring – it feels cleaner to _always_ push stuff to the ringbuf, even if we're going to pop it right away.  This lets us make the loop conditional on `is_empty` and move `pop_front` out of the loop entirely.
```suggestion
        // Do we have any leftover bytes from a previous input byte
        // (possibly from a previous iteration of this loop)?
        while self.prev.is_empty() {
            match self.inner.next()? {
                raw::CR if self.rules.cr.is_some() => {
                    self.prev.extend(self.rules.cr.unwrap());
                }
                raw::LF if self.rules.lf.is_some() => {
                    self.prev.extend(self.rules.lf.unwrap());
                }
                raw::BS if self.rules.bsdel => self.prev.push_back(raw::DEL),
                raw::DEL if self.rules.delbs => self.prev.push_back(raw::BS),
                b => self.prev.push_back(b),
            }
        }
        self.prev.pop_front()
```

What do you think – too clever?

cmd/console-proxy/src/posix/picocom_map.rs
```diff
@@ -0,0 +1,185 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! picocom-style character remapping; does not support the "... to hex" rules.
+
+use std::{collections::VecDeque, str::FromStr};
+
+use anyhow::{bail, ensure, Error, Result};
+
```
mkeeter, line 10: Nit: move `mod raw` up here, so I'm not confused where those values come from in the table below.

cmd/console-proxy/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// `control_plane_agent` task, setting the client back to MGS.
+    ///
+    /// This has the side effect of putting the SP back into its default mode in
+    /// which it constantly reads from the console uart, discarding data as
+    /// needed (or forwarding to MGS, if one is attached). While Humility is
+    /// attached as the client, that behavior is disabled and hardware flow
+    /// control is used to apply backpressure to the host.
+    Detach,
+
+    /// Print the current console client (Humility or Mgs).
```
mkeeter, line 105:
```suggestion
    /// Print the current console client (Humility or MGS).
```

cmd/console-proxy/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+        )]
+        raw: bool,
+
+        #[clap(
+            long,
+            help = "Specifies the input character map (i.e., special characters to be replaced when reading from the serial port). See picocom's manpage."
+        )]
+        imap: Option<String>,
+
+        #[clap(
```
mkeeter, line 82: I always find myself having to look up how to invoke `picocom` (`--imap lfcrlf --omap crlf,delbs`).  Should we make those the default?

jgallagher: If I make those the defaults, how does one run this with no remappings enabled? (I'll admit it doesn't seem likely one would want to, but...)

cmd/console-proxy/src/posix.rs
```diff
@@ -0,0 +1,401 @@
...
+
+            let mut done = false;
+            loop {
+                select! {
+                    recv(stdin_rx) -> data => {
+                        if let Ok(mut data) = data {
+                            tx_buf.append(&mut data);
+                        } else {
+                            done = true;
+                            break;
```
mkeeter, line 166: This could be done with a loop label instead of a flag, e.g. `'outer: loop` and `break 'outer;`

cmd/console-proxy/Cargo.toml
jgallagher: I'd lean toward making them all workspace deps, because otherwise if I want to add a new dep, I first ought to grep and see if it's used by something else, and if so, move it to be a workspace dep now that it's becoming shared.

That's slightly annoying but not the end of the world; are there similar annoyances with "everything at the workspace level" I'm not thinking of?

cmd/console-proxy/src/posix.rs
```diff
@@ -0,0 +1,401 @@
...
+        let mut tx_buf = Vec::new();
+        loop {
+            if !tx_buf.is_empty() {
+                let nwritten = self.write(&tx_buf)?;
+                tx_buf.drain(0..nwritten);
+            }
+
+            let nread = self.read(&mut rx_buf)?;
+            if nread > 0 {
+                stdout_tx.send(rx_buf[..nread].to_owned()).unwrap();
```
mkeeter, line 145: Why does this need to be in a separate thread, versus writing `rx_buf[..nread]` to `stdout` right here?

cmd/console-proxy/src/posix.rs
```diff
@@ -0,0 +1,401 @@
...
+
+        let (stdin_tx, stdin_rx) = crossbeam_channel::unbounded();
+        let (stdout_tx, stdout_rx) = crossbeam_channel::unbounded();
+        thread::spawn(move || stdin_reader(raw, stdin_tx, omap));
+        thread::spawn(move || stdout_writer(stdout_rx, imap, log));
+
+        let mut rx_buf = vec![0; HIFFY_BUF_SIZE];
+        let mut tx_buf = Vec::new();
+        loop {
+            if !tx_buf.is_empty() {
```
mkeeter, line 138: Would there be any advantage to looping here until `tx_buf` is empty, rather than only doing one write?

cmd/console-proxy/src/posix.rs
```diff
@@ -0,0 +1,401 @@
...
+        let context = HiffyContext::new(hubris, core, hiffy_timeout)?;
+        Ok(Self {
+            hubris,
+            core,
+            context,
+            poll_interval: Duration::from_millis(u64::from(poll_interval)),
+        })
+    }
+
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
```
mkeeter, line 56: Nit: rename to `uart_read` to make it obvious in `attach`

(same with `fn write` below)

cmd/console-proxy/src/posix.rs
jgallagher: Yes, although I'm slightly worried about deadlocking (e.g., if we're trying to send until `tx_buf` is empty, but the console uart rx fifo is full and blocked waiting for us to receive). Maybe that's not possible with a well behaving host; not sure.

Alternatively: there's a check below to shorten the sleep time if we just had a full-length read; I could make that "if we had a full-length read or we still have data to transmit". I think that would be nearly equivalent to adding the loop here, except that we pause momentarily to check for incoming data after each write.

cmd/console-proxy/src/posix.rs
```diff
@@ -0,0 +1,401 @@
...
+                        .append(true)
+                        .create(true)
+                        .open(&path)
+                        .with_context(|| {
+                            format!("failed to open {}", path.display())
+                        })?;
+                    Some(f)
+                }
+                None => None,
+            };
```
mkeeter, line 389:
```suggestion
            let log = log
                .map(|path| {
                    File::options()
                        .append(true)
                        .create(true)
                        .open(&path)
                        .with_context(|| {
                            format!("failed to open {}", path.display())
                        })
                })
                .transpose()?;
```

cmd/console-proxy/src/lib.rs
mkeeter: Would `--imap ""` disable them?  Or a separate `--no-imap`?

cmd/console-proxy/src/posix.rs
mkeeter: Agreed, the latter seems like a good idea!

cmd/console-proxy/src/lib.rs
jgallagher: Ah yeah `--imap ""` would work. Kinda weird, but I don't actually expect anyone to need to do that. I'll change this.

cmd/console-proxy/src/posix.rs
jgallagher: I don't think it does! When I originally wrote this I was thinking that we'd offload I/O to other threads so we could quickly service repeated reads/writes, but, uh, yeah, not so much. I'll get rid of the stdout thread.

cmd/console-proxy/src/posix/picocom_map.rs
jgallagher: I don't think so; combined it with making `rules.{cr,lf}` non-optional to avoid the unwrap in 385d2d2

cmd/console-proxy/src/posix.rs
jgallagher: TIL `transpose()` - thanks!

Cargo.lock
jgallagher: #289 landed first, so I updated to 0.9.15

PR #289, opened 2022-12-12
add error reporting to "humility sensors"

bcantrill:
Draft commit message:

```
This work is the Humility side of oxidecomputer/hubris#1013: it adds
an option to "humility i2c" to get at the last selected mux/segment
(--lastmux) as well as an expansion of "humility sensors" to dump
error counters.

Beyond this, it fixes some little issues: 

- "humility probe" could previously panic on a mismatched archive
- "humility sensors -s" would print data less frequently than 1 Hz
- "humility sensors" should print sensor IDs in hex as well as decimal

```


cmd/sensors/src/lib.rs
```diff
@@ -120,7 +123,8 @@ fn list(
                 };
 
                 println!(
-                    "{:2} {:7} {:2} {:2} {:3} 0x{:02x} {:13} {:<1}",
+                    "{:3} 0x{:03x} {:7} {:2} {:2} {:3} 0x{:02x} {:13} {:<1}",
```
mkeeter, line 126: Use `#` and the formatter will automatically add `0x`
```suggestion
                    "{:3} {:#05x} {:7} {:2} {:2} {:3} {:#04x} {:13} {:<1}",
```

cmd/sensors/src/lib.rs
```diff
@@ -133,13 +137,14 @@ fn list(
             }
             HubrisSensorDevice::Other(device, _) => {
                 println!(
-                    "{:2} {:7} {:2} {:2} {:3}   {:02} {:13} {:<1}",
+                    "{:3} 0x{:03x} {:7} {:2} {:2} {:3} {:02} {:13} {:<1}",
```
mkeeter, line 140:
```suggestion
                    "{:3} {:#05x} {:7} {:2} {:2} {:3} {:02} {:13} {:<1}",
```

mkeeter (approved):

cmd/sensors/src/lib.rs
bcantrill: Great tip! And allows me to lose the leading zeroes, which were somewhat unsightly...

PR #291, opened 2022-12-12
288 enhancements

steveklabnik:
Two commits to deal with the two problems. Feel free to take what you like, if anything, and discard anything you don't. :)

The second commit is small, and should make a lot of sense. I didn't *use* it to disable things because I can't test it, since I'm purely on Windows. I did check that the state gets set properly, though, so it shouldn't be too hard to do so!

The first commit... I used an inline module here, but I don't *like* it. It may make more sense to make a different one in a new file, and leave it at that, but I figure that is up to you. If you want to do this in a different fashion, that's also your prerogative! Let me know if you want some additional manual testing on Windows to make sure the behaviors work out the way you want after you've got that nailed down.


steveklabnik:
Lol, of course doing the module thing means I broke some stuff on non-Windows, apparently. Should be easy enough to take care of, but I'm gonna leave it as-is since you may not even like the module idea in the first place.

jgallagher:
Thanks! I'm took the second commit as-is onto #288; for the first I took your suggestion to move it to a separate file (31630be35dde1bdbba1d5d267830b746c4888e00). Since these changes have been absorbed into #288 in a way github doesn't understand, I'll close this PR.

PR #292, opened 2022-12-13
Remove structopt (now unused)

mkeeter:


jgallagher (approved):

steveklabnik (approved):
Thank you! I had noticed this the other day but didn't track it down, looks like it was easy enough!

PR #293, opened 2022-12-14
Add `get_idol_command` to `HubrisArchive`, reducing boilerplate

mkeeter:
The most common way to build an `IdolOperation` is `IdolOperation::new(hubris, "Interface", "function", None)?;`

This PR uses an extension trait to add `get_idol_command` to `HubrisArchive`, which handles this common case.


steveklabnik (approved):
Looks like a good cleanup to me!

PR #294, opened 2022-12-15
Duplicate type definitions cause hiffy failures

bcantrill:


humility-core/src/hubris.rs
```diff
@@ -3034,6 +3034,38 @@ impl HubrisArchive {
...
     }
 
+    ///
+    /// Takes a list of potentially similar types and deduplicates the list.
+    ///
+    fn dedup(&self, goffs: Vec<HubrisGoff>) -> Result<Vec<HubrisGoff>> {
+        let mut cmp = Vec::new();
+        let mut out = Vec::new();
+
+        if goffs.len() <= 1 {
```
mkeeter, line 3044: Nit: this conditional isn't strictly necessary, since a 1-item `Vec` would dedup to itself.

humility-core/src/hubris.rs
```diff
@@ -3034,6 +3034,38 @@ impl HubrisArchive {
...
+                    if !cmp[i].1.differs(self, &cmp[j].1)? {
+                        dup = true;
+                        break;
+                    }
+                }
+
+                if !dup {
+                    out.push(cmp[i].0);
+                }
+            }
```
mkeeter, line 3064: We could remove (some) explicit indexing here:
```suggestion
            for (i, a) in cmp.iter().enumerate() {
                let mut dup = false;

                for b in &cmp[i + 1..] {
                    if !a.1.differs(self, &b.1)? {
                        dup = true;
                        break;
                    }
                }

                if !dup {
                    out.push(a.0);
                }
            }
```

cmd/hiffy/src/lib.rs
```diff
@@ -111,49 +115,43 @@ pub enum HiffyLease<'a> {
...
-                    warn!("Mismatch between expected reply and operation");
+            Ok((_, Some(e))) => {
+                match &op.1.reply {
+                    Reply::Result { ok, .. } => {
+                        println!("{}{:<27} {}", margin, "<ok>", ok.ty.0);
+                        println!("{}{:<27} {}", margin, "<error>", e.name);
+                    }
+                    _ => {
+                        warn!("mismatch on reply: expected Reply::Result, found {:?}",
+                    op);
```
jgallagher, line 142: Nit - `cargo fmt` seems to have given up on this line. If I manually change this to

```rust
warn!(
    "mismatch on reply: expected Reply::Result, found {:?}",
    op
);
```

it then kicks back in and does some minor reformatting of the whole `match`.

humility-core/src/hubris.rs
```diff
@@ -3034,6 +3034,38 @@ impl HubrisArchive {
         !self.modules.is_empty()
     }
 
+    ///
+    /// Takes a list of potentially similar types and deduplicates the list.
+    ///
+    fn dedup(&self, goffs: Vec<HubrisGoff>) -> Result<Vec<HubrisGoff>> {
```
mkeeter, line 3040: Looking at how this is used, we could save a few allocations by instead taking an `Iterator<Item = &HubrisGoff>`.

This would also require modifying call sites; here's a patch:
```diff
diff --git a/humility-core/src/hubris.rs b/humility-core/src/hubris.rs
index dd807c0..7a2f143 100644
--- a/humility-core/src/hubris.rs
+++ b/humility-core/src/hubris.rs
@@ -3037,33 +3037,32 @@ impl HubrisArchive {
     ///
     /// Takes a list of potentially similar types and deduplicates the list.
     ///
-    fn dedup(&self, goffs: Vec<HubrisGoff>) -> Result<Vec<HubrisGoff>> {
+    fn dedup<'a, I: Iterator<Item = &'a HubrisGoff>>(
+        &self,
+        goffs: I,
+    ) -> Result<Vec<HubrisGoff>> {
         let mut cmp = Vec::new();
         let mut out = Vec::new();
 
-        if goffs.len() <= 1 {
-            Ok(goffs)
-        } else {
-            for goff in goffs {
-                cmp.push((goff, self.lookup_type(goff)?));
-            }
+        for &goff in goffs {
+            cmp.push((goff, self.lookup_type(goff)?));
+        }
 
-            for i in 0..cmp.len() {
-                let mut dup = false;
+        for i in 0..cmp.len() {
+            let mut dup = false;
 
-                for j in i + 1..cmp.len() {
-                    if !cmp[i].1.differs(self, &cmp[j].1)? {
-                        dup = true;
-                        break;
-                    }
+            for j in i + 1..cmp.len() {
+                if !cmp[i].1.differs(self, &cmp[j].1)? {
+                    dup = true;
+                    break;
                 }
+            }
 
-                if !dup {
-                    out.push(cmp[i].0);
-                }
+            if !dup {
+                out.push(cmp[i].0);
             }
-            Ok(out)
         }
+        Ok(out)
     }
 
     ///
@@ -3075,7 +3074,7 @@ impl HubrisArchive {
     pub fn lookup_struct_byname(&self, name: &str) -> Result<&HubrisStruct> {
         match self.structs_byname.get_vec(name) {
             Some(v) => {
-                let m = self.dedup(v.to_vec())?;
+                let m = self.dedup(v.iter())?;
 
                 if m.len() > 1 {
                     Err(anyhow!("{} matches more than one structure", name))
@@ -5603,12 +5602,8 @@ impl HubrisModule {
     ) -> Result<&'a HubrisStruct> {
         match hubris.structs_byname.get_vec(name) {
             Some(v) => {
-                let m = hubris.dedup(
-                    v.iter()
-                        .filter(|g| g.object == self.object)
-                        .copied()
-                        .collect::<Vec<HubrisGoff>>(),
-                )?;
+                let m = hubris
+                    .dedup(v.iter().filter(|g| g.object == self.object))?;
 
                 if m.len() > 1 {
                     Err(anyhow!("{} matches more than one structure", name))
@@ -5632,12 +5627,8 @@ impl HubrisModule {
     ) -> Result<&'a HubrisEnum> {
         match hubris.enums_byname.get_vec(name) {
             Some(v) => {
-                let m = hubris.dedup(
-                    v.iter()
-                        .filter(|g| g.object == self.object)
-                        .copied()
-                        .collect::<Vec<HubrisGoff>>(),
-                )?;
+                let m = hubris
+                    .dedup(v.iter().filter(|g| g.object == self.object))?;
 
                 if m.len() > 1 {
                     Err(anyhow!("{} matches more than one enum", name))
```

humility-core/src/hubris.rs
```diff
@@ -5235,25 +5306,37 @@ impl HubrisEnum {
...
-    // type information.
-    //
-    pub fn approximately_same(&self, rhs: &HubrisEnum) -> bool {
-        for (l, r) in self.variants.iter().zip(rhs.variants.iter()) {
-            if l.name != r.name || l.offset != r.offset || l.tag != r.tag {
-                return false;
+    fn differs(&self, hubris: &HubrisArchive, rhs: &Self) -> Result<bool> {
+        if self.size != rhs.size
+            || self.discriminant != rhs.discriminant
+            || self.variants.len() != rhs.variants.len()
```
mkeeter, line 5312: Do we need to check `self.tag` here?  I'm not exactly sure what it's doing!

cmd/hiffy/src/lib.rs
```diff
@@ -401,13 +403,19 @@ fn hiffy(context: &mut humility::ExecutionContext) -> Result<()> {
     let subargs = HiffyArgs::try_parse_from(subargs)?;
 
     if subargs.list {
-        hiffy_list(hubris, subargs.verbose)?;
+        hiffy_list(hubris, subargs.filter)?;
         return Ok(());
+    } else if !subargs.filter.is_empty() {
+        bail!("filters can only be provided with --list");
```
jgallagher, line 409: Nit: I don't think we can get here; `filter` already has the clap attribute `requires = "list"` so it should bail for us if we get filters without `--list`.

humility-core/src/hubris.rs
bcantrill: Yeah, that was a (lame) attempt to avoid the unnecessary `lookup_type` in the (common) case of a single element.

humility-core/src/hubris.rs
```diff
@@ -5362,6 +5475,24 @@ impl<'a> HubrisType<'a> {
             }
         }
     }
+
+    pub fn differs(&self, hubris: &HubrisArchive, rhs: &Self) -> Result<bool> {
```
mkeeter, line 5479: Architecture question:

It looks like `HubrisType::differs` is only ever called after using `HubrisArchive::lookup_type` to go from `HubrisGoff -> HubrisType`.

Should we instead write the function `HubrisArchive::differs(a: HubrisGoff, b: HubrisGoff) -> Result<bool>`?

Here's a diff of what that looks like:
```diff
diff --git a/humility-core/src/hubris.rs b/humility-core/src/hubris.rs
index dd807c0..d7e2dfd 100644
--- a/humility-core/src/hubris.rs
+++ b/humility-core/src/hubris.rs
@@ -3045,21 +3045,21 @@ impl HubrisArchive {
             Ok(goffs)
         } else {
             for goff in goffs {
-                cmp.push((goff, self.lookup_type(goff)?));
+                cmp.push(goff);
             }
 
             for i in 0..cmp.len() {
                 let mut dup = false;
 
                 for j in i + 1..cmp.len() {
-                    if !cmp[i].1.differs(self, &cmp[j].1)? {
+                    if !self.differs(cmp[i], cmp[j])? {
                         dup = true;
                         break;
                     }
                 }
 
                 if !dup {
-                    out.push(cmp[i].0);
+                    out.push(cmp[i]);
                 }
             }
             Ok(out)
@@ -4963,6 +4963,30 @@ impl HubrisArchive {
     pub fn read_cmpa(&self) -> Result<Option<Vec<u8>>> {
         self.read_file("img/CMPA.bin")
     }
+
+    pub fn differs(&self, a: HubrisGoff, b: HubrisGoff) -> Result<bool> {
+        let lhs = self.lookup_type(a)?;
+        let rhs = self.lookup_type(b)?;
+        match (lhs, rhs) {
+            (HubrisType::Base(lt), HubrisType::Base(rt)) => {
+                Ok(lt.encoding != rt.encoding || lt.size != rt.size)
+            }
+            (HubrisType::Struct(lt), HubrisType::Struct(rt)) => {
+                lt.differs(self, rt)
+            }
+            (HubrisType::Enum(lt), HubrisType::Enum(rt)) => {
+                lt.differs(self, rt)
+            }
+            (HubrisType::Array(lt), HubrisType::Array(rt)) => {
+                lt.differs(self, rt)
+            }
+            (HubrisType::Union(lt), HubrisType::Union(rt)) => {
+                lt.differs(self, rt)
+            }
+            (HubrisType::Ptr(l), HubrisType::Ptr(r)) => self.differs(l, r),
+            _ => Ok(true),
+        }
+    }
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
@@ -5124,10 +5148,7 @@ impl HubrisStruct {
                     return Ok(true);
                 }
 
-                let l = hubris.lookup_type(lm.goff)?;
-                let r = hubris.lookup_type(rm.goff)?;
-
-                if l.differs(hubris, &r)? {
+                if hubris.differs(lm.goff, rm.goff)? {
                     return Ok(true);
                 }
             }
@@ -5154,10 +5175,7 @@ impl HubrisArray {
         if self.count != rhs.count {
             Ok(true)
         } else {
-            let l = hubris.lookup_type(self.goff)?;
-            let r = hubris.lookup_type(rhs.goff)?;
-
-            l.differs(hubris, &r)
+            hubris.differs(self.goff, rhs.goff)
         }
     }
 }
@@ -5323,10 +5341,7 @@ impl HubrisEnum {
 
                 match (lm.goff, rm.goff) {
                     (Some(lg), Some(rg)) => {
-                        let l = hubris.lookup_type(lg)?;
-                        let r = hubris.lookup_type(rg)?;
-
-                        if l.differs(hubris, &r)? {
+                        if hubris.differs(lg, rg)? {
                             return Ok(true);
                         }
                     }
@@ -5387,10 +5402,7 @@ impl HubrisUnion {
 
                 match (lm.goff, rm.goff) {
                     (Some(lg), Some(rg)) => {
-                        let l = hubris.lookup_type(lg)?;
-                        let r = hubris.lookup_type(rg)?;
-
-                        if l.differs(hubris, &r)? {
+                        if hubris.differs(lg, rg)? {
                             return Ok(true);
                         }
                     }
@@ -5475,24 +5487,6 @@ impl<'a> HubrisType<'a> {
             }
         }
     }
-
-    pub fn differs(&self, hubris: &HubrisArchive, rhs: &Self) -> Result<bool> {
-        match (self, rhs) {
-            (Self::Base(lt), Self::Base(rt)) => {
-                Ok(lt.encoding != rt.encoding || lt.size != rt.size)
-            }
-            (Self::Struct(lt), Self::Struct(rt)) => lt.differs(hubris, rt),
-            (Self::Enum(lt), Self::Enum(rt)) => lt.differs(hubris, rt),
-            (Self::Array(lt), Self::Array(rt)) => lt.differs(hubris, rt),
-            (Self::Union(lt), Self::Union(rt)) => lt.differs(hubris, rt),
-            (Self::Ptr(l), Self::Ptr(r)) => {
-                let lt = hubris.lookup_type(*l)?;
-                let rt = hubris.lookup_type(*r)?;
-                lt.differs(hubris, &rt)
-            }
-            _ => Ok(true),
-        }
-    }
 }
 
 impl<'a> From<&'a HubrisBasetype> for HubrisType<'a> {
```

It's a small LOC decrease, but I'm not 100% sold on this, because it does more work in `dedup` (calling `lookup_type` repeatedly).

jgallagher (approved):
Thank you!

humility-core/src/hubris.rs
jgallagher: I don't think I've ever used it, but `Vec::dedup_by()` exists; I don't _think_ we can use it here because `differs` is fallible, though?

humility-core/src/hubris.rs
bcantrill: Looks good; merged this with your other suggestion on avoiding the indexing, and the result is cleaner. Thank you!

humility-core/src/hubris.rs
bcantrill: That's only a temporary, so... no.

cmd/hiffy/src/lib.rs
```diff
@@ -401,13 +403,19 @@ fn hiffy(context: &mut humility::ExecutionContext) -> Result<()> {
...
     }
 
     let mut context = HiffyContext::new(hubris, core, subargs.timeout)?;
 
     if let Some(call) = subargs.call {
+        if core.is_dump() {
```
mkeeter, line 415: This check should go above `HiffyContext::new`; otherwise, HIF version mismatches cause it to fail first with a misleading error message:
```
➜  humility git:(test) ✗ cargo run -- -d tests/cmd/cores/hubris.core.kiowa.27 hiffy -c UserLeds.led_on
    Finished dev [unoptimized + debuginfo] target(s) in 0.21s
     Running `target/debug/humility -d tests/cmd/cores/hubris.core.kiowa.27 hiffy -c UserLeds.led_on`
humility: attached to dump
humility hiffy failed: HIF version mismatch: target has 0.2; ours is 0.3
```

humility-core/src/hubris.rs
bcantrill: Yeah, dunno on that one. I feel I could go either way. It *does* feel slightly cleaner -- and may make for a more useful future interface...

cmd/hiffy/src/lib.rs
bcantrill: I will remove the `requires = "list"` which doesn't, it turns out, actually work in this context. :|

cmd/hiffy/src/lib.rs
bcantrill: That's just a consequence of trying to run it on a very old dump. If you run it on a dump in which the HIF version doesn't mismatch, you'll get the expected error:

```
$ ./target/release/humility -d ./tests/cmd/cores/hubris.core.sprot_status hiffy -c Validate.validate_i2c
humility: attached to dump
humility hiffy failed: can't make HIF calls on a dump
```

And leaving it there allows for HIF functions to also be listed out of a dump, which has utility:

```
$ ./target/release/humility -d ./tests/cmd/cores/hubris.core.sprot_status hiffy -L
humility: attached to dump
 ID FUNCTION                       #ARGS
  0 Sleep                          1
  1 Send                           4
  2 SendLeaseRead                  5
  3 SendLeaseWrite                 5
  4 I2cRead                        7
  5 I2cWrite                       8
  6 I2cBulkWrite                   8
  7 GpioInput                      1
  8 GpioToggle                     2
  9 GpioSet                        2
 10 GpioReset                      2
 11 GpioConfigure                  7
 12 SpiRead                        4
 13 SpiWrite                       3
 14 QspiReadId                     0
 15 QspiReadStatus                 0
 16 QspiBulkErase                  0
 17 QspiPageProgram                3
 18 QspiRead                       2
 19 QspiSectorErase                1
 20 QspiVerify                     3
 21 QspiHash                       2
 22 HashDigest                     1
 23 HashInit                       0
 24 HashUpdate                     1
 25 HashFinalize                   0
 26 StartUpdate                    0
 27 WriteBlock                     2
 28 FinishUpdate                   0
 29 BlockSize                      0
 30 SpRotSendRecv                  2
 31 SpRotSendRecvRetries           3
 32 SpRotPulse                     1
 33 SpRotSink                      2
 34 SpRotStatus                    0
 35 SpRotStartUpdate               0
 36 SpRotWriteBlock                2
 37 SpRotFinishUpdate              0
 38 SpRotBlockSize                 0
```

cmd/hiffy/src/lib.rs
```diff
@@ -166,23 +164,27 @@ pub fn hiffy_list(hubris: &HubrisArchive, verbose: bool) -> Result<()> {
...
+            let mut ops = iface.ops.iter().peekable();
 
-            match ops.next() {
-                None => {
-                    println!("-");
-                }
-                Some(op) => {
-                    print!("{:<20}", op.0);
-                    print_args(&op, module, 49);
+            println!("{:<28} TASK", "INTERFACE");
```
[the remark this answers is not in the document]
bcantrill: Yeah. The output is so long, that you definitely lose what's what. See sample output from the test runs.

cmd/hiffy/src/lib.rs
```diff
@@ -166,23 +164,27 @@ pub fn hiffy_list(hubris: &HubrisArchive, verbose: bool) -> Result<()> {
...
+
+                let last = ops.peek().is_none();
+                let c = if last { "" } else { "|" };
+                let margin = format!("  {:<8}", c);
+
+                print_args(&op, module, margin);
+                println!("  {}", c);
             }
         }
     }
```
mkeeter, line 190: It would be nice to print a message if nothing matches, e.g.
```rust
    if !any_matches {
        println!(
            "warning: no matching tasks or interfaces found.  \
             Options are shown below:"
        );
        println!("{:<28} TASK", "INTERFACE");
        for i in 0..hubris.ntasks() {
            let module = hubris.lookup_module(HubrisTask::Task(i as u32))?;
            if let Some(iface) = &module.iface {
                println!("{:<28} {}", iface.name, module.name);
            }
        }
    }
```
(assuming that you have a `any_matches` variable updated in the loop)

cmd/hiffy/src/lib.rs
mkeeter: Yeah, I understand that; I'm thinking about the case where you've specified `--call`,  are running on a dump, **and** have a HIF mismatch.  In that case, I think "can't make HIF calls on a dump" message should be prioritized over the HIF mismatch error.

e.g.
```diff
@@ -409,13 +427,12 @@ fn hiffy(context: &mut humility::ExecutionContext) -> Result<()> {
         bail!("filters can only be provided with --list");
     }

+    if subargs.call.is_some() && core.is_dump() {
+        bail!("can't make HIF calls on a dump");
+    }
     let mut context = HiffyContext::new(hubris, core, subargs.timeout)?;

     if let Some(call) = subargs.call {
-        if core.is_dump() {
-            bail!("can't make HIF calls on a dump");
-        }
-
         let func: Vec<&str> = call.split('.').collect();

         if func.len() != 2 {
```

This should still allow you to list functions on a dump with a HIF mismatch.

cmd/hiffy/src/lib.rs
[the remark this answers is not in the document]
mkeeter: 👍🏻

cmd/hiffy/src/lib.rs
bcantrill: Yeah, good call, though I would rather explicitly bail in this case.

cmd/hiffy/src/lib.rs
bcantrill: Ugh, thank you for spotting that!

mkeeter (approved):

humility-core/src/hubris.rs
bcantrill: Ended up doing the refactor per the suggestion -- thank you!

PR #295, opened 2022-12-15
DRAFT: isp user kek and read key store

flihp:


flihp:
nope

labbott:
For future archeologists: UserKEK can only go to the AES block and can't be read out which isn't helpful for us.

PR #298, opened 2022-12-19
debugmailbox: commands for LPC55 debug authentication

### LINKED ISSUE #371: Debug mailbox commands for debug authentication
LPC55 debug authentication allows a device to have the debug port normally disabled until a signed challenge is provided. As described in NXP AN13037 and UM11126 §51.7, this authentication is performed by sending two commands to the debug mailbox.  The first requests a debug auth challenge from the target device and the second provides a signed debug auth response.

labbott:
Fixes #371

### DIFF
--- a/cmd/debugmailbox/src/lib.rs
+++ b/cmd/debugmailbox/src/lib.rs
@@ -121,13 +121,16 @@ fn poll_raw_ap_register(
 fn alive<'a>(
     probe: &mut Box<dyn ArmProbeInterface + 'a>,
     addr: &ApAddress,
+    reset: bool,
 ) -> Result<()> {
-    probe.write_raw_ap_register(*addr, CSW, 0x21)?;
+    if reset {
+        probe.write_raw_ap_register(*addr, CSW, 0x21)?;
+        println!("Resetting chip via SYSREQRESET!");
+    }
 
     let _ = poll_raw_ap_register(probe, addr, CSW, |csw| Ok(csw == 0), 1000)
         .context("Waiting for debugmailbox to be ready for requests")?;
 
-    println!("Reset chip successfully!");
     Ok(())
 }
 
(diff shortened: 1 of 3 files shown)


cmd/debugmailbox/src/lib.rs
```diff
@@ -64,21 +67,41 @@ enum DebugMailboxCmd {
...
-                break;
+        match probe.read_raw_ap_register(*addr, CSW) {
+            Ok(val) => {
+                println!("CSW = {val:x}");
+                if val == 0 {
+                    break;
+                }
+            }
+            Err(DebugProbeError::ArchitectureSpecific(arch_err)) => {
+                match arch_err.downcast::<DapError>() {
```
cbiffle, line 96: This method for detecting and tolerating a WAIT appears in at least two places -- it'd be nice to collect it and give it a name.

cbiffle (approved):

cmd/debugmailbox/src/lib.rs
kc8apf: PTAL

cmd/debugmailbox/src/lib.rs
```diff
@@ -64,36 +68,82 @@ enum DebugMailboxCmd {
...
     /// Force the device into ISP mode
     Isp,
+    /// Get a Debug Authentication Challenge
+    AuthChallenge { out: PathBuf },
+    /// Send a Debug Authentication Response
+    AuthResponse { dar: PathBuf },
 }
 
-fn alive<'a>(
+fn poll_raw_ap_register<'a, F>(
```
cbiffle, line 77: FWIW, you could eliminate the parameters and write this as

```rust
fn poll_raw_ap_register(
    probe: &mut Box<dyn ArmProbeInterface + '_>,
    ap: &ApAddress,
    addr: u8,
    f: impl Fn(u32) -> Result<bool>,
    timeout_ms: usize,
) -> Result<u32> {
```
I think this is preferable because
- it doesn't introduce the `'a` named lifetime just to use it once (I went searching to see where the other use was)
- I personally find `impl Trait` on parameters easier to read than separate where clauses; in cases where the parameter (here, `F`) appears only once they're basically equivalent to the compiler

For generality I'd also consider changing `Fn` to `FnMut`, which would allow the caller to e.g. maintain a counter. (That will also require you to `mut f`.)

cmd/debugmailbox/src/lib.rs
```diff
@@ -64,36 +68,82 @@ enum DebugMailboxCmd {
...
-            timeout -= -1;
         }
     }
 
-    if timeout == 0 {
-        bail!("Timed out waiting for reset. Chip is unlikely to be alive!");
-    } else {
-        println!("Reset chip successfully!");
-        Ok(())
+    Err(anyhow!("Timed out polling AP {ap:?} address 0x{addr:02x}"))
```
cbiffle, line 130: (I notice this is your only use of `anyhow!` -- if you wanted to remove the import, this line is equivalent to `bail!`)

cbiffle (approved):
Just a couple possible simplification suggestions, otherwise looks good.

cmd/debugmailbox/src/lib.rs
kc8apf: oh, that is nice.  I tried to do this initially but kept getting hung up on the probe lifetime.  Using an anonymous lifetime makes perfect sense.

PR #300, opened 2022-12-20
Better support for hubpack-encoded idol operations

jgallagher:
This PR fixes a couple of size-related bugs:

1. When humility determines the idol reply size to send to hiffy, it now uses the type size only for `zerocopy` encodings; for ssmarshal/hubpack, it uses the new `hubpack_serialized_maxsize()` function which attempts to replicate how hubpack will compute `MAX_SIZE` when deriving `SerializedSize`. This is important because it's easy to have types whose hubpack encoding is slightly larger than the type size, and without this change we would see faults in hiffy like

```
6 hiffy                        7   5 FAULT: reply fault: task id #15/gen0, reason ReplyBufferTooSmall (was: wait: reply from power/gen0)
```

2. Similarly, when serializing arguments, we no longer always assume a payload vec with length matching the type size; instead, we start with an empty vec and grow it as we push on each serialized argument.

The larger change is the removal of `payload_arg_serialized()` and the expansion of `serialize_arg()` to support enums and newtype structures. `payload_arg_serialized()` had two problems: (1) it tried to look up types by name, but we already had a goff at hand, and (2) before trying to look up enum/struct names, it would check for `RecvStrategy::FromBytes`. `RecvStrategy::FromBytes` is the default, and is not really meaningful with hubpack/ssmarshal encoding. (@cbiffle mentioned planning to restrict recv strategy to only zerocopy-encoded operations.)

I do not understand why we were doing lookups by name when we already had the member goff available; if there's a problem that was fixing, it's not covered by the tests, and looking enums up by names was causing me problems in practice (with unqualified enum names that were repeated across multiple tasks). Using the goff as in this PR resolved those problems.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1019,7 +1019,7 @@ dependencies = [
 
 [[package]]
 name = "humility"
-version = "0.9.17"
+version = "0.9.18"
 dependencies = [
  "anyhow",
  "bitfield 0.13.2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ name = "humility"
 #
 # Be sure to check in and push all of the files that change.  Happy versioning!
 #
-version = "0.9.17"
+version = "0.9.18"
 authors = ["Bryan Cantrill <bryan@oxide.computer>"]
 edition = "2018"
 license = "MPL-2.0"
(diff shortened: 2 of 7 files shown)


bcantrill (approved):
Ironically, I hit this same issue independently (while -- suspiciously? -- integrating review feedback from @jgallagher), and I have verified that this fixes it; emphatically approved!

PR #301, opened 2022-12-20
jefe: update docs to reflect new control operations

cbiffle:


bcantrill (approved):
This looks great -- thank you! My only comment would be to be sure to run `cargo xtask readme` and be sure to check in the README as well.

PR #302, opened 2022-12-20
humility-core: update "are we ROM?" message

cbiffle:
I believe this message was trying to hint to the user that the chip might be running in e.g. a boot ROM, but it just said "ROM," which confused me because our code also runs out of read-only memory.

Added more words to the warning to try not to confuse folks.


bcantrill (approved):
Looks great! I really like "other program" because it sounds so... troubling. Like, WHAT OTHER PROGRAM?! And, well, yeah... exactly.

bcantrill:
Note that the test output needs to be updated.  This can be done by running:

```console
% TRYCMD=overwrite cargo test
```

...and then `git add`'ing modified files.

PR #303, opened 2022-12-21
Use correct reply size in `humility rpc`

jgallagher:
Related to #300: for hubpack-encoded replies, it's possible for the encoding size to be greater than the typesize. This PR fixes the `nreply` calculation in `humility rpc` for encoded replies.

Note that even with this change, if the reply is _less_ than the maximum hubpack encoded size, currently the hubris task will return an `NReplyMismatch` error. I'll open a fix for that in hubris shortly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1019,7 +1019,7 @@ dependencies = [
 
 [[package]]
 name = "humility"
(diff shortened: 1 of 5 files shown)


jgallagher:
This is included in #304

PR #304, opened 2022-12-22
Add `humility powershelf` command

jgallagher:
This is paired with https://github.com/oxidecomputer/hubris/pull/1040 to allow us to remotely dump the ACAN-114 properties of an mwocp68: https://gist.github.com/jgallagher/ce929e2819971e02409865120184e768
 
In addition to adding the new command, this PR includes:

* A fix for calculating the reply size in `humility rpc` for hubpack or ssmarshal encoded operations (0bfbd9efb8d94512062a89ba36464771b2713067)
* Refactoring `humility rpc`, creating a public `RpcClient` that can be used by other commands (3ce46e7997c1c748d6f4ff5d1fd63c046ea433bb)
* Bumping to version 0.9.19

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1070,6 +1070,7 @@ dependencies = [
  "humility-cmd-reset",
  "humility-cmd-ringbuf",
  "humility-cmd-rpc",
+ "humility-cmd-rpc-powershelf",
  "humility-cmd-sensors",
  "humility-cmd-spctrl",
  "humility-cmd-spd",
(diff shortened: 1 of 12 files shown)


bcantrill (approved):
This looks great! What do you think about calling it just `powershelf` instead of `rpc-powershelf`? It just feels like we're going to want to start having a few of these, so this one will likely be setting some precedent... ;)

Also, one comment: be sure to run `cargo xtask readme` and check in `README.md` as well.

PR #307, opened 2023-01-03
Even more flexible HIF over the network

mkeeter:
#305 added support for HIF over the network by searching the program text for `Call` operations, reading their neighboring arguments (which must be in a specific format), and then executing them over the network.

This is neat, but means that HIF operations that use loops (e.g. `humility net mac`) don't work:

```
$ humility --ip $HUMILITY_IP -a $ARCHIVE net mac
humility: connecting to fe80::0c1d:ddff:fef8:fb69%en0
humility: Reading 5 MAC addresses...
humility net failed: can't make non-Idol calls over RPC
```

However, we've got a perfectly good HIF VM sitting right here!

This PR modifies the `HiffyContext::perform_rpc` to actually run the entire HIF program on the host, swapping the `Send` function for one which operates over the network.

```
$ humility --ip $HUMILITY_IP -a $ARCHIVE net mac
humility: connecting to fe80::0c1d:ddff:fef8:fb69%en0
humility: Reading 5 MAC addresses...
 PORT |        MAC
------|-------------------
    1 | 6a:e5:5c:b7:51:c6
      | d8:07:b6:da:28:dc
      | f0:2f:4b:09:a6:ec
    3 | 0e:1d:dd:f8:fb:69
      | 0e:1d:dd:f8:fb:6a
```

This is a _little cursed_, because HIF functions have a simple C API without any place to wedge a context pointer; I had to do some `unsafe` stuff to shove a `HubrisArchive` and `dyn Core` into a global.

Also, I'm seeing the same `Err(ffffff01)` that you noticed before, e.g.
```
matt@jeeves ~ () $ pfexec ./humility -a $SIDECAR rpc --ip fe80::aa40:25ff:fe05:101%axf1 -c Net.get_mac_address
fe80::aa40:25ff:fe05:101%axf1 Net.get_mac_address() => Err(ffffff01)
```

I don't think this is part of the new code, because it also happened in your demo.

When I tested on a Gimletlet, this was due to the `net` task hitting a stack overflow and restarting.  Indeed, `ffffff01` looks like a dead code of some kind.  If we can get a Gimlet or Sidecar on `jeeves` connected to both an ST-Link and the network, this should be pretty easy to debug.

### DIFF
--- a/humility-cmd/src/hiffy.rs
+++ b/humility-cmd/src/hiffy.rs
@@ -474,6 +474,8 @@ impl<'a> HiffyContext<'a> {
                 let mut workspace = HIFFY_SEND_WORKSPACE.lock().unwrap();
                 workspace.hubris = None;
                 workspace.core = None;
+                workspace.results = vec![];
+                workspace.errors = vec![];
             }
         }
         let _cleanup = WorkspaceCleanup;
@@ -494,6 +496,11 @@ impl<'a> HiffyContext<'a> {
 
         self.state = State::Kicked;
         let workspace = HIFFY_SEND_WORKSPACE.lock().unwrap();
+        if let Some(e) = workspace.errors.first() {
+            // We have to translate from Error -> String to work around
+            // ownership
+            return Err(anyhow!(e.to_string()));
+        }
         for buf in &workspace.results {
             //
             // If udprpc gave us an error, it's because something was
@@ -999,8 +1006,10 @@ struct HiffySendWorkspace {
     hubris: Option<*const HubrisArchive>,
     core: Option<*mut dyn Core>,
 
-    /// If we receive an RPC error, then record the buffer here
+    /// If we receive an RPC result, then record the buffer here
     results: Vec<Vec<u8>>,
+
+    errors: Vec<anyhow::Error>,
 }
 
 // SAFETY:
@@ -1014,6 +1023,7 @@ lazy_static::lazy_static! {
             hubris: None,
             core: None,
             results: vec![],
+            errors: vec![],
         }
     ));
 }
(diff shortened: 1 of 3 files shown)


humility-cmd/src/hiffy.rs
```diff
@@ -1006,3 +998,144 @@ impl<'a> HiffyContext<'a> {
...
+        // SAFETY: we only ever call this function when the pointers are
+        // populated, and reset them to None afterwards.  This means we should
+        // fail at the initial unwrap() if someone violates the rules.
+        unsafe {
+            (
+                workspace.hubris.unwrap().as_ref().unwrap(),
+                workspace.core.unwrap().as_mut().unwrap(),
+            )
+        }
+    };
```
jgallagher, line 1090: Unlocking `HIFFY_SEND_WORKSPACE` while we hold references to data it contains seems really dicey, even given the warnings about the cursedness of this code. :) Could we lock `HIFFY_SEND_WORKSPACE` once and keep it locked for the duration of this function instead? At least then we know we're not going to accidentally create multiple references to the archive or core.

mkeeter: ah yeah, that's a good catch – I was aggressively unlocking it because otherwise a panic in this function causes mutex poisoning, but that seems sketchy.  Take a look at f23209896c885c36fecf2ecdc86567da96b469b3

jgallagher: That looks good, thanks!

mkeeter:
The dead code issue was https://github.com/oxidecomputer/hubris/pull/1046, so I think this is clear to merge

humility-cmd/src/hiffy.rs
```diff
@@ -1006,3 +1011,147 @@ impl<'a> HiffyContext<'a> {
...
+        // Try to receive a reply
+        match core.recv(buf.as_mut_slice()) {
+            Err(e) => {
+                workspace.errors.push(e);
+                return Err(Failure::FunctionError(0));
+            }
+            Ok(_n) => (),
+        }
+
+        workspace.results.push(buf.to_vec());
```
jgallagher, line 1139: Should this be `buf[..n].to_vec()` (at which point maybe it moves into the `match` above?

jgallagher (approved):

humility-cmd/src/hiffy.rs
mkeeter: This matches the previous code, probably because the decoder allows for trailing data in the slice.  Still, it would probably be cleaner to just push the valid packet data, so I'll do that.

PR #312, opened 2023-01-05
Bank erase

labbott:


cmd/bankerase/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+        default_value_t = 100, value_name = "timeout_ms",
+        parse(try_from_str = parse_int::parse)
+    )]
+    reset_delay: u64,
+
+    #[clap(long, parse(try_from_str = parse_int::parse))]
+    address: u32,
+
+    #[clap(long, parse(try_from_str = parse_int::parse))]
+    len: u32,
```
andrewjstone, line 36: We probably want to note that due to alignment requirements this will always erase at least 32kib. Because of this, do we actually even want to allow a `len`? Maybe instead we size things in terms of **erase blocks**?

andrewjstone: I changed this to taking erase pages, and it works fine, but I also realized I did not notice in prior readings of the code that the erase command doesn't ensure it aligns the address on a 32KiB boundary. When I ran `humility bankerase 0x40001` it erased the full page, but didn't overwrite the first byte with a zero. I think this is fine, and keeps with the existing behavior. I just wanted to double check.

andrewjstone: @labbott I think I"m actually going to revert my (local) change and keep things as is. The reason is that erasing a page doesn't mean much if you are in the middle of the page. Do you erase only the current page, or the current and next page because it's 8k bytes. And since you overwrite with zeroes only the `len` given, the semantics that exist make more sense to me.

andrewjstone (approved):
This looks good and works for me. I suggest we merge it in to remove the dependency on pyocd for this operation on the RoT.

cmd/bankerase/src/lib.rs
```diff
@@ -0,0 +1,111 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! ## `humility bankerase`
+//!
+//! Erase some memory please write some words
```
bcantrill, line 7: Worth fleshing this out with a sentence or two. ;)

cmd/bankerase/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+    force: bool,
+
+    /// reset delay
+    #[clap(
+        long = "reset-delay", short = 'd',
+        default_value_t = 100, value_name = "timeout_ms",
+        parse(try_from_str = parse_int::parse)
+    )]
+    reset_delay: u64,
+
```
bcantrill, line 30: Would be good to indicate what these options are for so they show up in `humility bankerase --help`

bcantrill (commented):
Looks good -- worth fleshing out the docs a tad?

cmd/bankerase/src/lib.rs
andrewjstone: Docs added in https://github.com/oxidecomputer/humility/pull/312/commits/ea6b8bc523789eb14f41624a2e7342ced5d65a6c

bcantrill (approved):

PR #313, opened 2023-01-06
Dumper

bcantrill:


humility-core/src/core.rs
```diff
@@ -1371,7 +1378,69 @@ impl NetCore {
...
+        for shdr in elf.section_headers.iter() {
+            if shdr.sh_type != goblin::elf::section_header::SHT_PROGBITS {
+                continue;
+            }
+
+            flash_regions.insert(
+                shdr.sh_addr as u32,
+                (shdr.sh_size as u32, shdr.sh_offset as usize),
+            );
+        }
```
mkeeter, line 1410: Optional mutation-free version:
```suggestion
        let flash_regions = elf
            .section_headers
            .iter()
            .filter(|shdr| {
                shdr.sh_type == goblin::elf::section_header::SHT_PROGBITS
            })
            .map(|shdr| {
                (
                    shdr.sh_addr as u32,
                    (shdr.sh_size as u32, shdr.sh_offset as usize),
                )
            })
            .collect();
```

humility-core/src/hubris.rs
```diff
@@ -4546,7 +4547,11 @@ impl HubrisArchive {
         //
         let mut written = 0;
 
-        let started = Instant::now();
+        let started = match started {
+            Some(s) => s,
+            None => Instant::now(),
+        };
```
mkeeter, line 4553:
```suggestion
        let started = started.unwrap_or_else(Instant::now);
```

humility-core/src/core.rs
```diff
@@ -1371,7 +1378,69 @@ impl NetCore {
...
+
+        Ok(Self { socket, flash_contents, flash_regions })
+    }
+
+    fn read(&self, addr: u32, data: &mut [u8]) -> Result<()> {
```
mkeeter, line 1415: This seems like it's got the same logic as `AgentCore::read_flash`; dunno if there's an easy way to refactor them.

cmd/dump/src/lib.rs
```diff
@@ -48,34 +48,895 @@
...
+    force_read: bool,
+
+    /// initialize dump state, clearing any dump at the dump agent
+    #[clap(long, conflicts_with = "simulation")]
+    initialize_dump_agent: bool,
+
+    /// overwrite any dump state as part of taking dump
+    #[clap(
+        long, short = 'F',
+        conflicts_with_all = &["simulation", "initialize-dump-agent"]
```
mkeeter, line 113: I think this should only conflict with `simulate_dumper`, not the entire `simulation` group; right now I see the following

```console
➜  humility git:(dumper) ✗ cargo run -- dump --emulate-dumper --force-dump-agent
    Finished dev [unoptimized + debuginfo] target(s) in 0.49s
     Running `target/debug/humility dump --emulate-dumper --force-dump-agent`
humility dump failed: there appears to already be a dump in situ; clear with --initialize-dump-agent or force dump to be overwritten with --force-overwrite
➜  humility git:(dumper) ✗ cargo run -- dump --emulate-dumper --force-dump-agent --force-overwrite
    Finished dev [unoptimized + debuginfo] target(s) in 0.36s
     Running `target/debug/humility dump --emulate-dumper --force-dump-agent --force-overwrite`
error: The argument '--force-overwrite' cannot be used with:
    --simulate-dumper
    --emulate-dumper

USAGE:
    humility dump --emulate-dumper --force-dump-agent --force-overwrite

```

(this may be true of `--initialize-dump-agent` as well)

cmd/dump/src/lib.rs
```diff
@@ -48,34 +48,895 @@
...
+
+            let len = (size as usize) - start;
+            data[..len]
+                .copy_from_slice(&self.flash_contents[roffs..roffs + len]);
+
+            self.read_flash(addr + len as u32, &mut data[len..])
+        } else {
+            bail!("address 0x{:08x} not found", addr);
+        }
+    }
```
mkeeter, line 221: Nit: this may be more clear as an explicit loop, instead of recursing

```suggestion
    fn read_flash(&self, mut addr: u32, mut data: &mut [u8]) -> Result<()> {
        while !data.is_empty() {
            let (&base, &(size, offset)) =
                self.flash_regions.range(..=addr).last().ok_or_else(|| {
                    anyhow!("address 0x{:08x} not found", addr)
                })?;

            if !(base..base + size).contains(&addr) {
                //
                // It's not here, and we have already tried RAM...
                //
                bail!("address 0x{:08x} not found", addr);
            }

            let start = (addr - base) as usize;
            let roffs = offset + start;

            let amount = (size as usize - start).min(data.len());
            data[0..amount]
                .copy_from_slice(&self.flash_contents[roffs..roffs + amount]);
            addr += amount as u32;
            data = &mut data[amount..];
        }
        Ok(())
    }
```

andrewjstone: You also prevent a possible stack overflow that way. I typically like recursion, but stack overflows make me sad.

andrewjstone: I guess if its only a few sections, probably not really an issue.

cmd/dump/src/lib.rs
bcantrill: Ha, yes -- had just independently made this change as part of testing your Nucleo fix. ;)

cmd/dump/src/lib.rs
bcantrill: Yeah, dunno on that one. To me, it's clearer as recursion -- but I may chalk this one up to personal preference?

mkeeter: Yup, I don't feel too strongly about it.

humility-core/src/core.rs
bcantrill: Yeah, that was Q&D to get it working -- it needed a refactor. Done, and thanks for the suggestion!

humility-core/src/core.rs
bcantrill: Did this as part of the refactor.

humility-core/src/hubris.rs
bcantrill: Yeah, really need to get my fingers better at typing this one. Fixed!

PR #315, opened 2023-01-10
probe: print out SAU and MPU(s) active configuration

kc8apf:


cmd/probe/src/lib.rs
```diff
@@ -422,6 +466,157 @@ fn probecmd(context: &mut humility::ExecutionContext) -> Result<()> {
         }
     }
 
+    if part.has_tz() {
```
andrewjstone, line 469: On the s28 I'm seeing the secure MPU registers show up (with the same values as NS). This appears to be because the dregion is not 0. On the SP, the S registers properly don't show. Not really a huge issue, but figured I'd mention it.

andrewjstone (approved):
LGTM. Would be good to get this in for debugging purposes.

cmd/probe/src/lib.rs
kc8apf: Hmm.  This is a bug in ARMCore::has_tz() which assumes that all Cortex-M33 implementations have TrustZone-M.  That is definitely not always true.  I'll file a separate bug.

PR #317, opened 2023-01-17
Listen on ports 8 and 8888 for udpbroadcast

mkeeter:
This will let us fix https://github.com/oxidecomputer/hubris/issues/1087 without breaking `humility rpc --listen`.

This code spawns one thread per port, instead of trying to listen to multiple ports simultaneously, because it's easier.

I also changed the table-printing code to skip the header if nothing was returned, which I think was a regression.

### DIFF
--- a/cmd/rpc/src/lib.rs
+++ b/cmd/rpc/src/lib.rs
@@ -134,7 +134,22 @@ fn rpc_listen_one(
     interface: u32,
     port: u32,
 ) -> Result<BTreeSet<Target>> {
-    let socket = UdpSocket::bind(&format!("[::]:{port}"))?;
+    let socket = match UdpSocket::bind(&format!("[::]:{port}")) {
+        Ok(s) => s,
+        Err(e) => {
+            if e.kind() == std::io::ErrorKind::PermissionDenied {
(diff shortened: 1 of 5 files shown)


cmd/rpc/src/lib.rs
```diff
@@ -200,15 +186,63 @@ fn rpc_listen(rpc_args: &RpcArgs) -> Result<BTreeSet<Target>> {
...
+        timeout.as_secs(),
+        ports
     );
 
-    println!("-------------------|---------------------------| -----------");
+    let threads = ports
+        .iter()
+        .map(|&port| {
+            std::thread::spawn(move || -> Result<BTreeSet<Target>> {
+                rpc_listen_one(timeout, interface, port)
```
jclulow, line 216: Would it be possible to ignore a permissions error here, so that you could run `humility` as an unprivileged user (albeit only being able to bind to port 8888)?  Right now I think it would just fail because 8 would fail, even though 8888 would likely then work.

mkeeter: Good idea, done in fdec592ef806255411b0ddcd0ef8a0512437c724

PR #325, opened 2023-02-01
IBC blackbox

mkeeter:
This PR adds `humility ibc`, which reads and decodes the BMR491's blackbox log.

Right now, it's all done through raw I2C; we'll eventually want to shift it to Idol and then expose it to MGS.


bcantrill (approved):
This looks awesome!

PR #329, opened 2023-02-13
Add support for persistent QSPI data

mkeeter:
This PR adds a `--write-sector0` flag to `humility qspi`, without which it will refuse to modify the contents of QSPI sector 0.  This is part of implementing https://github.com/oxidecomputer/hubris/issues/985, but should be backwards-compatible; Humility will fall back to older functions if the new ones aren't present.

I don't love the state of the code, but was trying not to get embroiled in a larger `humility qspi` refactor!

### DIFF
--- a/cmd/qspi/src/lib.rs
+++ b/cmd/qspi/src/lib.rs
@@ -77,6 +77,10 @@
 //! 120000..000080: 4d07112733efe240f990fad785726c52de4335d6c5c30a33e60096d4c2576742
 //! ```
 //!
+//! By default, Hubris reserved sector 0 (the lowest 64 KiB) of QSPI for its own
+//! internal bookkeeping.  To override this, use the `--write-sector0` flag;
+//! this is required for erases and writes that would otherwise modify sector 0,
+//! as well as bulk erase.
 
 use humility::cli::Subcommand;
 use humility::core::Core;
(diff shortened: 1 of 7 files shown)


jgallagher (approved):
LGTM :+1:

cmd/qspi/src/lib.rs
```diff
@@ -532,12 +614,18 @@ fn qspi(context: &mut humility::ExecutionContext) -> Result<()> {
...
         let mut offset = 0;
 
         let mut buf = vec![0u8; chunk as usize];
         let mut file = File::open(filename)?;
 
+        // Skip sector0 if the user didn't specify --write-sector0
+        if !subargs.write_sector0 {
+            offset += SECTOR_SIZE;
+            file.seek(std::io::SeekFrom::Start(SECTOR_SIZE as u64)).unwrap();
+        }
```
lzrd, line 627: Are we silently doing a partial write of the total requested writes? If so, shouldn't we fail this operation?

jgallagher (approved):

cmd/qspi/src/lib.rs
mkeeter: Yes, unless you pass the `--write-sector0` flag, Humility will skip sector 0 when writing images.  This is added to the documentation at the top of the file.

I've tweaked the logic in ee5a1632431a47f717f7bfec91c0bc59b5daf567: if you write an image that includes anything other than `0xFF` in sector 0, then Humility will stop with an error.  This matches the behavior in `control-plane-agent`, which is motivated by (a) doing the right thing and (b) not requiring more dramatic changes.

PR #341, opened 2023-03-18
support for task dumps

### LINKED ISSUE #149: humility readvar using unqualified variable names only
```
$ arm-none-eabi-nm --demangle target/demo-stm32h743-nucleo/dist/kernel  | grep TASK_DESCS
080048b8 r kern::startup::HUBRIS_TASK_DESCS

$ humility -a target/demo-stm32h743-nucleo/dist/build-demo-stm32h743-nucleo.zip readvar kern::startup::HUBRIS_TASK_DESCS
humility readvar failed: variable kern::startup::HUBRIS_TASK_DESCS not found

$ humility -a target/demo-stm32h743-nucleo/dist/build-demo-stm32h743-nucleo.zip readvar HUBRIS_TASK_DESCS
HUBRIS_TASK_DESCS (0x080048b8) = [
     {
....
```

Second command in that transcript ought to work, but doesn't. Third command arguably _should not_ work, but does; best case I guess it could print all variables with a name in all modules in all tasks?

bcantrill:
Draft commit message:

```

This work:

- Adds support for task dumps.  These dumps look like full system
  dumps, but only contain a single task.  (All commands should be able
  to run correctly on task dumps, but some will naturally not find
  data that they are looking for.)

- Fixes some long-standing issues whereby Humility would become
  confused by duplicate structure names in different scopes (which in
  turn induces failures in e.g. "humility hiffy", which would emit
  a wildly cryptic error message).

- Adds a "-e" option to "humility probe" to emit the envrionment
  variable associated with a particular probe

- Fixes #149, with "humility readvar" using fully qualified names

- Adds the task's object ID to "humility manifest" output (essential
  for making sense of error messages that include GOFF identifiers).

- Bumps the minor version (to 0.10.0) and adds the ability to read
  Hubris v8 dumps.  

```


Cargo.toml
```diff
@@ -88,7 +88,7 @@ members = [
 [workspace.dependencies]
 # `git`-based deps
 hif = { git = "https://github.com/oxidecomputer/hif" }
-humpty = { git = "https://github.com/oxidecomputer/humpty" }
+humpty = { git = "https://github.com/oxidecomputer/humpty", branch = "taskdumpy" }
```
mkeeter, line 91: TODO: switch this back

cmd/dump/src/lib.rs
```diff
@@ -51,7 +51,7 @@
 //! 12 hf                           0   3 recv
 //! 13 update_server                0   3 recv
 //! 14 sensor                       0   4 recv, notif: bit0(T+472)
-//! 15 host_sp_comms                0   7 recv, notif: bit0(irq82) bit1 bit2(T+59) bit3
+//! 15 host_sp_comms                0   7 recv, notif: bit0(irq82) bit1
```
mkeeter, line 54: This seems like a weird thing to change; intentional?

cmd/dump/src/lib.rs
```diff
@@ -426,6 +444,52 @@ fn emulate_dump(core: &mut dyn Core, base: u32, total: u32) -> Result<()> {
     Ok(())
 }
 
+fn emulate_task_dump(
```
mkeeter, line 447: I'm confused by this function name – it looks like it's just setting up the dump segments, rather than actually doing the dump, in contrast with `emulate_dump`.

cmd/dump/src/lib.rs
```diff
@@ -616,36 +693,154 @@ fn read_dump_header(
...
+    let mut base = 0;
     let mut rval = vec![];
-    let mut done = false;
-    let mut ndx = 0;
 
-    while !done {
-        let header =
-            read_dump_at(hubris, core, context, funcs, ndx, |_| false)?;
+    //
+    // Iterate over dump area headers in chunksize units.  Once we've hit an
```
mkeeter, line 745: Add a note that this only works because the area header is less than `rsize` here.

cmd/dump/src/lib.rs
```diff
@@ -616,36 +693,154 @@ fn read_dump_header(
...
+
 fn read_dump_headers(
     hubris: &HubrisArchive,
     core: &mut dyn Core,
     context: &mut HiffyContext,
     funcs: &HiffyFunctions,
-) -> Result<Vec<DumpAreaHeader>> {
+    raw: bool,
+) -> Result<Vec<(DumpAreaHeader, Option<DumpTask>)>> {
+    let rsize = 256;
```
mkeeter, line 738: Is it too much of a pain to pull this value from the `IdolOperation`?

cmd/dump/src/lib.rs
```diff
@@ -616,36 +693,154 @@ fn read_dump_header(
...
+                // a continuation of our previous header -- otherwise, we're
+                // done processing entirely.
+                //
+                if header.dumper != humpty::DUMPER_NONE {
+                    areas.push(*header);
+                } else {
+                    break;
+                }
+            }
+            Some(_task) => {
```
mkeeter, line 827: Add a comment that this marks the beginning of a new task dump, so we record the old one then start again.

cmd/stackmargin/src/lib.rs
```diff
@@ -47,10 +47,18 @@ fn stackmargin(context: &mut humility::ExecutionContext) -> Result<()> {
     let (base, size) = hubris.task_table(core)?;
     let task = hubris.lookup_struct_byname("Task")?;
     let taskdesc = hubris.lookup_struct_byname("TaskDesc")?;
+    let dump_task = hubris.dump_task();
 
     let mut taskblock: Vec<u8> = vec![];
     taskblock.resize_with(task.size * size as usize, Default::default);
-    core.read_8(base, taskblock.as_mut_slice())?;
+
+    if let Some(HubrisTask::Task(i)) = dump_task {
```
mkeeter, line 55: I don't understand the changes in this file; it looks like we're skipping reading the task block for the dump task?

cmd/stackmargin/src/lib.rs
```diff
@@ -72,6 +80,12 @@ fn stackmargin(context: &mut humility::ExecutionContext) -> Result<()> {
     };
 
     for i in 0..size {
+        if let Some(HubrisTask::Task(ndx)) = dump_task {
+            if ndx != i {
```
mkeeter, line 84: Same here, a comment would be helpful (I'm guessing this is to avoid reading the extra memories mapped to the dump task, but not sure).

cmd/dump/src/lib.rs
```diff
@@ -758,8 +1020,35 @@ fn dump_via_agent(
...
+            let ndx = match hubris.lookup_task(task) {
+                Some(HubrisTask::Task(ndx)) => ndx,
+                _ => {
+                    bail!("invalid task \"{task}\"");
+                }
+            };
+
+            Some(DumpTask::new(
+                *ndx as u16,
+                core.read_word_64(hubris.lookup_variable("TICKS")?.addr)?,
```
mkeeter, line 1047: You could use the new `hubris.ticks` here!

cmd/dump/src/lib.rs
bcantrill: Done

cmd/stackmargin/src/lib.rs
bcantrill: I think the nomenclature here is confusing; changing this from `dump_task` to `task_dump` and adding a comment.

mkeeter: Aha, I now understand: this is a special case for if we're operating on a dump and the dump only contains a single task.  A comment to that effect would be welcome!

cmd/tasks/src/lib.rs
```diff
@@ -200,18 +200,25 @@ fn tasks(context: &mut humility::ExecutionContext) -> Result<()> {
     loop {
         core.halt()?;
 
-        let cur =
-            core.read_word_32(hubris.lookup_symword("CURRENT_TASK_PTR")?)?;
+        let cur = hubris.current_task(core)?;
+        let dump_task = hubris.dump_task();
```
mkeeter, line 204: Same here – a comment explaining that this is a special case for dumps with only a single task would make this more obvious.  From the variable name, I was expecting this to be handling Hubris archives with a `dump-agent` task in them, which didn't make much sense.

humility-cmd/src/idol.rs
```diff
@@ -615,55 +618,48 @@ pub fn lookup_reply<'a>(
...
-                                }
-                            }
+    let lookup_ok_err = |ok| {
+        if let Err(e) = m.lookup_enum_byname(hubris, ok) {
+            Err(e)
+        } else if let Err(e) = m.lookup_struct_byname(hubris, ok) {
+            Err(e)
+        } else {
+            bail!("no type for {}.{op}", iface.name);
+        }
```
mkeeter, line 628: Since we just want to bail on errors, could we do this instead?
```suggestion
        m.lookup_enum_byname(hubris, ok)?;
        m.lookup_struct_byname(hubris, ok)?;
        bail!("no type for {}.{op}", iface.name);
```

humility-core/src/hubris.rs
```diff
@@ -435,6 +438,9 @@ pub struct HubrisArchive {
     // current object
     current: u32,
 
+    // non-None if a dump of a task
```
mkeeter, line 441:
```suggestion
    // non-None if a dump of a single task
```

humility-core/src/hubris.rs
```diff
@@ -570,9 +580,60 @@ impl HubrisArchive {
...
+    fn namespace(&self, namespace: Option<usize>) -> Result<Vec<&String>> {
+        let mut rval = vec![];
+        let mut current = namespace;
+
+        while let Some(namespace) = current {
+            if namespace >= self.namespaces.len() {
+                bail!("namespace id {namespace} is invalid");
+            }
+
+            let (name, parent) = &self.namespaces[namespace];
```
mkeeter, line 600:
```suggestion
            let (name, parent) =
                self.namespaces.get(namespace).ok_or_else(|| {
                    anyhow!("namespace id {namespace} is invalid")
                })?;
```

humility-core/src/hubris.rs
```diff
@@ -570,9 +580,60 @@ impl HubrisArchive {
...
+            rval.push(name);
```
mkeeter, line 601: I'd suggest making this function return `Result<Vec<&str>>`, so this would become
```suggestion
            rval.push(name.as_str());
```

humility-core/src/hubris.rs
```diff
@@ -570,9 +580,60 @@ impl HubrisArchive {
...
+
+        if !n.is_empty() {
+            n.push(name);
+
+            let full = n
+                .iter()
+                .copied()
+                .map(String::as_str)
+                .collect::<Vec<_>>()
+                .join("::");
```
mkeeter, line 629: If you change `fn namespace` to return a `Result<Vec<&str>>`, this can simply become
```suggestion
            let full = n.join("::");
```

humility-core/src/hubris.rs
```diff
@@ -1834,6 +1914,33 @@ impl HubrisArchive {
...
+                        while let Some(attr) = attrs.next()? {
+                            if attr.name() == gimli::constants::DW_AT_name {
+                                let name = dwarf_name(&dwarf, attr.value());
+
+                                if let Some(name) = name {
+                                    let id = self.namespaces.len();
+                                    let parent =
+                                        namespace.last().map(|(id, _)| *id);
+
+                                    self.namespaces
```
mkeeter, line 1934: This probably wants a struct of some kind; it's confusing to have `self.namespaces` and `namespaces`, each of which is a `Vec` containing two-item tuples with very different semantics.

It would probably be worthwhile to make types, e.g.
```rust
struct NamespacedType {
    name: String,
    parent: Option<usize>,
}
```
and
```rust
struct NamespaceEntry {
    index: usize,
    depth: usize,
}
```

humility-core/src/hubris.rs
```diff
@@ -1820,10 +1887,23 @@ impl HubrisArchive {
...
 
+                //
+                // See if our depth has become shallower than our namespace,
+                // trimming it until it fits.
+                //
+                while let Some((_, d)) = namespace.last() {
+                    if depth > *d {
+                        break;
+                    }
+
```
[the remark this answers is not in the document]
mkeeter: Alternatively, do we need to track `depth` in the `namespace` array at all, or could it be implicit based on the length of the `Vec`?

cmd/dump/src/lib.rs
bcantrill: I objected to its columnar width. ;)

cmd/stackmargin/src/lib.rs
bcantrill: I think the nomenclature change has made this much clearer, but let me know if you think it should be clearer still.

cmd/dump/src/lib.rs
bcantrill: Ha, good point!

humility-core/src/hubris.rs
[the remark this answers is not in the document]
bcantrill: It can't be implicit because the depth can be deeper than the length.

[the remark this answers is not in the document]
bcantrill: > You _could_ do this as a one-liner, but not sure if you want to:

So, this one is less clear to me, FWIW.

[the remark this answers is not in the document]
mkeeter: Okay, I'm fine leaving it as is

humility-core/src/hubris.rs
bcantrill: This was nagging me too; cleaned this up quite a bit.

mkeeter (approved):

PR #342, opened 2023-03-23
Support array arguments as strings in Idol

mkeeter:
This will let us eliminate stuff like
```
        "get_key_by_tag": (
            doc: "Scans the caboose for a key with the given tag",
            args: {
                "name": "[u8; 4]",
            },
            leases: {
                "data": (type: "[u8]", write: true),
            },
            reply: Result(
                ok: "u32",
                err: CLike("CabooseError"),
            ),
            idempotent: true,
        ),
        "get_key_by_u32": (
            doc: "Scans the caboose for a key with the given tag",
            args: {
                "tag": "u32",
            },
            leases: {
                "data": (type: "[u8]", write: true),
            },
            reply: Result(
                ok: "u32",
                err: CLike("CabooseError"),
            ),
            idempotent: true,
        ),
```
where the latter only exists because Idol couldn't take a `[u8; 4]` as an argument.

Also, this improves the pretty-printing of data returned by Idol operations that write to a lease.

Together, this means we can do stuff like this:
```
$ target/debug/humility hiffy -c Caboose.get_key_by_tag -aname=BORD -n32
Caboose.get_key_by_tag() => 0xb
             \/  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
0x00000000 | 67 69 6d 6c 65 74 6c 65 74 2d 32 00 00 00 00 00 | gimletlet-2.....
0x00000010 | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | ................
```

### DIFF
--- a/humility-cmd/src/idol.rs
+++ b/humility-cmd/src/idol.rs
@@ -340,6 +340,36 @@ fn call_arg(
                 }
             }
         }
+        HubrisType::Array(&HubrisArray { goff, count }) => {
+            let t = hubris.lookup_type(goff)?;
+            if !matches!(
+                t,
+                HubrisType::Base(HubrisBasetype {
+                    encoding: HubrisEncoding::Unsigned,
+                    size: 1,
+                })
+            ) {
+                bail!(
+                    "array type in {} ({t:?}) not yet supported",
+                    member.name
+                );
+            }
+            match value {
+                IdolArgument::String(value) => {
+                    if value.len() != count {
+                        bail!(
+                            "Cannot convert '{value}' to [u8; {count}]; \
+                             wrong length"
+                        );
+                    }
+                    let dest = &mut buf[member.offset..member.offset + count];
+                    dest.copy_from_slice(value.as_bytes());
+                }
+                IdolArgument::Scalar(v) => {
+                    bail!("Cannot convert scalar {v} to [u8; {count}]")
+                }
+            }
+        }
         _ => {
             bail!("type of {} ({:?}) not yet supported", member.name, t);
         }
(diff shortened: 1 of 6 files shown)


humility-cmd/src/idol.rs
```diff
@@ -340,6 +340,36 @@ fn call_arg(
...
+                    size: 1,
+                })
+            ) {
+                bail!(
+                    "array type in {} ({t:?}) not yet supported",
+                    member.name
+                );
+            }
+            match value {
+                IdolArgument::String(value) => {
```
jgallagher, line 358: Do we ever want to pass raw bytes? I could imagine wanting some way to pass in a hex string.

jgallagher (approved):

humility-cmd/src/idol.rs
mkeeter: Maybe, but I don't think it's worth doing now.

Doing a quick audit, the only other case where we pass an array as argument is `image_hash`, which doesn't seem like it would be called from the CLI:
```
"fetch_host_phase2_data": (
   doc: "Instruct the control plane agent to fetch a portion of a host phase 2 image, identified by `hash` and starting at `offset`. When the data arrives, the control plane agent will notify the calling task, at which point it should call `get_host_phase2_data`.",
   args: {
       // Hash of the phase 2 image the host wants.
       "image_hash": "[u8; 32]",
       // Offset within the image the host wants.
       "offset": "u64",
       // Notification bit (i.e., log2 of the actual IRQ value) we
       // should use when the data has been fetched.
       "notification_bit": "u8",
```

PR #348, opened 2023-04-04
Fix verify algorithm.

cbiffle:
Previously, the verify algorithm was only checking a subset of data written to flash (PHDRs that are read-only and executable), and was checking it against the source ELF objects from the archive. This was bad for a couple of reasons.

1. You could erase a sector in the middle of a vector table, or some task's rodata, and humility wouldn't notice it -- so flash --verify wouldn't fix it! (Corruption in such areas would also go unnoticed.)

2. We alter the image between when the source ELF objects are generated and when it hits flash. In particular: we sign it, on LPC55.

I've replaced the verify algorithm here with one that compares the contents of target memory with the `final.elf` file produced at the very end of the build process. This means that, compared to the original,

- errors in the vector table and rodata are detected
- on LPC55, the signature itself is checked for consistency with the archive

### DIFF
--- a/humility-core/src/hubris.rs
+++ b/humility-core/src/hubris.rs
@@ -2460,7 +2460,6 @@ impl HubrisArchive {
                 heapbss,
                 task,
                 iface,
-                buffer: buffer.to_vec(),
             },
         );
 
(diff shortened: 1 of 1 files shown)


bcantrill (approved):
Looks great -- and vastly improved!

PR #351, opened 2023-04-06
Store functions in HiffyContext instead of on-demand

mkeeter:
Right now, we have an odd pattern that occurs in many places:

- Get a map of functions (`HiffyFunctions`) out of a `HiffyContext` with `functions()`
- Pass that list of functions _back into_ calls on that `HiffyContext`, e.g. `context.idol_call_ops_inner(funcs, op, payload, ...)`

(likely due to the history of the `HiffyContext` and how it has grown functionality over time)

This PR changes the behavior to be a little simpler:

- `HiffyFunctions` are calculated _once_ and stored within the `HiffyContext` at construction
- The `funcs` map is no longer passed into `HiffyContext` calls
- A new `get_function` method is added to avoid the two-step dance of getting the `HiffyFunctions` object, then getting a function out of it; this function is used _almost_ everywhere (except in `humility hiffy`, where we care about the full length of the array)

### DIFF
--- a/cmd/dashboard/src/lib.rs
+++ b/cmd/dashboard/src/lib.rs
@@ -787,7 +787,6 @@ fn sensor_ops(
     capture: impl Fn(&HubrisSensor) -> bool,
 ) -> Result<Vec<String>> {
     let mut sensors = vec![];
-    let funcs = context.functions()?;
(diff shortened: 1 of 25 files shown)


bcantrill (approved):
Thank you again for doing this!

PR #353, opened 2023-04-06
Reduce duplicate builds

steveklabnik:
We only want to build on PR *or* on pushes to the master branch, we don't want to build on every branch, since PRs are also branches, or else we build twice for no reason.


cbiffle (approved):
Seems reasonable to me!

PR #355, opened 2023-04-06
Update dependencies

steveklabnik:
Some of these have been yanked, some of these have active dependabot alerts. None seem to be an active problem for us, but it's easy to update them, so let's get clean.


steveklabnik:
Updating `zip` gave us better error messages? Hence the updates to the fixtures.

PR #358, opened 2023-04-14
Add single-task dumping and memory reads in the `NetCore`

mkeeter:
This PR follows
- https://github.com/oxidecomputer/humpty/pull/5
- https://github.com/oxidecomputer/hubris/pull/1284

It implements a few things:
- Single-task dumping using the new APIs from the PRs above
- Letting the `NetCore` read RAM using the task region dump API

Things spiraled a _little_ out of control as I attempted to tame the Humility dependency tree.  This mostly involved splitting modules into crates of their own, to ease the blockage at `humility` and `humility_cmd`:

- Moved `NetCore` from `humility` to a separate crate `humility-net-core`
- Moved `hiffy`, `idol`, and `doppel` out of `humility_cmd` into standalone crates
- Added a new `humility-dump-agent` crate, which is used in both `humility_cmd_dump` and `NetCore`
- Fixed everything that these changes broke

This isn't done – it's still too easy to trigger a rebuild of every single command – but it's a step in the right direction.


Cargo.toml
```diff
@@ -88,7 +93,7 @@ members = [
 [workspace.dependencies]
 # `git`-based deps
 hif = { git = "https://github.com/oxidecomputer/hif" }
-humpty = { git = "https://github.com/oxidecomputer/humpty", version = "0.1.1" }
+humpty = { git = "https://github.com/oxidecomputer/humpty", branch = "single-task-dump", version = "0.1.2" }
```
bcantrill, line 96: Just a reminder to move this over before pushing.

humility-net-core/src/lib.rs
```diff
@@ -0,0 +1,300 @@
...
+        dump_agent_socket.connect(&dest[..])?;
+
+        let mut out = Self {
+            udprpc_socket,
+            dump_agent_socket,
+            flash: HubrisFlashMap::new(hubris)?,
+            ram: vec![], // filled in below
+        };
+
+        // Surprisingly, this works: `hubris.regions` only reads data from the
```
bcantrill, line 85: 😉

humility-net-core/src/lib.rs
```diff
@@ -0,0 +1,300 @@
...
+        // We're going to collect just this region into an ersatz DumpAgentCore
+        (&mut udp_dump as &mut dyn DumpAgent).read_dump(
+            Some(DumpArea::ByIndex(dump_index as usize)),
+            &mut agent_core,
+            false,
+        )?;
+
+        // Pop the most recent dump, since we were just using it to read memory
+        // and it doesn't need to take up a dump area forever.
+        udp_dump.reinitialize_dump_from(dump_index)?;
```
bcantrill, line 195: I don't know to what degree we care, but won't this also have the effect of popping the dump of any task that dumps in the interim?

bcantrill (approved):
This looks great; please bump the version because this is going to be a version to remember. ;)

humility-net-core/src/lib.rs
mkeeter: It will; I think that's a reasonable tradeoff in order to avoid filling up dump memory, but I'll add a comment to that effect.

Cargo.toml
mkeeter: Done!

PR #366, opened 2023-04-20
Remove LPC55 support code

mkeeter:
This code is a divergent evolutionary branch from [`lpc55_support`](https://github.com/oxidecomputer/lpc55_support), and we shouldn't keep both of them maintained.


mkeeter:
Hold off on merging this; we've only written to LPC55s using `humility isp`, and not `lpc55_flash` so far on illumos, for unknown reasons.

kc8apf (commented):
ISP commands can definitely be removed.  I'm less clear on PFR commands.  They certainly duplicate the functionality from lpc55_flash _but_ lp55_flash is exclusively over UART ISP mode while `humility pfr` is over SWD.   I tried to test whether `humility pfr read-cfpa` would work with a locked CMPA (`lpc55_flash <port> read-cfpa` fails when CMPA is locked because the ROM ISP rejects the commands) but ran into humility demanding an archive and failing when it didn't match the device contents.

kc8apf:
After temporarily changing `humility lpc55-pfr` to not require an archive, `humility lpc55-pfr read-cfpa` fails when trying to read the CFPA scratch page:
```
./target/release/humility lpc55-pfr read-cfpa
humility lpc55-pfr failed: failed to perform halted read at address 0x9de00 for length 512

Caused by:
    0: A core architecture specific error occured
    1: Failed to read register DRW at address 0x0000000c
    2: An error specific to a probe type occured
    3: CMSIS-DAP responded with an error
```

Guess that answers that.  It might be fixable but, as is, `humility lpc55-pfr` only duplicates the functionality of `lpc55_flash` despite being over SWD instead of UART.  I'm fine with removing it.

kc8apf (approved):

mkeeter:
...and the bringup yesterday worked with `lpc55_utils`, so we're clear to merge this.

PR #368, opened 2023-04-21
include DMA regions in dumps

bcantrill:
Draft comment:

```
This work allows for dumping of DMA regions, which it does by
explicitly not dumping external regions.  (Note that if this is run
against a Hubris that cannot dump DMA regions, these regions won't be
dumped.)  This also adds modifies "humility map" to indicate which
regions are external, and further to label those external regions if
sufficient metadata is available in the archive.  Finally, this
work also modifies task dumping to always include the name of the
task in the dump filename (if no filename is explicitly provided).
```


tests/cmd/registers/registers.extern-regions.toml
```diff
@@ -0,0 +1,10 @@
+
+#
+# This test case has been automatically created, but can be edited and
+# should be checked in.  Should it ever be regenerated, simply delete
+# it and re-run "cargo test"
+#
+fs.base = "../cores"
+bin.name = "humility"
+args = "-d hubris.core.extern-regions registers"
+
```
mkeeter, line 10: I'm confused why this passed; don't we need to specify
```
status.code = 1 # Hand-edit: no i2c task is present in this image
```

mkeeter (approved):

PR #369, opened 2023-04-21
Allow humility::msg to capture variables inline; do that more

mkeeter:
Previously, `humility::msg` could not capture variables because it used `concat!` to build the format string.

Also, secretly re-exports `colored`, so that downstream crates don't have to use it themselves.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1364,7 +1364,6 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "clap",
- "colored",
  "csv",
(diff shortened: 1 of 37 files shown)


mkeeter:
One more change: I stripped a trailing ' ' from a log message, then updated the test suite to pass.

steveklabnik (approved):
Looks good to me. I hadn't considered it before, but I love the various places where four lines turns into one, simply because of moving the variable name into the format string itself. A nice side effect of using that syntax.

PR #370, opened 2023-04-24
Implement `humility rendmp --blackbox`

mkeeter:
This is mostly boilerplate to mirror the data structures, plus a little plumbing to attach to the new Idol API (https://github.com/oxidecomputer/hubris/pull/1240 / https://github.com/oxidecomputer/hubris/pull/1310).

### DIFF
--- a/cmd/rendmp/src/blackbox.rs
+++ b/cmd/rendmp/src/blackbox.rs
@@ -15,10 +15,9 @@ fn format_faults<T: Into<u32> + std::fmt::Binary + Copy>(
     f: &mut Formatter,
 ) -> std::fmt::Result {
     let u: u32 = v.into();
-    let faults: Vec<String> = faults
+    let faults: Vec<_> = faults
         .iter()
-        .filter(|(i, _)| u & (1 << i) != 0)
-        .map(|(_, v)| v.to_string())
+        .filter_map(|&(i, v)| (u & (1 << i) != 0).then_some(v))
         .collect();
 
     let size = std::mem::size_of::<T>() * 8;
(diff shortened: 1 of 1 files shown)


cmd/rendmp/src/blackbox.rs
```diff
@@ -0,0 +1,774 @@
...
+fn format_faults<T: Into<u32> + std::fmt::Binary + Copy>(
+    v: T,
+    faults: &[(u32, &'static str)],
+    f: &mut Formatter,
+) -> std::fmt::Result {
+    let u: u32 = v.into();
+    let faults: Vec<String> = faults
+        .iter()
+        .filter(|(i, _)| u & (1 << i) != 0)
+        .map(|(_, v)| v.to_string())
```
jgallagher, line 21: Two tiny nits - maybe `filter_map()` instead of `.filter(..).map(..)`, and I don't think `.to_string()` is necessary since we already have a `&str`. Combining both:

```rust
     let faults: Vec<_> = faults
         .iter()
         .filter_map(|&(i, v)| (u & (1 << i) != 0).then_some(v))
         .collect();
```

cmd/rendmp/src/blackbox.rs
```diff
@@ -0,0 +1,774 @@
...
+                    "Internal Temperature Fault",
+                    "The controller internal temp has exceeded 130°C.",
+                ),
+                (3, "BBEVENT", "A Black Box event occurred."),
+                (2, "LMSEVENT", "A Last Man Standing event has occurred."),
+                (
+                    1,
+                    "SPSFault",
+                    "An SPS overcurrent and/or over-temperature \
+                     event has occurred.",
```
jgallagher, line 237: I think this will put a bunch of spaces in the string. Maybe `concat!()` instead to split across lines?

cmd/rendmp/src/blackbox.rs
```diff
@@ -0,0 +1,774 @@
...
+                (
+                    4,
+                    "VIN_UV_FAULT",
+                    "An input undervoltage fault has occurred.",
+                ),
+                (
+                    3,
+                    "VIN_ON/OFF",
+                    "Disabled due to insufficient input voltage. \
+                     This could be VIN or VMON.",
```
jgallagher, line 350: Same comment as above - spaces -> `concat!()`?

cmd/rendmp/src/blackbox.rs
```diff
@@ -0,0 +1,774 @@
...
+            (5, "PLL fault"),
+            (4, "oscillator fault"),
+            (2, "VCCS UV fault"),
+            (1, "VCC UV fault"),
+            (0, "VCC UV fault"),
+        ];
+        let faults: Vec<String> = faults
+            .iter()
+            .filter(|(i, _)| v & (1 << i) != 0)
+            .map(|(_, v)| v.to_string())
```
jgallagher, line 602: Same nits as above - `filter_map()` and no `.to_string()`

jgallagher (approved):

cmd/rendmp/src/blackbox.rs
mkeeter: Neat!

cmd/rendmp/src/blackbox.rs
mkeeter: Not sure why this isn't using `format_faults`, so let's just do that instead.

cmd/rendmp/src/blackbox.rs
bcantrill: Ha, yes, this is something I have done/do a lot of in Humility. ;)

Issue #374, opened 2023-04-26
Add ability to pull Open Pin Detection data from Renesas power controllers


ericaasen:
The DMA addresses and bit maps for open-pin detection are as follows:

### ISL68224 Bit Mapping
0x00BD
Bits [0:1]: Phase 0
Bits [2:3]: Phase 1
...
Bits [22:23]: Phase 12
Bits [24:25]: VSEN0
Bits [26:27]: VSEN1
Bits [28:29]: VSEN2

### RAA229618 Bit Mapping
**0x00BE**
Phases 0-15 in the same pattern as above

**0x00BF**
Bits [0:1]: Phase 16
...
Bits [6:7]: Phase 19
Bits [8:9]: VSEN0
Bits [10:11]: VSEN1
Bits [12:13]: VSEN2 (not used for us)

Based on the captures below: these are the values for the registers with processor in vs. out

**Processor Absent**
RAA229618 (both)
`0x00BE: 0x55 51 55 51`
`0x00BF: 0x00 00 0f 55`

ISL68224
`0x00BD: 0x30 00 04 00`

**Processor Present**
RAA229618 (both)
`0x00BE: 0x55 51 55 51`
`0x00BF: 0x00 00 00 55`

ISL68224
`0x00BD: 0x00 00 00 00`

For the RAA parts, we see open pin detection in both VESEN bits for VSEN0 and VSEN1, as expected, when the processor is missing
For the ISL part, we see bits 28 and 29 set as well, which indicates that VSEN2 is open

wrt the phase bits, it seems we need to combine the mask registers for the open pin detection with these to see what we should be looking at.

For the RAA229618 at 0x5A (which has phases 0-3 connected to VDD_MEM_ABCD and phases 12-19 connected to VDD_VCORE) the mask register addresses are:
`0xE904: AA 00 00 AA`
`0xE905: 00 00 0f AA`
Assuming the bit mapping is the same, it shows phases 0-3 have their high bits set to 1 and phases 12-19 have their high bits set to 1 and VSEN0 and VSEN1 have all their bits set.

If you bitwise-AND the mask of AA with the open pin detection register value, it ends up with all 0s for the open pin detection per-phase, meaning no open pins

For the RAA229618 at 0x5B, the mask registers are:
`0xE904: AA 00 00 00`
`0xE905: 00 00 0f AA`
Which indicates phases 0-3 are mapped and phases 16-19 are mapped, which matches the schematic for that controller.

[...]

PR #375, opened 2023-04-27
Implement `humility rendmp --open-pin`

mkeeter:
closes https://github.com/oxidecomputer/humility/issues/374


cmd/rendmp/src/lib.rs
```diff
@@ -920,6 +937,93 @@ fn rendmp_blackbox(
...
+    let regs: &'static [u16] = match dev {
+        ISL_DEV_NAME => &[
+            0x00BD, // open-pin
+            0xE925, // mask
+        ],
+        RAA_DEV_NAME => &[
+            0x00BE, 0x00BF, // open-pin
+            0xE904, 0xE905, // mask
+        ],
+        _ => panic!("wrong device ({dev})"),
```
jgallagher, line 957: Tiny nit - maaaaaybe `check_addr` should return `(u8, SomeDeviceEnum)` instead of a string to avoid needing this default case?

cmd/rendmp/src/lib.rs
```diff
@@ -920,6 +937,93 @@ fn rendmp_blackbox(
...
+            }
+            (open, mask, phases)
+        }
+        _ => unreachable!(), // checked above
+    };
+    println!("Open pin check for {} at {addr:#x}", dev.to_uppercase());
+    println!(" {} | {}", "PHASE".bold(), "STATE".bold());
+    println!("-------|-------");
+    for (name, bit) in phases {
+        let is_open = (open & mask & (0b11 << bit)) != 0;
```
jgallagher, line 1012: Even tinier nit - `(0b11 << bit)` caught me off guard because that's masking two bits, not one bit. I'm not sure there's a better name here though; `bit_offset` is maybe not any clearer? Feel free to ignore this.

jgallagher (approved):

cmd/rendmp/src/lib.rs
mkeeter: Good idea, done

cmd/rendmp/src/lib.rs
mkeeter: Renamed to `bit_offset` and (more importantly) added some comments

PR #380, opened 2023-05-02
add phase current support to "humility power"

### LINKED ISSUE #338: Please add the mV digit to pmbus -v readout
Currently, the voltage shown by pmbus -v on a specific rail only goes to 2 digits past the decimal, but many of our rails are set to the mV and the data exists in the part for that resolution, so please update at least the -v option to print out to 3 digits.  The rail in the example below is actually set to 1.875V, but the display rounds to 1.88V:

```
λ humility pmbus -r V1P8_TF2_VDD -v
0x00 PAGE                      0x01 = 1
0x01 OPERATION                 0x48
     |
     | b7     0b0 = output off               <= OnOffState
     | b6     0b1 = powers down based on TOFF_DELAY <= TurnOffBehavior
     | b5:4   0b0 = set by VOUT_COMMAND      <= VoltageCommandSource
     | b3:2   0b10 = act upon margin faults  <= MarginFaultResponse
     | b1     0b0 = not set                  <= TransitionControl
     +-----------------------------------------------------------------------

0x02 ON_OFF_CONFIG             0x16
     |
     | b4     0b1 = OPERATION/CONTROL        <= PowerControl
     | b3     0b0 = ignores OPERATION on/off <= CommandControl
     | b2     0b1 = abides by CONTROL pin    <= ControlControl
     | b1     0b1 = active high              <= ControlPolarity
     | b0     0b0 = use programmed delay     <= ControlOff
     +-----------------------------------------------------------------------

0x04 PHASE                     0x00
0x07 ZONE_CONFIG               0xfefe
0x08 ZONE_ACTIVE               0x0000
0x10 WRITE_PROTECT             0x00
0x19 CAPABILITY                0xd0
     |
     | b7     0b1 = supported                <= PacketErrorChecking
     | b6:5   0b10 = 1 MHz                   <= MaximumBusSpeed
     | b4     0b1 = supported                <= SMBusAlert
     | b3     0b0 = not IEEE Half Precision  <= NumericFormat
     | b2     0b0 = not supported            <= AVSBusSupport
     +-----------------------------------------------------------------------

0x20 VOUT_MODE                 0x40
     |
     | b7     0b0 = absolute setting         <= AbsoluteOrRelative
     | b6:5   0b10 = DIRECT format           <= Mode
     | b4:0   0x0                            <= Parameter
     +-----------------------------------------------------------------------

0x21 VOUT_COMMAND              0x0753 = 1.88V
```

bcantrill:
Draft comment:

```
In addition to adding phase current support to "humility power", this
fixes #338 and #270 -- and also adds the requirement that all commands
have documentation in order to build.
```


cmd/power/src/lib.rs
```diff
@@ -28,15 +35,84 @@ struct PowerArgs {
...
+    hubris: &HubrisArchive,
+    core: &mut dyn Core,
+    context: &mut HiffyContext,
+    devices: &mut BTreeMap<(&String, HubrisSensorDevice), Device>,
+) -> Result<()> {
+    let op = hubris.get_idol_command("Power.phase_current")?;
+    let ok = hubris.lookup_basetype(op.ok)?;
+    let mut ops = vec![];
+
+    if ok.encoding != HubrisEncoding::Float {
```
mkeeter, line 66: Nit: you could combine these conditionals, since I don't think we need to be _this_ fine-grained with error reporting
```suggestion
    if (ok.encoding, ok.size) != (HubrisEncoding::Float, 4) {
```

cmd/power/src/lib.rs
```diff
@@ -28,15 +35,84 @@ struct PowerArgs {
...
+    let mut ndx = 0;
+
+    for device in devices.values_mut() {
+        if let Some(phases) = device.phases {
+            let mut phase_currents = vec![];
+
+            for _ in phases.iter() {
+                phase_currents.push(match &results[ndx] {
+                    Ok(val) => Some(f32::from_le_bytes(val[0..4].try_into()?)),
+                    _ => None,
```
mkeeter, line 105: Should we be logging something about the error here?

cmd/power/src/lib.rs
```diff
@@ -137,38 +231,69 @@ fn power(context: &mut ExecutionContext) -> Result<()> {
...
         p(d.current);
+
         p(d.input_voltage);
         p(d.input_current);
         p(d.temperature);
 
         println!();
+
+        if let Some(phase_currents) = &d.phase_currents {
+            if phase_currents.len() > 1 {
```
mkeeter, line 279: This means we don't print phase current for the ISL68224, which only has one phase per rail.  Is that intentional?

mkeeter (approved):

cmd/power/src/lib.rs
bcantrill: Yes, that's intentional.

cmd/power/src/lib.rs
bcantrill: Fixed by using the error character ("x") rather than the absence character ("-") when printing the phase.

cmd/power/src/lib.rs
bcantrill: Good call -- and also did this in the code that I copy-and-pasted it from, and then the places that I cut-and-pasted that code to. ;)

PR #382, opened 2023-05-05
Add abstraction and `net` support to `humility pmbus`

mkeeter:
`humility pmbus` directly constructs I2C commands and executes them on-device.

Unfortunately, raw I2C commands can't be executed over the network.

The PR creates a trait for "a thing which does PMBus work", then implements that trait for both
- the existing raw I2C operations
- Idol operations (implemented in https://github.com/oxidecomputer/hubris/pull/1340)

All of the code in `humility pmbus` is refactored to use this trait, meaning it works transparently with both debugger and network connectivity.

### DIFF
--- a/cmd/pmbus/src/lib.rs
+++ b/cmd/pmbus/src/lib.rs
@@ -1727,6 +1727,7 @@ impl PmbusWorker for IdolWorker<'_> {
             .ok_or_else(|| anyhow!("could not find device matching {harg:?}"))?
             .0;
         self.dev_index = Some(dev_index);
+        self.rail_index = None;
         Ok(())
     }
 
@@ -1740,11 +1741,14 @@ impl PmbusWorker for IdolWorker<'_> {
     }
 
     fn read(&mut self, code: u8, op: pmbus::Operation) {
-        let index = IdolArgument::Scalar(
-            self.rail_index.unwrap_or_else(|| self.find_rail(0)) as u64,
-        );
+        let (index, has_rail) = match self.rail_index {
+            None => (self.find_rail(0), false),
+            Some(r) => (r, true),
+        };
+        let index = IdolArgument::Scalar(index as u64);
+        let has_rail = IdolArgument::Scalar(has_rail as u64);
         let code = IdolArgument::Scalar(code as u64);
-        let args = [("index", index), ("op", code)];
+        let args = [("index", index), ("has_rail", has_rail), ("op", code)];
         match op {
             pmbus::Operation::ReadBlock => {
                 let payload = self.read_block.payload(&args).unwrap();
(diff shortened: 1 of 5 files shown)


cmd/pmbus/src/lib.rs
```diff
@@ -45,7 +45,7 @@
 //! ```
 //!
 //! In the unusual case that a device is unknown to the system (that is, it does
-//! not appear in `humliity manifest`), you can force a particular PMBus driver
+//! not appear in `humility manifest`), you can force a particular PMBus driver
```
bcantrill, line 48: Ugh, thank you for fixing this typo!

cmd/pmbus/src/lib.rs
```diff
@@ -1461,6 +1401,519 @@ fn writes(
...
+
+        // Read CONTROLLER_CONFIG from the core, which we'll use to compute
+        // indices for rails when using raw PMBus functions.
+        let (_name, var) = hubris
+            .qualified_variables()
+            .find(|&(n, _v)| n == "task_power::bsp::CONTROLLER_CONFIG")
+            .ok_or_else(|| anyhow!("could not find CONTROLLER_CONFIG"))?;
+        let mut buf: Vec<u8> = vec![0u8; var.size];
+        core.halt()?;
+        core.read_8(var.addr, buf.as_mut_slice())?;
```
bcantrill, line 1644: A lot happening under this particular hood! ;)

mkeeter:
> Getting the `CONTROLLER_CONFIG` index is a bit gnarly; is the rationale for that that we want to simplify the (hotter) path of doing PMBus operations?

Mostly; it's also motivated by having Too Many Ways to select a rail (like you mentioned in the [Hubris PR](https://github.com/oxidecomputer/hubris/pull/1340)).

This is [my attempt to have one correct way to do it](https://xkcd.com/927/)  – instead of having the SP iterate through and find a match by some subset of address / rail / device type, the host is in charge of figuring that out.

bcantrill:
> > Getting the `CONTROLLER_CONFIG` index is a bit gnarly; is the rationale for that that we want to simplify the (hotter) path of doing PMBus operations?
> 
> Mostly; it's also motivated by having Too Many Ways to select a rail (like you mentioned in the [Hubris PR](https://github.com/oxidecomputer/hubris/pull/1340)).
> 
> This is [my attempt to have one correct way to do it](https://xkcd.com/927/) – instead of having the SP iterate through and find a match by some subset of address / rail / device type, the host is in charge of figuring that out.

Makes sense. It would be great to have a way of pulling that from *just* the TOML just because the memory read feels invasive -- but maybe I should just stop thinking about the number of kittens that need to be slaughtered to read remote task memory. Also, if there were no room for improvement here, there would be no incentive for the 15th (and universal!) implementation.

Ship it!

mkeeter:
`CONTROLLER_CONFIG` is a `static` variable, so we're reading it from the flash section of the image without hitting the network.  I hope that makes you feel better about the kittens.

Thanks for the review; I bumped the version and am queuing up automerge.

Issue #367, opened 2023-04-20
Add ability to enable/disable individual phases on Renesas power controllers


ericaasen:
This is an issue to track the various logic analyzer dumps of what is happening to the controller from PowerNavigator when I do this through the GUI.

A DSLogic U3Pro16 was used to capture all of these.  I included a csv decode of the transactions and the raw capture in each of the dumps below.  You can get pre-compiled versions of the Dream Source Lab GUI for Windows, Linux, and OS X [here](https://www.dreamsourcelab.com/download/)

If you are so inclined, the source code for the GUI is theoretically [here](https://github.com/DreamSourceLab/DSView)

ericaasen:
Folder containing all dumps is [here](https://drive.google.com/drive/folders/1OI4Eqn3dGi14G-yYl3GgK8FNu8HoFati?usp=sharing)

1. Starting with a default config, disable Phase 19 of U350 (VDD_VCORE controller): [`0x5A_VDD_VCORE_controller_phase19_disable`](https://drive.google.com/file/d/1OIKYixRZSkxjb4n2m87zP_IpHSApJCs0/view?usp=sharing)
2. Disable Phase 18: [`0x5A_VDD_VCORE_controller_phase18_disable`](https://drive.google.com/file/d/1OIQYUysPPHVTNKxVhSLiiLhFhx4KK1SB/view?usp=sharing)
3. Disable Phase 17: [`0x5A_VDD_VCORE_controller_phase17_disable`](https://drive.google.com/file/d/1OJFGgVmRhgEaCE4QazWzvVQRMTOQCvkM/view?usp=sharing)
4. Disable Phase 16: [`0x5A_VDD_VCORE_controller_phase16_disable`](https://drive.google.com/file/d/1OKebxyDSY4B4VqiVLzjgnQU3e3g8yp6m/view?usp=sharing)
5. Disable Phase 15: [`0x5A_VDD_VCORE_controller_phase15_disable`](https://drive.google.com/file/d/1OL1swYPZx6797Y2M5JYZpfNrcw0v0Pew/view?usp=sharing)
6. Disable Phase 14: [`0x5A_VDD_VCORE_controller_phase14_disable`](https://drive.google.com/file/d/1OM2tB68v9gr_lAfgdQnfGQLgsepholy7/view?usp=sharing)
[...]

PR #383, opened 2023-05-10
Renesas per-phase check

mkeeter:
This PR implements per-phase checks for Renesas multiphase power converts.

It's a bit of a mess, but that's mostly due to intrinsic complexity:
- We have do a mix of writes and read-modify-writes to configure the chip
- HIF calls are slow, so we want to group them into a single larger program

The result is a system which batches up a bunch of Idol calls, executes them, then iterates over the resulting values.

The actual implementation is based on https://github.com/oxidecomputer/humility/issues/367 as well as email communications with Renesas FAEs.  We understand about 90% of it; it would be nice to get the canonical names for DMA registers from the Power Navigator logs.

We're also suffering from https://github.com/oxidecomputer/hubris/issues/1342:
- DMA read/write operations expect an I2C address (which is easy to find, but could be a non-unique identifier); PMBus operations added in https://github.com/oxidecomputer/hubris/pull/1340 use index within the `CONTROLLER_CONFIG` array (which is globally unique but harder to compute)

This is totally untested, so I'm opening it as a draft for now; it's possible that we're overflowing the Idol text space with these large programs and will want to restructure.  We'll test it once Eric is back from Rochester.

TODO:
- [x] Test on real hardware
- [x] Improve output (print in tabular form)

### DIFF
--- a/cmd/rendmp/src/lib.rs
+++ b/cmd/rendmp/src/lib.rs
@@ -906,7 +906,7 @@ impl std::fmt::Display for SupportedDevice {
             SupportedDevice::ISL68224 => "ISL68224",
             SupportedDevice::RAA229618 => "RAA229618",
         };
-        write!(f, "{s}",)
+        write!(f, "{s}")
     }
 }
 
--- a/cmd/rendmp/src/lib.rs
+++ b/cmd/rendmp/src/lib.rs
@@ -1270,7 +1270,6 @@ impl<'a, 'b> HifWorker<'a, 'b> {
                     .map(|i| *i as u32)
             })
             .collect::<Result<Vec<_>>>()?;
-        println!("got rail_indexes: {rail_indexes:?}");
 
         Ok(Self {
             hubris,
@@ -1601,7 +1600,7 @@ fn rendmp_phase_check<'a>(
     use pmbus::commands::raa229618::CommandCode::VOUT_MODE;
     worker.read_byte(rail_indexes[0], false, VOUT_MODE as u8)?;
 
-    // Disable all rails
+    // Disable all phases
     const PHASE_ENABLE_REG: u16 = 0xE9C2;
     worker.write_dma(addr, PHASE_ENABLE_REG, 0)?;
 
@@ -1827,6 +1826,12 @@ fn rendmp_phase_check<'a>(
         match next()? {
             Ok(v) => v.expect_write_dma()?,
             Err(e) => bail!(
+                "failed to enable phase {phase_name}: {e}"
+            ),
+        }
+        match next()? {
+            Ok(v) => v.expect_write_dma()?,
+            Err(e) => bail!(
                 "failed to enable forced-freq pwm for phase {phase_name}: {e}"
             ),
         }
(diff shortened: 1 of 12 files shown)


cmd/rendmp/src/lib.rs
```diff
@@ -1002,55 +1047,959 @@ fn rendmp_open_pin(
...
+    }
+
+    fn run(
+        &mut self,
+        core: &mut dyn humility::core::Core,
+    ) -> Result<Vec<Result<Call, String>>> {
+        let mut ops = std::mem::take(&mut self.ops);
+        ops.push(Op::Done);
+        let r = self.context.run(core, &ops, None)?;
+        // TODO: decode error messages here
```
jgallagher, line 1430: Did you want to address this TODO before merging?

cmd/rendmp/src/lib.rs
```diff
@@ -1002,55 +1047,959 @@ fn rendmp_open_pin(
...
+                bail!("failed to modify LOOPCFG for rail {rail}: {e}")
+            }
+        }
+        match next()? {
+            Ok(v) => v.expect_write_word()?,
+            Err(e) => {
+                bail!("failed to modify PEAK_OCUC_COUNT for rail {rail}: {e}")
+            }
+        }
+        if dev == SupportedDevice::RAA229618 {
```
jgallagher, line 1749: Nit - would it be worth using a `match` here instead so if we add to `SupporteDevice` in the future, we get a compile-time reminder to check whether the new device needs anything here?

jgallagher (commented):
I'm assuming all the detailed bitfield stuff is correct and really only reviewed the Rust code. Left a few minor nits - LGTM.

cmd/rendmp/src/lib.rs
```diff
@@ -842,19 +858,55 @@ fn rendmp_ingest(subargs: &RendmpArgs) -> Result<()> {
...
+}
+
 impl std::fmt::Display for SupportedDevice {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         write!(
             f,
             "{}",
             match self {
-                SupportedDevice::ISL68221 => "ISL68221",
+                SupportedDevice::ISL68224 => "ISL68221",
```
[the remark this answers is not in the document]
mkeeter: Yes, done in ae603d5c8b18488427624f3ad3ea44cb30f247ba

cmd/rendmp/src/lib.rs
```diff
@@ -983,10 +1030,8 @@ fn rendmp_open_pin(
     };
     let mut ops = vec![];
     for r in regs {
-        let payload = op.payload(&[
-            ("addr", IdolArgument::Scalar(addr as u64)),
-            ("reg", IdolArgument::Scalar(*r as u64)),
-        ])?;
+        let payload =
+            op.payload(&[("addr", addr.into()), ("reg", (*r).into())])?;
```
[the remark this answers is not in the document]
mkeeter: Sure, done in 37f177d299e582955be7838aa2ed151d9a63a649

cmd/rendmp/src/lib.rs
mkeeter: This is actually done in the loop + match below, so I've removed the comment in 9c0d4b6125ed915d839915abd98bdeadab30e8c0

cmd/rendmp/src/lib.rs
```diff
@@ -1002,55 +1047,959 @@ fn rendmp_open_pin(
...
+    write_word: IdolOperation<'a>,
+    write_word32: IdolOperation<'a>,
+
+    ops: Vec<hif::Op>,
+    calls: Vec<Call>,
+
+    rail_indexes: Vec<u32>,
+
+    /// Mapping from a phase (in the chip sense) to a rail (as an index in
+    /// `self.rail_indexes`).
```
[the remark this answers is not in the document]
mkeeter: Yup, updated the comment in 8ef19b3ac323d71f52dfd7c7c6bf3ce404b826d8

cmd/rendmp/src/lib.rs
mkeeter: Sure, although this decoding is paired with a write operation in the loop above, so it's not strictly necessary.  Done in eb4676c7ad703b0c2769543fe837c1d5e6662794

PR #384, opened 2023-05-17
Fix `humility update` after hubris#1357

mkeeter:
I don't know if anyone is actually using this subcommand, but it's an easy enough fix.

### DIFF
--- a/cmd/update/src/lib.rs
+++ b/cmd/update/src/lib.rs
@@ -42,7 +42,7 @@ struct UpdateArgs {
 
     #[clap(long)]
     /// Which image to update
-    target: String,
+    target: Option<String>,
     #[clap(long)]
     /// Don't commit the image, only write
     skip_commit: bool,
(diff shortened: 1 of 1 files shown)


cmd/update/src/lib.rs
```diff
@@ -75,12 +75,24 @@ fn update(context: &mut ExecutionContext) -> Result<()> {
...
+    // always update the alternate image.  Older SP images _also_ always update
+    // the alternate image, but take an image_target argument to match the RoT's
+    // API.
+    //
+    // We check the number of arguments here and behave appropriately.
+    let args = match (start.operation.args.is_empty(), &subargs.target) {
+        (true, Some(..)) => bail!("no target expected by prep_image_update"),
+        (false, None) => bail!("must provide target for prep_image_update"),
+        (true, None) => None,
+        (false, Some(t)) => Some(("image_type", IdolArgument::String(t))),
```
jgallagher, line 88: Nit - instead of the `.as_ref().map_or()` below, could we put the args in these two arms?

```rust
(true, None) => &[],
(false, Some(t)) => &[("image_type", IdolArgument::String(&subargs.target)],
```

jgallagher (approved):

cmd/update/src/lib.rs
mkeeter: Despite my best efforts, I couldn't find a way to make this work!

The code that you wrote fails because `&[]`, `&[("image_type", ...)]` don't automatically coerce to slices:
```
   = note: expected reference `&[_; 0]`
              found reference `&[(&str, IdolArgument<'_>); 1]`
```

However, if you _do_ coerce them to slices with
```rust
        (true, None) => [].as_slice(),
        (false, Some(t)) => &[("image_type", IdolArgument::String(t))],
```
then it fails due to a borrow lifetime issue:
```
error[E0716]: temporary value dropped while borrowed
  --> cmd/update/src/lib.rs:88:30
   |
84 |     let args = match (start.operation.args.is_empty(), &subargs.target) {
   |         ---- borrow later stored here
...
88 |         (false, Some(t)) => &[("image_type", IdolArgument::String(t))],
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^-
   |                              |                                       |
   |                              |                                       temporary value is freed at the end of this statement
   |                              creates a temporary value which is freed while still in use
   |
help: consider using a `let` binding to create a longer lived value
```

PR #388, opened 2023-05-25
Add cmd for interacting with the 'attest' task.

flihp:
This implementation is currently limited to obtaining the cert chain for the Alias keypair held by the attest task.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1094,6 +1094,7 @@ dependencies = [
  "humility-cli",
  "humility-cmd",
  "humility-cmd-apptable",
+ "humility-cmd-attest",
  "humility-cmd-auxflash",
  "humility-cmd-bankerase",
  "humility-cmd-console-proxy",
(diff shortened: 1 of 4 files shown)


flihp:
This commit requires https://github.com/oxidecomputer/hubris/pull/1378 otherwise there's no `attest` task for it to talk to. To test this new command and validate the certs on an RoT try:
```sh
#!/usr/bin/env bash

ARCHIVE=path/to/archive.zip
CA_FILE=path/to/root/cert.pem

# Get the cert chain starting from the Alias cert / leaf and ending at the
# last intermediate before the root (using ordering from RFC 5246 § 7.4.2).
# Split the chain up into individual certs and name them according to their
# position in the chain.
humility --archive "$ARCHIVE" \
    attest cert-chain 2> /dev/null \
| csplit --elide-empty-files \
    --prefix "cert-" \
    --suffix-format "%02d.pem" \
    - '/-----BEGIN CERTIFICATE-----/' '{*}' > /dev/null

CHAIN_LEN=$(humility --archive "$ARCHIVE" attest cert-chain-len 2> /dev/null)

# Use the possitional parameter variable to build up the '-untrusted'
# parameters to `openssl verify`. The cert chain starts at the leaf and
# ends at the last intermediate before the root so we iterate backward.
for i in $(seq $(("$CHAIN_LEN" - 1)) -1 1); do
    set -- "$@" -untrusted "$(printf "cert-%02d.pem\n" "$i")"
done

CMD="openssl verify -ignore_critical -CAfile $CA_FILE $@ cert-00.pem"
eval "$CMD" > /dev/null
if [ $? -eq 0 ]; then
    echo "cert chain is valid"
else
    echo -e "cert chain *failed* validation w/ command:\n    $CMD"
    exit 1
fi
```

flihp:
Closing this in favor of using the `hiffy` interface to the attest task. This pushes the work implemented in this PR out of humility and into the consumer & whatever wrapper scripting they've got.

PR #391, opened 2023-06-08
"humility power", "humility rendmp --phase-check" tweaks

### LINKED ISSUE #390: "humility rendmp --phase-check" on VDD_MEM_ABCD can lock I2C bus
On some rails on some machines, running `humility rendmp --phase-check` results in a very locked I2C bus.  To the user, this will look like this:

```console
$ humility rendmp --phase-check --device 0x5a
Phase check for RAA229618 at 0x5a
PHASE |   VOUT   |   IOUT   |   TEMP   | RAIL
------|----------|----------|----------|----------
humility: restoring default configuration...
humility: WARNING: error restoring default cfg when recovering from error: failed to restore default cfg

Caused by:
    BusReset
humility rendmp failed: failed to READ_VOUT: ControllerBusy
```

At this point, the bus is borked, e.g.:

```console
$ humility i2c -s -b mid

Device scan on controller I2C3:

    R = Reserved   - = No device   \o/ = Device found   X = Timed out

ADDR     0x0 0x1 0x2 0x3 0x4 0x5 0x6 0x7 0x8 0x9 0xa 0xb 0xc 0xd 0xe 0xf
0x00       R   R   R   R   R   R   R   R Err Err Err Err Err Err Err Err
0x10     Err Err Err Err Err Err Err Err Err Err Err Err Err Err Err Err
0x20     Err Err Err Err Err Err Err Err Err Err Err Err Err Err Err Err
0x30     Err Err Err Err Err Err Err Err Err Err Err Err Err Err Err Err
0x40     Err Err Err Err Err Err Err Err Err Err Err Err Err Err Err Err
0x50     Err Err Err Err Err Err Err Err Err Err Err Err Err Err Err Err
0x60     Err Err Err Err Err Err Err Err Err Err Err Err Err Err Err Err
0x70     Err Err Err Err Err Err Err Err Err Err Err Err   R   R   R   R

Error summary:

  COUNT ERROR
    116 BusReset

```

Here is logic analyzer output of the mid bus when it locks:

2023-05-26-21-19-53_2886x1094

From this point on, SDA is held low and SCL is high.  Of note, resetting the SP does *not* result in the bus being let go -- and indeed, the SCL wiggles that are a part of I2C initialization do not result in SDA being released, e.g.:

2023-05-26-21-23-38_2177x1094

A few things to note about this:

- This always happens immediately after 0 has been written to `ON_OFF_CONFIG`, enabling the rail in PWM mode
- There is always a ~150  μsec delay between the final stop condition that is the write to `ON_OFF_CONFIG` and SDA being pulled low
- On Gimlet, this has only been seen on `VDD_MEM_ABCD` -- and not on `VDD_VCORE` (which has different phases but otherwise shares the same controller)
- It happens on about 80% of the Gimlets that I tried (both C and B) -- but there *are* Gimlets that are able to perform `--phase-check` cleanly on all rails
- The only thing that appears to clear the condition is a return to A3 (that is, powering off the RAA229618).

bcantrill:
Fixes #390. Draft comment:

```
This fixes #390 and makes some other small tweaks to "humility power"
(taking a rail as an argument) and "humility rendmp --phase-check"
(taking a phase as an argument + cleaning up output a tad).
```


cmd/rendmp/src/lib.rs
```diff
@@ -1520,20 +1553,49 @@ fn rendmp_phase_check<'a>(
         bail!("must be in A2 when checking phases");
     }
 
+    if cmd_spd::spd_any(hubris, core)? {
```
mkeeter, line 1556: Nit: In my Humility reorganization, I've been trying to avoid command-to-command dependencies, since that introduces a longer dependency chain when building.  IMHO the Right Way™ to do this would be to add a new crate (`humility-spd`), then use it both here and in `cmd-spd`.

I don't feel strongly enough about this to hold up the PR, though.

mkeeter (approved):

cmd/rendmp/src/lib.rs
bcantrill: I'm going to split the difference on this and add a comment to make clear that this is not a practice to be emulated!

PR #392, opened 2023-06-08
readme: add link to huburis

will:
There was a link to Humility in a blog post, and since it wasn't the point of the blog post that post did not explain what Humility was. So I clicked that link and came to this repo. This repo's readme says that it's a debugger for Hubris and assumes you know what Hubris is, which I would argue is hubris in and of itself. I had no idea what it was. I took a lucky guess that it would be another project on the same org and was correct. I think that hyperlink to that other project would be both an easy and natural way for others to not have to make that same guess.


bcantrill (approved):
Looks great -- thank you!

will:
Oh okay. That means I'll have to actually check this out. One second

will:
I tried. It added my new link, but then deleted the last 2/3 of the readme, and I'm not sure why, so I manually did the change. 

I enjoyed your "don't anthropomorphize the lawnmower" digression, by the way

PR #398, opened 2023-06-22
UdpDumpAgent does not check image ID, sowing confusion

bcantrill:
Draft comment:

```
This fixes humility#395 and (as an unrelated change) also allows
sensors to be specified by ID -- and for "sensors --list" to be run
from either an archive or dump.
```


Cargo.toml
```diff
@@ -98,7 +98,7 @@ members = [
 [workspace.dependencies]
 # `git`-based deps
 hif = { git = "https://github.com/oxidecomputer/hif" }
-humpty = { git = "https://github.com/oxidecomputer/humpty", version = "0.1.2" }
+humpty = { git = "https://github.com/oxidecomputer/humpty", version = "0.1.3", branch = "imageid" }
```
mkeeter, line 101: Reminder to fix this before merging

cmd/sensors/src/lib.rs
```diff
@@ -80,36 +80,54 @@ struct SensorsArgs {
...
-    named: &Option<HashSet<&String>>,
-) -> Result<()> {
+enum Spec<'a> {
+    Params {
+        types: &'a Option<HashSet<HubrisSensorKind>>,
+        devices: &'a Option<HashSet<&'a String>>,
+        named: &'a Option<HashSet<&'a String>>,
+    },
+    Id(HashSet<usize>),
+}
```
mkeeter, line 100: Putting this into a type makes sense to me; I'd suggest taking it a step further and making "should we accept this sensor" into a method on that type.

Here's one way to doing that:
```diff
diff --git a/cmd/sensors/src/lib.rs b/cmd/sensors/src/lib.rs
index 9260f08c..bc84b7e6 100644
--- a/cmd/sensors/src/lib.rs
+++ b/cmd/sensors/src/lib.rs
@@ -99,6 +99,40 @@ enum Spec<'a> {
     Id(HashSet<usize>),
 }
 
+impl Spec<'_> {
+    fn matches(
+        &self,
+        hubris: &HubrisArchive,
+        s: &HubrisSensor,
+        ndx: usize,
+    ) -> bool {
+        match self {
+            Spec::Params { types, devices, named } => {
+                if let Some(types) = types {
+                    if !types.contains(&s.kind) {
+                        return false;
+                    }
+                }
+                if let Some(devices) = devices {
+                    if let HubrisSensorDevice::I2c(device) = s.device {
+                        let device = &hubris.manifest.i2c_devices[device];
+                        if !devices.contains(&device.device) {
+                            return false;
+                        }
+                    }
+                }
+                if let Some(named) = named {
+                    if !named.contains(&s.name) {
+                        return false;
+                    }
+                }
+                true
+            }
+            Spec::Id(ids) => ids.contains(&ndx),
+        }
+    }
+}
+
 fn list(hubris: &HubrisArchive, spec: &Spec) -> Result<()> {
     println!(
         "{:3} {:5} {:<13} {:>2} {:>2} {:3} {:4} {:13} {:4}",
@@ -106,32 +140,13 @@ fn list(hubris: &HubrisArchive, spec: &Spec) -> Result<()> {
     );
 
     for (ndx, s) in hubris.manifest.sensors.iter().enumerate() {
-        if let Spec::Params { types: Some(types), .. } = spec {
-            if types.get(&s.kind).is_none() {
-                continue;
-            }
-        }
-
-        if let Spec::Id(ids) = spec {
-            if ids.get(&ndx).is_none() {
-                continue;
-            }
+        if !spec.matches(hubris, s, ndx) {
+            continue;
         }
 
         match &s.device {
             HubrisSensorDevice::I2c(device) => {
                 let device = &hubris.manifest.i2c_devices[*device];
-                if let Spec::Params { devices: Some(devices), .. } = spec {
-                    if devices.get(&device.device).is_none() {
-                        continue;
-                    }
-                }
-
-                if let Spec::Params { named: Some(named), .. } = spec {
-                    if named.get(&s.name).is_none() {
-                        continue;
-                    }
-                }
 
                 let mux = match (device.mux, device.segment) {
                     (Some(m), Some(s)) => format!("{}:{}", m, s),
@@ -197,34 +212,9 @@ fn print(
     let mut sensors = vec![];
 
     for (i, s) in hubris.manifest.sensors.iter().enumerate() {
-        if let Spec::Id(ids) = spec {
-            if ids.get(&i).is_none() {
-                continue;
-            }
+        if !spec.matches(hubris, s, i) {
+            continue;
         }
-
-        if let Spec::Params { types: Some(types), .. } = spec {
-            if types.get(&s.kind).is_none() {
-                continue;
-            }
-        }
-
-        if let Spec::Params { devices: Some(devices), .. } = spec {
-            if let HubrisSensorDevice::I2c(i) = &s.device {
-                let d = &hubris.manifest.i2c_devices[*i];
-
-                if devices.get(&d.device).is_none() {
-                    continue;
-                }
-            }
-        }
-
-        if let Spec::Params { named: Some(named), .. } = spec {
-            if named.get(&s.name).is_none() {
-                continue;
-            }
-        }
-
         sensors.push((i, s));
     }
```

humility-dump-agent/src/udp.rs
```diff
@@ -79,6 +79,37 @@ impl<'a> UdpDumpAgent<'a> {
 
         Ok(reply)
     }
+
+    pub fn check_imageid(&mut self, image_id: &Vec<u8>) -> Result<()> {
+        let r = self.dump_remote_action(humpty::udp::Request::GetImageId)?;
+
+        match r {
+            Err(humpty::udp::Error::VersionMismatch { ours, theirs })
+                if ours == 2 && theirs > 2 =>
```
mkeeter: Maybe add a comment here explaining that `ours` and `theirs` are swapped from our perspective (and/or an `assert_eq!(theirs, humpty::udp::version::CURRENT)`, to drive the point home)

bcantrill: Heh -- look at the latest. ;)

humility-dump-agent/src/udp.rs
```diff
@@ -79,6 +79,37 @@ impl<'a> UdpDumpAgent<'a> {
 
         Ok(reply)
     }
+
+    pub fn check_imageid(&mut self, image_id: &Vec<u8>) -> Result<()> {
```
mkeeter, line 83: It looks like we always call `UdpDumpAgent::new` followed by `dump_agent.check_imageid`, so maybe just fold this into the constructor (making it return a `Result<Self>`)?

humility-dump-agent/src/udp.rs
bcantrill: Added the additional check, but made it a `bail!` rather than a assert (because it points to a broken Hubris, not a broken Humility).

mkeeter (approved):

PR #399, opened 2023-06-22
Make `udprpc` optional in `NetCore`

mkeeter:
Production images don't have `udprpc`, but we can still use `dump-agent` to read memory remotely.

### DIFF
--- a/humility-net-core/src/lib.rs
+++ b/humility-net-core/src/lib.rs
@@ -55,12 +55,7 @@ impl NetCore {
 
         let scopeid = decode_iface(iface)?;
 
-        let udprpc_socket = if let Some(t) = hubris.lookup_task("udprpc") {
-            // Look up the reply type just to make sure it exists
-            let _rpc_reply_type = hubris
-                .lookup_module(*t)?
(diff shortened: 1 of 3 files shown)


humility-net-core/src/lib.rs
```diff
@@ -314,7 +312,13 @@ impl Core for NetCore {
 
     fn send(&self, buf: &[u8], target: NetAgent) -> Result<usize> {
         match target {
-            NetAgent::UdpRpc => self.udprpc_socket.send(buf),
+            NetAgent::UdpRpc => {
+                if let Some(d) = self.udprpc_socket.as_ref() {
+                    d.send(buf)
+                } else {
+                    bail!("no `udprpc` socket");
```
bcantrill, line 319: Not sure if this is a message that an errant user could hit, but do we want to prompt around the image type as in the other error messages?

bcantrill (approved):

humility-net-core/src/lib.rs
mkeeter: I _think_ this isn't hittable by users, because we check during `HiffyContext` construction whether the `udprpc` task exists and bail out otherwise.

PR #407, opened 2023-07-26
Fail if `humility readvar` doesn't find the variable

mkeeter:
Previously, it would return successfully, without any warnings.


lzrd (approved):
A friendlier error message on name not found. Looks good.

PR #409, opened 2023-07-27
Check auxflash in `humility flash`

### LINKED ISSUE #408: humility flash should verify auxflash contents independent of whether primary ROM was intact
For archives that contain not just a primary hubris image but also an auxiliary flash image, `humility flash` will write the auxflash image into the next slot iff it is not already present.  It appears that we presently do not _verify_ that the auxflash image is present and complete iff the primary hubris image is completely intact in the ROM at the time of the check.

During manufacturing, when programming the SP, we perform essentially this sequence of commands:

```bash
if ! humility flash --check; then
        printf 'contents incorrect; programming...\n'
        if ! humility flash --verify; then
                printf 'programming failed!\n'
                exit 1
        fi
fi
printf 'ok\n'
```

On a programming station, we were having a reproducible issue that occurred each time we tried to erase and write an auxiliary flash slot on a **Sidecar/B**.  The precise nature of the issue is not germane; the upshot is that the hubris image was written, but the auxflash write was left incomplete.

When the issue was corrected and we went to retry programming, it appears that `humility flash --check` confirmed that the hubris image in the device had a valid checksum and then _did not check the auxflash_.  When we went to power on the system, the system was unable to start because the FPGA image was not available.

A manual `humility flash -F` corrected the issue: it reported that it did not _need_ to write the hubris image again, but did anyway; it then did the auxflash update correctly.

With my apologies if this has already been fixed in a later version, this behaviour was seen with:

```
$ humility --version
humility 0.10.24
```

mkeeter:
This adds an auxflash check to `humility flash`:

- `humility flash` will now reflash if the image matches but the auxflash is not loaded
  - This is a _little_ inefficient, because it will also re-flash the main firmware, but was easiest to implement
- `humility check` will now return an error if the auxflash is not loaded

We never actually read back the auxflash; we verify the Hubris image, and the Hubris image contains a checksum for the auxflash.  If the auxflash was invalid, then we would not see an active slot in the running Hubris image.

Fixes #408

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1080,7 +1080,7 @@ dependencies = [
 
 [[package]]
 name = "humility"
-version = "0.10.25"
+version = "0.10.26"
 dependencies = [
  "anyhow",
  "bitfield",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ name = "humility"
 #
 # Be sure to check in and push all of the files that change.  Happy versioning!
 #
-version = "0.10.25"
+version = "0.10.26"
 authors = ["Bryan Cantrill <bryan@oxide.computer>"]
 edition = "2018"
 license = "MPL-2.0"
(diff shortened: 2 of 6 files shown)


jclulow:
Thanks, that sounds great!

PR #415, opened 2023-08-22
Add readmem option to save to file

labbott:
It's useful to be able to save arbitrary memory ranges as a binary like gdb/pyOCD/openocd do

### DIFF
--- a/cmd/readmem/src/lib.rs
+++ b/cmd/readmem/src/lib.rs
@@ -121,6 +121,8 @@ use humility::hubris::*;
 use humility_cli::{ExecutionContext, Subcommand};
 use humility_cmd::{Archive, Attach, Command, CommandKind, Dumper, Validate};
 use std::convert::TryInto;
+use std::io::Write;
+use std::path::PathBuf;
 
 //
 // We allow the size to be specified as a number (e.g., with an optional `0x`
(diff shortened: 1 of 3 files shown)


cmd/readmem/src/lib.rs
```diff
@@ -199,6 +205,24 @@ fn readmem(context: &mut ExecutionContext) -> Result<()> {
...
 
+    if let Some(file) = subargs.file {
+        let mut f = std::fs::File::create(&file)?;
+        for (i, addr) in (addr..addr + (length as u32)).step_by(max).enumerate()
+        {
+            let mut bytes = vec![
+                0u8;
+                std::cmp::min(
+                    max,
+                    length.checked_sub(i * max).unwrap_or(length)
```
mkeeter, line 216: Can this ever actually underflow?  I can't think of a way that `i * max` would ever exceed `length`.

labbott: I think you're right and I was being too cautious here

cmd/readmem/src/lib.rs
```diff
@@ -199,6 +205,24 @@ fn readmem(context: &mut ExecutionContext) -> Result<()> {
         bail!("address must be {}-byte aligned", size);
     }
 
+    if let Some(file) = subargs.file {
+        let mut f = std::fs::File::create(&file)?;
+        for (i, addr) in (addr..addr + (length as u32)).step_by(max).enumerate()
+        {
+            let mut bytes = vec![
```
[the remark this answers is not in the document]
labbott: Oh that's spffiy

PR #416, opened 2023-08-30
Add new hiffy based testing infrastructure

labbott:


labbott:
See https://github.com/oxidecomputer/hubris/pull/1512 for the associated hubris changes

cmd/test/src/lib.rs
```diff
@@ -227,441 +207,126 @@ impl fmt::Display for TestRunResult {
...
+        println!("{:#}", result);
+        writeln!(
+            out,
+            "==== Test {} result: {:?}",
+            test_name,
+            result.to_str_no_color()
+        )?;
+        writeln!(out, "==== Task state")?;
+
+        cmd_tasks::print_tasks(
```
bcantrill, line 325: I think that this is fine for now, but we may want to stick this into a different crate if we find we're using it a bunch....

bcantrill (changes requested):
This looks great, and removes a bunch of gross code.  I would also like us to update the docs (which make reference to ITM -- and then have to explicitly apologize for its sporadic misbehavior), and then also bump the version.

bcantrill (approved):
Looks awesome -- thank you!

PR #418, opened 2023-09-27
Online debug authentication with permission-slip

plotnick:
For debugging locked (production) roots of trust, we must use LPC55 debug authentication to protect our secrets. Debug authentication requires access to a credential certificate which must be signed by an (offline) root, and an online debug credential key which we would like to treat as sensitive (but not as sensitive as a root). [Permission Slip](https://github.com/oxidecomputer/permission-slip) lets us keep the sensitive keys locked down (in AWS KMS) while still being able to sign various kinds of artifacts; oxidecomputer/permission-slip#102 adds support for LPC55 debug authentication artifact types.

This change adds a new `debugmailbox auth-online` sub-command that gets a debug authentication challenge from an LPC55, sends it to the permslip server (the "online signing service"), receives the signed response, and writes it back to the LPC55. It is like a combination of the commands `debugmailbox auth-challenge foo.dac && lpc55_sign debug-auth-response ... foo.dar && debugmailbox auth-response foo.dar`, but without requiring direct access to the debug credential private key:
```
lamia:~/oxide/humility% humility reset
humility: Opened probe MCU-LINK (r0FF) CMSIS-DAP V2.263
lamia:~/oxide/humility% alias beacon='humility readmem -H 0x50000fc0 4'
lamia:~/oxide/humility% beacon
               \/    2    4    6    8    a    c    e
0x50000fc0 | 0000 0000                               | ....            
lamia:~/oxide/humility% humility debugmailbox auth-online bart --beacon=0x5678
Looks like a plausible debug mailbox
Resetting chip via SYSREQRESET!
lamia:~/oxide/humility% beacon                                                
               \/    2    4    6    8    a    c    e
0x50000fc0 | 2710 5678                               | .'xV            
```

### DIFF
--- a/cmd/debugmailbox/src/lib.rs
+++ b/cmd/debugmailbox/src/lib.rs
@@ -21,11 +21,17 @@
 //! entered ISP mode!
 //! ```
 
-use std::{fs::File, path::PathBuf};
+use std::{
+    fs::File,
+    path::PathBuf,
+    process::{Command as Process, Stdio},
+    thread::spawn,
+};
 
 use anyhow::{bail, Context, Result};
 use byteorder::{ByteOrder, LittleEndian, WriteBytesExt};
 use clap::{CommandFactory, Parser};
+use clap_num::maybe_hex;
 use humility_cli::{ExecutionContext, Subcommand};
 use humility_cmd::{Archive, Command, CommandKind};
 use probe_rs::{
@@ -82,6 +88,14 @@ enum DebugMailboxCmd {
     AuthChallenge { out: PathBuf },
     /// Send a Debug Authentication Response
     AuthResponse { dar: PathBuf },
+    /// Debug Authentication Challenge/Response with online signing via `permslip`
+    AuthOnline {
+        /// Debug credential key name (try `permslip list-keys -t`)
+        key_name: String,
+        /// Authentication beacon (UM11126 §51.7)
+        #[clap(long, default_value_t = 0, value_parser=maybe_hex::<u16>)]
+        beacon: u16,
+    },
 }
 
 fn poll_raw_ap_register(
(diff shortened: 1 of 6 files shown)


cmd/debugmailbox/src/lib.rs
```diff
@@ -344,6 +358,58 @@ fn debugmailboxcmd(context: &mut ExecutionContext) -> Result<()> {
...
+            // Ask permission-slip to sign it.
+            let mut permslip = Process::new("permslip")
+                .arg("sign")
+                .arg(key_name)
+                .arg("--kind=debug-authn-challenge")
+                .arg(format!("--beacon={beacon}"))
+                .stdin(Stdio::piped())
+                .stdout(Stdio::piped())
+                .spawn()
+                .context("Unable to execute `permslip`, is it in your PATH and executable?")?;
```
labbott, line 380: I almost wish we could just use the permslip client library but given everything is async I think this is probably the best solution. It would be good to document this dependency somewhere, maybe in the doc at the top.

labbott (commented):
Any reason this wouldn't work on windows?

plotnick:
> Any reason this wouldn't work on windows?

I can't think of any inherent reasons, but I don't really know or have easy access to a system to test it on. AFAIU, Rust's `std::process::Command` supports Windows, and I _think_ `permslip` runs on Windows, but we should verify that, too.

cmd/debugmailbox/src/lib.rs
plotnick: So, I started going down that path, but immediately ran into build problems. The main problem is that `permission-slip-client` currently makes no distinction between the client library (which we could use here) and the client binary (which has dependencies that conflict with humility's). This was the fast and easy way to get it working; when we do the split (fix oxidecomputer/permission-slip#103), then we can link in just the client library (and an async runner, etc.).

labbott (approved):
LGTM!

PR #420, opened 2023-10-05
Update Rust toolchain

plotnick:
Use the `1.72.1` channel, which is the latest `stable` version as of 2023-10-05.

Should fix the CI build failures seen in #418, including clippy.

### DIFF
--- a/cmd/doc/build.rs
+++ b/cmd/doc/build.rs
@@ -48,6 +48,7 @@ fn main() -> Result<()> {
     write!(
         output,
         r##"
+#[allow(clippy::needless_raw_string_hashes)]
 fn cmd_docs(lookup: &str) -> Option<&'static str> {{
     let mut m = HashMap::new();
 "##
@@ -107,6 +108,7 @@ fn cmd_docs(lookup: &str) -> Option<&'static str> {{
     m.get(lookup).as_ref().map(|result| result as _)
 }}
 
+#[allow(clippy::needless_raw_string_hashes)]
 fn docs() -> &'static str {{
     r##""##
     )?;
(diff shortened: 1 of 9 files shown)


cmd/sbrmi/src/lib.rs
```diff
@@ -212,7 +212,7 @@ fn cpuid(
     }
     #[derive(Clone)]
     struct WorkspaceRef<'a, 'b> {
-        cell: Arc<RefCell<SbrmiWorkspace<'a, 'b>>>,
+        cell: Rc<RefCell<SbrmiWorkspace<'a, 'b>>>,
```
plotnick, line 215: I originally just silenced [the clippy warning](https://rust-lang.github.io/rust-clippy/master/index.html#/arc_with_non_send_sync), but on reading the description of the lint decided to try replacing `Arc` with `Rc`. That compiles, the tests pass, and clippy is happy. But I do not understand what this code is doing or why, so if someone that does could please double-check that this is acceptable, I'd appreciate it.

cmd/doc/build.rs
```diff
@@ -48,6 +48,7 @@ fn main() -> Result<()> {
     write!(
         output,
         r##"
+#[allow(clippy::needless_raw_string_hashes)]
```
plotnick, line 51: I don't think generated code needs to worry about how many hashes surround its strings.

cmd/sbrmi/src/lib.rs
labbott: I think this is okay but this is probably a @bcantrill question

jgallagher: This looks fine to me. I think @mkeeter put this in to replace some unsafe hackery once the `raw-cpuid` crate added some functionality, and there's no `unsafe` left so it's fine to rely on clippy. It (clippy) is certainly correct that `Arc<RefCell<_>>` is very suspicious, since `Arc` is for sharing with multiple threads and `RefCell` is not threadsafe.

labbott (approved):
LGTM thanks for doing this!

PR #426, opened 2023-11-01
Add `humility host` subcommand

### LINKED ISSUE #425: Pretty-print `ipcc_panic_data`
It would be nice to pretty-print `LAST_HOST_PANIC` based on [`struct ipcc_panic_data`](https://github.com/oxidecomputer/illumos-gate/blob/24d1bae12beef37addeb22401ae8d98d302f4021/usr/src/uts/oxide/sys/kernel_ipcc.h#L72-L73)

(we could also build an automatic workaround for https://github.com/oxidecomputer/hubris/issues/1554)

mkeeter:
This saves us from having to decode `LAST_HOST_PANIC` with our eyeballs (closes #425)

### DIFF
--- a/cmd/host/src/lib.rs
+++ b/cmd/host/src/lib.rs
@@ -60,7 +60,7 @@ use zerocopy::FromBytes;
 use humility::{core::Core, hubris::HubrisArchive};
 use humility_cli::{ExecutionContext, Subcommand};
 use humility_cmd::{Archive, Attach, Command, CommandKind, Validate};
-use humility_log::msg;
+use humility_log::{msg, warn};
 
 #[derive(Parser, Debug)]
 enum HostCommand {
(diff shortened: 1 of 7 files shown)


cmd/host/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+        "ipd_cause:   {}",
+        match p.ipd_cause {
+            0xca11 => "IPCC_PANIC_CALL",
+            0xa900 => "IPCC_PANIC_TRAP",
+            0x5e00 => "IPCC_PANIC_USERTRAP",
+            0xeb00 => "IPCC_PANIC_EARLYBOOT",
+            0xeb97 => "IPCC_PANIC_EARLYBOOT_PROM",
+            0xeba9 => "IPCC_PANIC_EARLYBOOT_TRAP",
+            0xebff => "IPCC_PANIC_EARLYBOOT_*",
+            b => bail!("unknown `ipd_cause` {b:#04x}"),
```
jgallagher, line 190: Tiny nit - since we already confirmed `ipd_version` above, maybe don't `bail!` here and just print this string instead? I think I'd expect additional causes being added would not cause a version number bump (if the structure of the data didn't change). Maybe the same argument applies to the `bonus_byte` workaround above, and that should similarly result in an "unknown cause" but still print the rest of the data.

jgallagher (approved):

cmd/host/src/lib.rs
mkeeter: Sounds good, done in f5168a91fb7e69f3159efb0e15c1164850e1d31b

PR #427, opened 2023-11-02
add more BMR491 decoding

bcantrill:
Based on the discovery of another document (application note 326), this adds a little more BMR491 black box decoding.

### DIFF
--- a/cmd/ibc/src/lib.rs
+++ b/cmd/ibc/src/lib.rs
@@ -16,55 +16,89 @@
 //!
 //! Here's an example:
 //! ```console
-//! $ humility -tsn5 ibc black-box
+//! $ humility ibc black-box
 //! humility: attached to 0483:374e:001B00083156501320323443 via ST-Link V3
 //! FAULT EVENT
 //!   EVENT_INDEX:        0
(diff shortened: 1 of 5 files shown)


cmd/ibc/src/lib.rs
```diff
@@ -324,6 +351,49 @@ impl<'a> IbcHandler<'a> {
...
+
+        // Always print the actual raw bits, even if they are 0.
+        println!("  {: <18}  {:#04x}", "STATUS_MFR", e.status_mfr);
+
+        if let Some(&(_, bits)) = mfr.iter().find(|&&(v, _)| v == status_word) {
+            for (value, meaning) in bits {
+                if e.status_mfr == *value {
+                    println!("    {value:#04x}: {meaning}");
+                }
+            }
```
mkeeter, line 394: Printing unknown values would be good, something like (untested)
```suggestion
            if let Some((value, meaning)) = bits.iter()
                .find(|(value, _)| e.status_mfr == *value)
            {
                println!("    {value:#04x}: {meaning}");
            } else {
                println!("    {value:#04x}: Unknown meaning!");
            }
```

mkeeter (approved):

Issue #430, opened 2023-11-06
Extend support for passing array parameters to hubpack encoded idol operations


flihp:
The Zerocopy encoding is currently the only one that can be used to pass arrays to idol ops ~~but this encoding is not currently used in Hubris~~. The hubpack encoding is now preferred over Zerocopy so any hubris task that wants to take a byte array as a parameter to an idol operation will need this functionality.

flihp:
I keep forgetting that the default encoding for idol ops is `Zerocopy` and not `Hubpack`. A more generic byte array encoding will need to be added while preserving this existing behavior.

PR #431, opened 2023-11-06
Idol array params

### LINKED ISSUE #429: Zerocopy encoding of array parameters passed through hiffy results in unexpected values
The existing implementation accepts an `--arguments` string like `array=57a3` but the array that shows up at the task is `[ 0x35, 0x37, 0x61, 0x33 ]`. I was expecting instead that the task would receive `[ 0x05, 0x07, 0x0a, 0x03 ]`. The [implementation](https://github.com/oxidecomputer/humility/blob/3629bfd6f7371205baeba19d8316ec7974167adb/humility-idol/src/lib.rs#L408) is copying the string `57a3` `as_bytes` into the payload and so the task gets the ASCII values for each character in the string ... or worse if we do something weird with Unicode.

Updating the code to do the conversion from character to byte is straightforward, but the expected string representation will likely need to change to accommodate bytes that must be represented with multiple characters / digits (0xff / 255). The comma / ',' character is often used to separate array elements in their string representation but the hiffy argument parsing code splits the argument string on this character causing confusion. Using the space character / ' ' is probably a reasonable alternative.

The hiffy CLI is pretty much an API and changing the format of this string may break code that relies on the old format. Currently no tasks in the hubris source tree are using the `Zerocopy` encoding so we may be able to make such a change w/o breakage though we can't make guarantees.

flihp:
this PR resolves ~~#429 and~~ #430 and #432

### DIFF
--- a/humility-idol/src/lib.rs
+++ b/humility-idol/src/lib.rs
@@ -610,6 +610,41 @@ fn serialize_arg(
                 bail!("non-newtype structure arguments currently unsupported");
             }
         }
+        HubrisType::Array(&HubrisArray { goff, count }) => {
+            let t = hubris.lookup_type(goff)?;
+            if !matches!(
+                t,
+                HubrisType::Base(HubrisBasetype {
+                    encoding: HubrisEncoding::Unsigned,
+                    size: 1,
+                })
+            ) {
+                bail!(
+                    "array type in {} ({t:?}) not yet supported",
+                    member.name
+                );
+            }
+            match value {
+                IdolArgument::String(value) => {
+                    let bytes = bytes_from_str(value)?;
+                    if bytes.len() != count {
+                        bail!(
+                            "Cannot convert '{value}' to [u8; {count}]; \
+                             wrong length"
+                        );
+                    }
+
+                    for byte in bytes {
+                        hubpack_serialize_append(buf, &byte)?;
+                    }
+
+                    Ok(())
+                }
+                _ => {
+                    bail!("Bad idol argument value");
+                }
+            }
+        }
         _ => {
             bail!("type of {} ({:?}) not yet supported", member.name, t);
         }
--- a/humility-idol/src/lib.rs
+++ b/humility-idol/src/lib.rs
@@ -285,6 +285,43 @@ fn lookup<'a>(
     }
 }
 
+/// Convert array from hiffy `--argument` list string to a vector of bytes.
+///
+/// The humility hiffy cmd accepts multiple encodings of array arguments:
+/// - When passed a string of characters like `--arguments array=5432` the
+/// string is passed to the operation 'as_bytes'. An idol op that takes a
+/// 4 byte array will receive [ 53, 52, 51, 50 ] given the argument string
+/// above. This is intended as a mechanism for passing ASCII characters to a
+/// task.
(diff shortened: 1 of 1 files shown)


humility-idol/src/lib.rs
```diff
@@ -285,6 +285,36 @@ fn lookup<'a>(
     }
 }
 
+// Convert string representation of array from the hiffy `--argument` list
+// to a vector. Arrays must be enclosed in square brackets with array
+// elements separated by the space character.
+// E.g.: '[37 1 255 127]'
```
mkeeter, line 291: Nit: I'm on the fence whether we should support `[37, 1, 255, 127]`, since that's standard Rust syntax.  Perhaps we should just strip `,` from the string before processing it further, so that both flavors work?

humility-idol/src/lib.rs
```diff
@@ -285,6 +285,36 @@ fn lookup<'a>(
     }
 }
 
+// Convert string representation of array from the hiffy `--argument` list
```
mkeeter, line 288: Style nit: since this is a docstring, it should have `///` (since some editors will highlight docstrings differently and show them through LSP hints).

If you want to be Very Strict, there should be a one-line summary followed by details after a blank line, per [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text).

humility-idol/src/lib.rs
```diff
@@ -398,14 +428,15 @@ fn call_arg(
             }
             match value {
                 IdolArgument::String(value) => {
-                    if value.len() != count {
+                    let bytes = bytes_from_str(value)?;
```
mkeeter, line 431: This change will break the use case shown in https://github.com/oxidecomputer/humility/pull/342.

I'd suggest having `bytes_to_str` fall back to the previous behavior if the value isn't bracketed `[]`, i.e. directly interpreting the string as bytes if the length matches.

flihp: I had convinced myself that nothing in hubris relied on this behavior but was obviously wrong! Thanks for catching this. I've updated the PR to allow these two representations to coexist.

humility-idol/src/lib.rs
flihp: Should be addressed in my last push. Hopefully the text makes sense.

humility-idol/src/lib.rs
flihp: I went back and forth on this. I ended up using only the space character because it was the path of least resistance: the argument parsing code that the larger `--arguments` list expects arguments to be separated by commas. Having commas in the array string confuses the `--arguments` parser so there's additional work to make all of that play nice together. I'm happy to spend some cycles on getting to the bottom of this if you prefer.

mkeeter (approved):

humility-idol/src/lib.rs
mkeeter: Looks great, thanks!

humility-idol/src/lib.rs
mkeeter: Gotcha, I don't feel strongly enough to volunteer you for a fight with `clap`'s argument parser.

PR #439, opened 2024-01-16
lsusb: new subcommand for portable USB listings

cbiffle:
It occurred to us this morning that actually obtaining the `pid:vid:serial` combination used by Humility's `-p` argument is complex if you're on an unfamiliar platform. Illumos in particular has no direct equivalent to `lsusb`.

So, here's a subcommand in Humility that uses libusb as the portability layer, and prints things in the appropriate format.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1096,7 +1096,7 @@ dependencies = [
 
 [[package]]
 name = "humility"
-version = "0.10.36"
+version = "0.10.37"
 dependencies = [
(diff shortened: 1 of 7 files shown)


cbiffle:
This PR reflects the effects of `cargo readme` -- it didn't change anything. Patch revision bumped.

cbiffle:
Ah. That comment and the docs are wrong -- the command is `cargo xtask readme`.

PR #440, opened 2024-01-16
DEVELOPING instructions because I always forget

cbiffle:
The main audience for this, if I'm being honest, is me: I work in Humility intermittently enough that each time I return, I find I've forgotten all the little things you have to remember to do.


bcantrill (approved):
This looks awesome -- thank you for doing this!

PR #442, opened 2024-01-31
gimlet inspector command: first version

cbiffle:
This only does hex dumping of the register contents, which requires the user to manually decode the bits -- but since we don't yet have pretty printing code that consumes the register layout JSON, I thought I'd try to get this landed while I did the rest.

### DIFF
--- a/cmd/gimlet/src/lib.rs
+++ b/cmd/gimlet/src/lib.rs
@@ -1,0 +1,149 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! ## `humility gimlet`
+//!
+//! `humility gimlet` contacts a (recent firmware version) Gimlet over the
+//! management network and extracts certain live diagnostic information.
+//!
+//! ```console
+//! $ humility gimlet --ip fe80::0c1d:9aff:fe64:b8c2%en0 read-seq-regs
+//! ```
+//!
+//! For a complete list of subcommands, use `humility gimlet --help`.
+
+use std::net::{SocketAddrV6, UdpSocket};
+use std::time::Duration;
+
+use anyhow::{bail, Context, Result};
+use clap::{ArgGroup, IntoApp, Parser};
+use humility::net::decode_iface;
+use humility_cli::{ExecutionContext, Subcommand};
+use humility_cmd::{Archive, Command, CommandKind, Dumper};
+
+/// This is defined in the gimlet TOML.
+const HARDCODED_PORT: u16 = 23547;
+
+#[derive(Parser, Debug)]
+#[clap(
+    name = "gimlet", about = env!("CARGO_PKG_DESCRIPTION"),
+    group = ArgGroup::new("target").multiple(false)
+)]
+struct Args {
+    /// How long to wait for a response from the target, in milliseconds.
+    #[clap(
+        long, short = 'T', default_value_t = 2000, value_name = "ms",
+        parse(try_from_str = parse_int::parse)
+    )]
+    timeout: u64,
+
+    /// UDP port to contact on the target.
+    #[clap(long, default_value_t = HARDCODED_PORT)]
+    port: u16,
+
+    /// IPv6 address, e.g. `fe80::0c1d:9aff:fe64:b8c2%en0`
+    #[clap(
+        long,
+        env = "HUMILITY_RPC_IP",
+        group = "target",
+        use_value_delimiter = true
+    )]
+    ip: String,
+
+    #[clap(subcommand)]
+    cmd: SubCmd,
+}
+
+#[derive(Parser, Debug)]
+enum SubCmd {
+    /// Read the current contents of the sequencer FPGA registers.
+    ReadSeqRegs,
+}
+
+pub struct Client {
+    socket: UdpSocket,
+    // Allocating this overly-large buffer because we're on the big computer
+    // where RAM is cheap, and it's less likely to need adjustment this way.
+    buf: [u8; 1024],
+}
+
+impl Client {
+    pub fn new(ip: &str, port: u16, timeout: Duration) -> Result<Self> {
+        let Some((ip, scope)) = ip.rsplit_once('%') else {
+            bail!("Missing scope id in IP (e.g. '%en0')")
+        };
+
+        let scopeid = decode_iface(scope)?;
+
+        let dest = SocketAddrV6::new(
+            ip.parse()?,
+            port,
+            0, // flow info
+            scopeid,
+        );
+
+        let socket = UdpSocket::bind("[::]:0")?;
+        socket.set_read_timeout(Some(timeout))?;
+        socket.connect(dest)?;
+
+        Ok(Self { socket, buf: [0; 1024] })
+    }
+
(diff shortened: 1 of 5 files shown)


cmd/gimlet/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+
+#[derive(Parser, Debug)]
+enum SubCmd {
+    /// Read the current contents of the sequencer FPGA registers.
+    ReadSeqRegs,
+}
+
+pub struct Client {
+    socket: UdpSocket,
+    // Allocating this overly-large buffer because we're on the big computer
```
mkeeter, line 59: "overly large" 🤣

cmd/gimlet/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub fn new(ip: &str, timeout: Duration) -> Result<Self> {
+        let Some((ip, scope)) = ip.rsplit_once('%') else {
+            bail!("Missing scope id in IP (e.g. '%en0')")
+        };
+
+        let scopeid = decode_iface(scope)?;
+
+        let dest = SocketAddrV6::new(
+            ip.parse()?,
+            23547,
```
mkeeter, line 74: Add a comment here that this is a hard-coded socket address for the `gimlet-inspector` task

cmd/gimlet/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+fn run(context: &mut ExecutionContext) -> Result<()> {
+    let Subcommand::Other(subargs) = context.cli.cmd.as_ref().unwrap();
+    let subargs = Args::try_parse_from(subargs)?;
+
+    let mut client = Client::new(
+        &subargs.ip,
+        Duration::from_millis(subargs.timeout.into()),
```
mkeeter, line 117: Nit: `Args::timeout` could be a `u64` (instead of a `u32`) , making the conversion unnecessary.

mkeeter (approved):
A few small comments, but LGTM

cmd/gimlet/src/lib.rs
cbiffle: Current max packet size: **2 bytes** ;)

cmd/gimlet/src/lib.rs
cbiffle: Yeah, I copied this from udprpc, which uses a u32 for reasons I don't totally understand. Will fix.

cmd/gimlet/src/lib.rs
cbiffle: I can do 'ya one better:
- Extracted a constant
- Added a comment explaining that it corresponds to the number in the TOML
- Also made it overrideable using a command line argument.

cmd/gimlet/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+    port: u16,
+
+    /// IPv6 address, e.g. `fe80::0c1d:9aff:fe64:b8c2%en0`
+    #[clap(
+        long,
+        env = "HUMILITY_RPC_IP",
+        group = "target",
+        use_value_delimiter = true
+    )]
+    ip: String,
```
hawkw, line 52: I was going to mention that  [`std::net::Ipv6Addr` implements `FromStr`](https://doc.rust-lang.org/stable/std/net/struct.Ipv6Addr.html#impl-FromStr-for-Ipv6Addr), so this could be changed from `String` to `Ipv6Addr` and `clap` would print a nicer error if it can't be parsed. But, I see that we do some additional parsing of the string to extract a scope, so that would be a bit more complex.

We could, potentially, add a new type like
```rust
struct ScopedV6Addr {
   scope: u32,
   addr: Ipv6Addr,
}

impl FromStr for ScopedV6Addr {
    // ...
}
```

but, this might be a bit more effort and it's quite reasonable if you feel like it's out of scope (hah!) for this PR...

cmd/gimlet/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+    match subargs.cmd {
+        SubCmd::ReadSeqRegs => match client.read_sequencer_registers() {
+            Ok(bytes) => {
+                println!("Received register contents follow:");
+                let mut dumper = Dumper::new();
+                dumper.size = 1;
+                dumper.dump(&bytes, 0);
+            }
+            Err(e) => bail!(e),
+        },
```
hawkw, line 138: Style nit: isn't this `match` equivalent to just
```suggestion
        SubCmd::ReadSeqRegs => {
            let bytes = client.read_sequencer_registers()?;
            println!("Received register contents follow:");
            let mut dumper = Dumper::new();
            dumper.size = 1;
            dumper.dump(&bytes, 0);
        },
```

cmd/gimlet/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[derive(Parser, Debug)]
+#[clap(
+    name = "gimlet", about = env!("CARGO_PKG_DESCRIPTION"),
+    group = ArgGroup::new("target").multiple(false)
+)]
+struct Args {
+    /// How long to wait for a response from the target, in milliseconds.
+    #[clap(
+        long, short = 'T', default_value_t = 2000, value_name = "ms",
```
hawkw, line 36: TIL about `value_name`, that's cool!

cmd/gimlet/src/lib.rs
hawkw: I might just go and throw together a separate branch making this change, actually.

cbiffle: The scope-parsing code is now duplicated across probably a half-dozen locations in different repos, so centralizing it would be awesome. Rust std doesn't parse scopes in a useful manner -- it only parses integers, which is a narrow subset of the RFC -- so we wind up needing to do it ourselves.

hawkw: @cbiffle I've opened  #443 to make this change in Humility. In the longer term, I could go ahead and make a (tiny) crate for it, so that we can use it in  other repos as well.

cmd/gimlet/src/lib.rs
cbiffle: Yeah, reckon so. I wrote it the expanded way when I was expecting to do more in the Err branch, but I'm not sure I need to do anything else right now.

cmd/gimlet/src/lib.rs
cbiffle: By default it uses your flag, uppercased, as the value name. I left this lowercase because the rest of Humility seems to use lowercase value names (manpages and Unix commandline help typically use uppercase value names).

cmd/gimlet/src/lib.rs
hawkw: yup, makes sense to me! not a big deal either way!

hawkw (approved):

cmd/gimlet/src/lib.rs
cbiffle: Neat! I see you picked up the `rsplit_once` simplification too. I'll take a look!

hawkw: Yeah, I liked that --- although I went back and forth on it for a sec, because I thought that in the case where the IP is invalid _and_ there's no `%scope`, it might be nicer to error on "hey, 10.0.01 is not an IPv6 address!" than "also you forgot %eth0", but :woman_shrugging: --- i don't know if that actually matters.

PR #443, opened 2024-01-31
Parse `ip%scope` in CLI arguments

hawkw:
Currently, the Humility CLI accepts IP addresses as a `String`, and then
attempts to parse them as an `Ipv6Addr` and scope. This is a bit
unfortunate, as when argument parsing is performed by `clap`, the CLI
will emit a somewhat nicer error message when parsing fails (indicating
which argument the error occurred while parsing, including the invalid
value, and suggesting the help text). On the other hand, when we accept
the argument as a `String`, `clap` doesn't add that useful context to
the error message.

This branch adds a `ScopedV6Addr` newtype, consisting of an `Ipv6Addr`
and a scope identifier, and adds a `FromStr` implementation for that
type. Now, we can use `ScopedV6Addr` as the type of the CLI argument,
and `clap` performs the parsing for us, getting us somewhat nicer error
messages. Compare the error output from master:

```console
$ cargo run -- --ip ::%fakeiface0 probe
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
     Running `target/debug/humility --ip '::%fakeiface0' probe`
humility probe failed: Could not find interface for fakeiface0

$ cargo run -- --ip :: probe
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
     Running `target/debug/humility --ip '::' probe`
humility probe failed: Missing scope id in IP (e.g. '%en0')

$ cargo run -- --ip barf%eth0 probe
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
     Running `target/debug/humility --ip 'barf%eth0' probe`
humility probe failed: invalid Zip archive: Invalid zip header
```

with the error output on this branch:

```console
$ cargo run -- --ip :: probe
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/humility --ip '::' probe`
error: Invalid value "::" for '--ip <IP>': missing scope ID (e.g. "%en0") in IPv6 address

$ cargo run -- --ip ::%fakeiface0 probe
    Finished dev [unoptimized + debuginfo] target(s) in 8.45s
     Running `target/debug/humility --ip '::%fakeiface0' probe`
error: Invalid value "::%fakeiface0" for '--ip <IP>': Could not find interface for fakeiface0

$ cargo run -- --ip barf%eth0 probe
    Finished dev [unoptimized + debuginfo] target(s) in 12.87s
     Running `target/debug/humility --ip 'barf%eth0' probe`
error: Invalid value "barf%eth0" for '--ip <IP>': "barf" is not a valid IPv6 address

```

Additionally, using the `ScopedIpv6Addr` type allows us to provide a
`fmt::Display` implementation that combines the IP address and scope ID
when formatting. This lets us get rid of a little bit of repeated logic
for formatting an address with the delimiter, which seems kinda nice to
me.

### DIFF
--- a/humility-cli/src/lib.rs
+++ b/humility-cli/src/lib.rs
@@ -7,7 +7,7 @@ mod env;
 use anyhow::Result;
 use clap::{AppSettings, ArgGroup, ArgMatches, Parser};
 use env::Environment;
-use humility::{core::Core, hubris::HubrisArchive, msg, warn};
+use humility::{core::Core, hubris::HubrisArchive, msg, net, warn};
 
 #[derive(Parser, Debug, Clone)]
 #[clap(
@@ -61,7 +61,7 @@ pub struct Cli {
     /// HUMILITY_IP environment variable. Run "humility doc" for more
     /// information on running Humility over a network.
     #[clap(long, short, group = "hubris")]
-    pub ip: Option<String>,
+    pub ip: Option<net::ScopedV6Addr>,
 
     /// Hubris environment file. Thie may also be set via the
     /// HUMILITY_ENVIRONMENT environment variable. Run "humility doc" for
@@ -158,7 +158,7 @@ impl ExecutionContext {
             if let Ok(e) = env::var("HUMILITY_PROBE") {
                 cli.probe = Some(e);
             } else if let Ok(e) = env::var("HUMILITY_IP") {
-                cli.ip = Some(e);
+                cli.ip = Some(e.parse()?);
             } else if let Ok(e) = env::var("HUMILITY_TARGET") {
                 cli.target = Some(e);
             } else if let Ok(e) = env::var("HUMILITY_DUMP") {
--- a/humility-core/src/net.rs
+++ b/humility-core/src/net.rs
@@ -5,6 +5,16 @@
 use anyhow::{bail, Context, Result};
 use std::{fmt, net::Ipv6Addr, str::FromStr};
 
+/// An IPv6 address, plus a scope ID.
+///
+/// The Rust standard library's [`Ipv6Addr`] type does not parse scoped IPv6
+/// addresses with [zone indices], which are used to disambiguate link-local
+/// addresses. This type implements [`FromStr`] by parsing an IPv6 address
+/// (using [`Ipv6Addr::from_str`]) and a trailing zone index, which may either
+/// be numeric or an interface name, if the host OS supports this. Interface
+/// names are translated to numeric zone indices using [`decode_iface`].
+///
+/// [zone indices]: https://en.wikipedia.org/wiki/IPv6_address#Scoped_literal_IPv6_addresses_(with_zone_index)
 #[derive(Copy, Clone, Eq, PartialEq)]
 pub struct ScopedV6Addr {
     pub ip: Ipv6Addr,
--- a/humility-core/src/net.rs
+++ b/humility-core/src/net.rs
@@ -2,7 +2,47 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
+use std::{fmt, net::Ipv6Addr, str::FromStr};
+
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub struct ScopedV6Addr {
+    pub ip: Ipv6Addr,
+    pub scopeid: u32,
+}
+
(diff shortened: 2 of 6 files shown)


cbiffle:
> humility probe failed: invalid Zip archive: Invalid zip header

Hoo boy. I can't even explain to myself why this would happen, particularly for `probe` which doesn't use the archive.

humility-cli/src/lib.rs
```diff
@@ -158,7 +158,7 @@ impl ExecutionContext {
             if let Ok(e) = env::var("HUMILITY_PROBE") {
                 cli.probe = Some(e);
             } else if let Ok(e) = env::var("HUMILITY_IP") {
```
cbiffle, line 160: Huh, this is news to me -- some commands use a _different_ env var for this, like `HUMILITY_RPC_IP`. Wonder how that happened.

humility-core/src/net.rs
```diff
@@ -2,7 +2,47 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
+use std::{fmt, net::Ipv6Addr, str::FromStr};
+
+#[derive(Copy, Clone, Eq, PartialEq)]
```
cbiffle, line 8: Please add a doc comment here (doesn't have to be super long, but, what is this thing and why would I reach for it).

hawkw:
> > humility probe failed: invalid Zip archive: Invalid zip header
> 
> Hoo boy. I can't even explain to myself why this would happen, particularly for `probe` which doesn't use the archive.

Yeah, I, uh...have no idea what's going on with that...

cbiffle (approved):
This is an improvement! The two remaining sites of manually copy-pasted scope ID parsing that I'm aware of are in `cmd/rpc` and `cmd/gimlet` -- if you've got the bandwidth to fix those in this PR that'd rock, though IMO it's an improvement either way. Couple suggestions below.

cbiffle:
> Yeah, I, uh...have no idea what's going on with that...

Humility does some manual wrangling of Clap, e.g. not using its built-in subcommand facility, so it's possible we confused it somehow or messed up an error handling path somewhere.

cbiffle:
...more generally, it looks like `--ip` is accepted as a global flag, which is news to the `rpc` command that reimplements it, and causes all commands to start requiring an archive whether they need it or not -- failing to provide an archive gets you this error message, currently, because of assumptions in `NetCore`.

humility-core/src/net.rs
```diff
@@ -5,6 +5,16 @@
 use anyhow::{bail, Context, Result};
 use std::{fmt, net::Ipv6Addr, str::FromStr};
 
+/// An IPv6 address, plus a scope ID.
```
cbiffle, line 8: Nice.

hawkw:
> When no archive is given, Humility still creates a `HubrisArchive` and passes it around, only a bunch of its methods now fail. Some of them don't, however, and in particular `.archive()` appears to return `&[]` rather than, like, being an `Option`. This angers the Zip gods.

Ah, yeah, that makes sense...it's a bit unfortunate that the failure thus ends up being "the zip file is missing a header" rather than "there is no zip file at all".  I could look into fixing that separately, if you like?

hawkw:
> The two remaining sites of manually copy-pasted scope ID parsing that I'm aware of are in `cmd/rpc` and `cmd/gimlet` -- if you've got the bandwidth to fix those in this PR that'd rock,

Now that your `gimlet-inspector` PR has merged, I'll go ahead and pick up those changes so we can get that code using this as well, and I'll see about getting the `rpc` command on board too!

cbiffle:
> Ah, yeah, that makes sense...it's a bit unfortunate that the failure thus ends up being "the zip file is missing a header" rather than "there is no zip file at all". I could look into fixing that separately, if you like?

I'm prepping a fix, unfortunately the place where the failure happens is outside Humility's normal "I have no archive but I must scream" failure path, so the reporting will be a tad shoddy, but I can get it to say

`humility probe failed: archive is required but was not provided`

(The reason it's outside the failure path: `probe` literally does not require an archive, this requirement is coming from the internals of `NetCore`.)

hawkw:
@cbiffle I've updated the `humility gimlet` and `humility rpc` commands to also use the new parsing logic. Let me know if you want to take another look at the latest changes? Otherwise, I'm happy to merge when CI passes.

cbiffle (approved):

PR #444, opened 2024-01-31
Actually, cargo doesn't "mostly just work"

hawkw:
The `DEVELOPING.mkdn` file currently states that `cargo build` and `cargo test` "mostly just works". Actually, when building Humility for the first time, they may not just work, if the `cargo readme` cargo subcommand or `libudev` headers are missing. I've updated the `DEVELOPING.mkdn` file to state that these dependencies are necessary to build Humility.

### DIFF
--- a/DEVELOPING.mkdn
+++ b/DEVELOPING.mkdn
@@ -3,9 +3,17 @@
 This file is intended to collect recipes and procedures for working on Humility,
 oriented toward the casual or intermittent contributor.
 
-## Mostly Cargo Just Works
+## Build dependencies
 
-`cargo build` and `cargo test` will both do the thing you expect.
+In order to build Humility, the following dependencies are required:
+
+- [`cargo readme`](https://crates.io/crates/cargo-readme), which can be
+  installed using `cargo install cargo-readme`
+- `libudev` and `pkg-config` (at least on Linux), which you can install
+  using your package manager of choice.
+
+Once these build dependencies are present, `cargo build` and `cargo test` will
+both do the thing you expect.
 
 ## Bumping the crate version requires an extra step
 


cbiffle:
Oh, fantastic, the implicit reliance on `cargo readme` and lack of versioning have always bugged me.

Cargo.lock
```diff
@@ -344,15 +382,37 @@ checksum = "71655c45cb9845d3270c9d6df84ebe72b4dad3c2ba3f7023ad47c144e4e473a5"
...
+ "indexmap 1.9.3",
  "once_cell",
  "strsim",
  "termcolor",
  "textwrap",
 ]
 
+[[package]]
+name = "clap"
+version = "4.4.18"
```
cbiffle, line 396: Huh, I'm guessing humility itself is still on clap 3? (This is not me asking you to change that -- iirc porting from 3 to 4 is nontrivial)

cmd/doc/build.rs
```diff
@@ -60,40 +59,50 @@ fn cmd_docs(lookup: &str) -> Option<&'static str> {{
...
         write!(output, "        m.insert(\"{}\", r##\"", cmd)?;
 
         let header = format!("### `humility {}`\n", cmd);
 
-        if contents.stdout.len() == 1 {
+        if contents.len() == 1 {
             bail!("no documentation for {cmd}");
         } else {
-            if !contents.stdout.starts_with(header.as_bytes()) {
+            if !contents.starts_with(&header) {
```
cbiffle, line 96: (this check looks redundant now?)

cbiffle (commented):
I like it, fwiw... should get @bcantrill 's eyes on it

bcantrill (approved):
This looks great -- thank you!

cmd/doc/build.rs
hawkw: ah, yup, will remove that!

Cargo.lock
hawkw: yeah, i believe the `cargo-readme` binary depends on `clap` v4, thus the duplicate dep. i'm happy to look into a potential `clap` v4 update separately, if you'd like!

PR #446, opened 2024-02-05
Support bitflags 2.x, fixing `humility dump`

mkeeter:
Reading region attributes (as part of the dump) was broken, because `bitflags 2.x` changed the way that it encodes the inner field.


cbiffle (approved):
Thanks!

PR #447, opened 2024-02-08
Remove misfiring IP address presence check.

cbiffle:
This check assumed that IP addresses are only given for debugging with NetCore, but since @hawkw's improvements recently, that is no longer the case.

Now, this check is actually preventing anyone from using the gimlet subcommand, which is a Detached command.

I glanced over the commands that use this, and I don't think this check is pulling its weight -- I don't think it's worth introducing another piece of command metadata to specify "I'm detached but also I'd like an IP address." So, I've removed it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1195,7 +1195,7 @@ dependencies = [
 
 [[package]]
 name = "humility"
-version = "0.10.38"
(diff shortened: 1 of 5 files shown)


bcantrill (approved):
LGTM -- thank you!

PR #453, opened 2024-03-11
counters: add CSV and JSON output formats

hawkw:
Depends on #454 

This branch adds an `--output` argument to `humility counters`, allowing the selection of either plain human-readable text (the default), JSON, or CSV as the output format, to make it easier to consume `humility counters` data programmatically.  Individual commits have their own descriptions and can be reviewed independently.

Currently, the `--output` flag is not supported by the `counters ipc` subcommand, as `ipc` is just a different way of _presenting_ the same data as the standard `counters` command. I can add CSV or JSON output to that subcommand as well, but would prefer to do it separately.


mkeeter (approved):

bcantrill (approved):
Looks great -- let's bump the version to 0.11.1.

PR #456, opened 2024-03-13
Bump deps to eliminate some yanked packages.

cbiffle:
Concretely, we were getting warnings about
- zerocopy 0.6.1
- hermit-abi 0.3.1
- atomic-polyfill 0.1.11

Had to also update heapless to get everything to resolve.


bcantrill (approved):
Thank you!

hawkw (approved):

PR #458, opened 2024-03-18
"humility tasks" should not abort on failure to read registers

bcantrill:
Draft commit comment:

```
On a stack overflow, we will potentially fail to read all of the
registers.  We don't want "humility tasks" to fail completely in this
case, but rather to warn -- treating the failure to read the registers
the same way as it does the failure to unwind the stack.
```


cmd/tasks/src/lib.rs
```diff
@@ -404,19 +404,29 @@ pub fn print_tasks(
...
-                let regs = hubris.registers(core, t)?;
+                match hubris.registers(core, t) {
+                    Ok(regs) => {
+                        if stack {
+                            let initial = desc.initial_stack;
+
+                            match hubris.stack(core, t, initial, &regs) {
+                                Ok(stack) => printer.print(hubris, &stack),
+                                Err(e) => {
+                                    writeln!(
```
hawkw, line 415: In some other Humility commands, such as `humility ringbuf`, we write errors like this to stderr rather than stdout with the rest of the output. Do you think it would make sense to do that here as well? I realize that this code is writing to a writer rather than directly to stdout, so I'm not really sure if it's great to have a function taking a writer sometimes write directly to stderr...so perhaps this would require more refactoring. Not a blocker for this PR, in any case!

hawkw (approved):

cmd/tasks/src/lib.rs
bcantrill: Yeah, these errors are slightly different in that we want to print them inline and drive on. (That is, you don't want them to stderr because they can be hard to associate with the task that induced them.)  If we do decide to change the way this is formatted, we should make sure that the message emitted upon failure to unwind the stack is handled in the same way.

PR #461, opened 2024-03-19
reflect: best-effort binary format for `bitflags`

### LINKED ISSUE #460: some level of `bitflags` awareness?
Hubris uses the `bitflags` crate in several places. Currently, `humility reflect` will treat types generated using `bitflags` as integers; in output from commands like `humility ringbuf`, they are formatted as hexadecimal. 

This is not awful, but it would be nicer if we could format bitflags types as binary, instead. It would be even nicer if we could format them as the bitwise OR of the named bit fields (e.g. an `IrqStatus` of `0b111` could be formatted as `ENABLED | PENDING | POSTED`), the way the `bitflags`-generated `Debug` impl does, but this might be a bit tricky.

I think that, as of `bitflags` 2.0, `humility reflect` could probably detect `bitflags` -generated types based on whether an `InternalBitFlags` type occurs inside of them, and represent any integers found inside of an `InternalBitFlags` differently. This would, IIUC, only work with types declared using the "struct style" of the `bitflags!` macro, rather than the `"impl style". For example, I believe that:

```rust
// This *will* generate an `InternalBitFlags` struct
bitflags::bitflags! {
    pub struct MyBitflags: u8 {
       const FOO = 1 << 1;
       const BAR = 1 << 2;
   }
}


// This *will NOT* generate an `InternalBitFlags` struct
pub struct MyOtherBitflags(u8);
bitflags::bitflags! {
    impl MyOtherBitflags: u8 {
       const FOO = 1 << 1;
       const BAR = 1 << 2;
   }
}
```
So, we would only end up formatting types defined using the first style as binary values...but that's still better than the current state of things, and bitflags types defined using the "impl style" would just get hex formatting, which is the current state of the world anyway.

hawkw:
Hubris uses the `bitflags` crate in several places. Currently, humility's reflection module treats types generated using `bitflags` the same as all other integers; in output from commands like `humility ringbuf, they are formatted as hexadecimal.

This is not the worst, but it would be nicer if we could format `bitflags` types as binary, instead. That way, a reader need not convert hex to binary to determine which bits are set in a bitflags type.

As of `bitflags` 2.0, types generated using the "struct style" `bitflags!` macro always contain a single field whose type is `InternalBitFlags`. This means that `humility`'s reflection module can detect bitflag types generated using this style of the macro, and format them as binary instead of hex. This PR implements that behavior. For example, in the `task_thermal` ringbuf, we used to see:
```rust
PowerModeChanged(PowerBitmask(InternalBitFlags(0x1)))
...
PowerModeChanged(PowerBitmask(InternalBitFlags(0x2)))
```
but now we see:
```rust
PowerModeChanged(PowerBitmask(0b1))
...
PowerModeChanged(PowerBitmask(0b10))
```

which is IMO quite a bit nicer!

It is worth noting that when the "impl style" of the `bitflags!` macro is used, there is no `InternalBitFlags` struct, as the user declares the struct type directly. In this case, we cannot currently detect that the type would like to be formatted as bits. For example:

```rust
// This *will* generate an `InternalBitFlags` struct
bitflags::bitflags! {
    pub struct MyBitflags: u8 {
       const FOO = 1 << 1;
       const BAR = 1 << 2;
   }
}

// This *will NOT* generate an `InternalBitFlags` struct
pub struct MyOtherBitflags(u8);
bitflags::bitflags! {
    impl MyOtherBitflags: u8 {
       const FOO = 1 << 1;
       const BAR = 1 << 2;
   }
}
```

However, I think this is still better than the current state of things, as we currently format `bitflags` types as hex with _both_ styles of macro invocation.

Closes #460


humility-core/src/reflect.rs
```diff
@@ -564,6 +588,15 @@ impl Format for Tuple {
     ) -> Result<()> {
         fmt.indent += 4;
 
+        // Is this a bitflags-generated type? If so, just format it as a binary
+        // value.
+        if self.name().contains("InternalBitFlags") {
+            if let Some(flags) = self.1.get(0).and_then(|v| v.as_base().ok()) {
+                write!(out, "{flags:#b}")?;
```
hawkw, line 595: perhaps we should always format the type with every zero bit included? I could be convinced either way on that.

bcantrill: I could go either way too -- but I think it looks as is though!

bcantrill (approved):

PR #463, opened 2024-03-20
vpd: support list on firmware predating 2023-05

cbiffle:
rm noticed today that the vpd -l command fails on very old machines, because it has started to rely on the is_locked operation. (As far as I can tell, it has _always_ relied on that operation.)

That operation is sort of incidental to what it's doing, however, so I've made it optional here. You'll still get the table on very old machines, just the locked column won't be useful.

### DIFF
--- a/cmd/vpd/src/lib.rs
+++ b/cmd/vpd/src/lib.rs
@@ -201,28 +201,38 @@ fn list(
 ) -> Result<()> {
     let devices = vpd_devices(hubris).collect::<Vec<_>>();
     let mut context = HiffyContext::new(hubris, core, subargs.timeout)?;
-    let op = hubris.get_idol_command("Vpd.is_locked")?;
     let read_op = hubris.get_idol_command("Vpd.read")?;
 
-    let mut ops = vec![];
+    let locked_op = hubris.get_idol_command("Vpd.is_locked").ok();
+    let results = if let Some(locked_op) = &locked_op {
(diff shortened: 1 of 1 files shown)


cmd/vpd/src/lib.rs
```diff
@@ -201,28 +201,38 @@ fn list(
...
-    ops.push(Op::Done);
+        ops.push(Op::Done);
 
-    let results = context.run(core, ops.as_slice(), None)?;
+        Some(context.run(core, ops.as_slice(), None)?)
+    } else {
+        println!(
+            "Note: firmware does not support the Vpd.is_locked operation."
+        );
+        println!("Lock status will be missing from the table below.");
```
hawkw, line 226: should these maybe go to stderr rather than stdout? maybe they should use the `humility_log::warn!` macro?

hawkw (approved):
looks good to me modulo one tiny nitpick about stderr vs stdout

cmd/vpd/src/lib.rs
cbiffle: I wanted to make sure they appeared right next to the table, if the table was e.g. piped into a file or copy-pasted. Otherwise stderr would have been great... I was just worried they wouldn't get seen and then someone would be like "what is this too old stuff"

I can be talked out of that.

hawkw: makes sense to me!

PR #464, opened 2024-03-22
Add Nix flake

hawkw:
This commit adds a [Nix flake] for building Humility for Nix users, providing a dev environment with build dependencies for use with `nix develop` or `direnv`, and (on NixOS) configuring udev rules to allow all debug probes supported by `probe-rs`. Parts of this flake --- in particular, the code for cross-compiling for multiple targets --- are based on code written by @faithanalog.

## What The Flake Provides

The flake provides the following:

- Packages for `humility` (for the host platform), as well as `humility-aarch64` and `humility-armv7l` for cross-compiling for ARM linux environments. These were in @faithanalog's config as I believe she's got dev boards hooked up to a Raspberry Pi or something that she SSHes into, but would like to build Humility on her dev machine for the Pi target.
- The Linux Packages contains the canonical `udev`` rules published by `probe-rs`, allowing NixOS users to add ```nix services.udev.packages = [ pkgs.humility ]; ``` to their `configuration.nix`` to allow all the debug probes that `probe-rs` supports to be run by non-root users. This config is fetched from the `probe-rs/website` Git repository, so if upstream adds new supported probes, we can pick up their changes.
- A flake `app` for use with `nix run`, which runs a Humility command.
- A dev shell including Humility's build time dependencies, which can be used with `nix develop` or `direnv`
- A NixOS module `programs.humility` which can be used by NixOS users to add Humility to the system's packages, add the udev rules, and add user account(s) to the `plugdev` group to allow access to debug probes.

## On Checking In Nix Configurations

I understand that checking in configurations that are specific to a package manager/distro that not everyone uses may be controversial. And, if folks don't want Nix configs in the main Humility repo, that's understandable, and there are alternative options for Nix users to build Humility --- we can also have a Nix flake in a separate repo that fetches the Humility sources from Git. However, checking in this flake to the Humility repo has some meaningful advantages for Nix users:

1. It allows the Nix build environment for packaging Humility to be used as a dev environment for `direnv` or `nix-develop`. If the Nix configuration is not in the source tree, it can't easily be used with `direnv`. Of course, Nix users can have their own dev environment that isn't checked in. However, this would no longer be the *same* build environment definition that would be used for building a Humility nix package, meaning that there is the potential for drift between individual contributors' dev environments and the canonical build environment.
2. I'd like to eventually set up publishing to [FlakeHub] on CI when new release tags are pushed to the repo, requiring an in-repo Nix flake.
3. I know at least a few folks at Oxide are Nix users (at the very least, myself and @faithanalog), and it seems nice to have a canonical configuration in the repository rather than requiring folks to write their own (potentially differing in minor ways) whenever a NixOS user wants to build or contribute to Humility.

If this rationale isn't convincing, that's okay, and I'm happy to close this PR. :)

[Nix flake]: https://zero-to-nix.com/concepts/flakes
[FlakeHub]: https://flakehub.com/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,5 @@
 /target
+# created by `nix-direnv`
+.direnv
+# created by `nix build`
+/result
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,82 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1705309234,
+        "narHash": "sha256-uNRRNRKmJyCRC/8y1RqBkqWBLM034y4qN7EprSdmgyA=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "1ef2e671c3b0c19053962c07dbda38332dcebf26",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1710951922,
+        "narHash": "sha256-FOOBJ3DQenLpTNdxMHR2CpGZmYuctb92gF0lpiirZ30=",
+        "owner": "nixos",
+        "repo": "nixpkgs",
+        "rev": "f091af045dff8347d66d186a62d42aceff159456",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nixos",
+        "ref": "nixos-23.11",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs",
+        "rust-overlay": "rust-overlay"
+      }
+    },
+    "rust-overlay": {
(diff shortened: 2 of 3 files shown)


faithanalog:
I like the chunky comments for making the different outputs distinct.

I'm assuming that if we wanted to FlakeHub it, we'd need to proactively keep the flake up to date. Right now needing to manually update the output hashes of the git dependencies. I have heard tell of something that might be able to automate updates to them but I don't remember what it is- I can go asking around. I know that I can PR updates to it as necessary but I only pay attention when I'm updating humility on my jumpbox pi. Maybe you'd run into it frequently enough to keep that up to date though.

hawkw:
> I'm assuming that if we wanted to FlakeHub it, we'd need to proactively keep the flake up to date. Right now needing to manually update the output hashes of the git dependencies. I have heard tell of something that might be able to automate updates to them but I don't remember what it is- I can go asking around.

Yeah, figuring this out is why I didn't include publishing in this PR...having to update those hashes manually is unfortunate. Fortunately, a majority of them haven't changed since you published your blog post a while ago.

labbott (approved):
This may have gotten lost but I support adding this  if the Nixos fans agree it's what we want

Issue #466, opened 2024-03-28
humility hangs in hid_close() after failed USB transfer


jclulow:
This `humility` invocation is reliably hanging on **edgar** right now:

```
edgar # pargs $(pgrep humility)
686:    humility hiffy -c Sequencer.get_state
argv[0]: humility
argv[1]: hiffy
argv[2]: -c
argv[3]: Sequencer.get_state
```

Long story short, we're hanging in `hid_close()` waiting for some event that will never arrive:

```
edgar # pstack $(pgrep humility) | demangle
686:    humility hiffy -c Sequencer.get_state
--------------------- thread# 1 / lwp# 1 ---------------------
 fffffc7fe7f4aa7a lwp_wait (6, fffffc7fffdf1abc)
 fffffc7fe7f3fbb8 _thrp_join (6, 0, 0, 1) + 168
 fffffc7fe7f3fc5f pthread_join (6, 0) + 2f
 000000000291dd3b hid_close () + 3b
 0000000002919482 <hidapi::HidDevice as core::ops::drop::Drop>::drop::hc2ceee22d996f5ec () + 12
 0000000002807aa6 probe_rs::probe::cmsisdap::CmsisDap::new_from_device::hc1d7ce7318b3d614 () + 316
 00000000021ef5ea <probe_rs::probe::cmsisdap::CmsisDap as probe_rs::probe::DebugProbe>::new_from_selector::ha22bd7c67f6ff832 () + 12a
 00000000022208d0 probe_rs::probe::Probe::open::h8f67aefa5a7fc578 () + 70
 00000000022688c4 humility_core::core::attach_to_chip::haf79f49127feb2e9 () + 594
 000000000225f8d2 humility_core::core::attach::h9ad2d2f438fdac03 () + 52
 00000000020f6b44 humility_cmd::attach_live::h896f2eec12a397b0 () + 84
 000000000165425d humility_cmd::attach::h84d6f52eae82d20d () + 1fd
 000000000165bdc3 humility::cmd::subcommand::h51158805441e4648 () + 483
 000000000165f9bd humility::main::h2d64e5203bc5454e () + b1d
 0000000001656736 std::sys_common::backtrace::__rust_begin_short_backtrace::h7b1b5b868d3f8ad1 () + 6
 0000000001656781 std::rt::lang_start::_$u7b$$u7b$closure$u7d$$u7d$::hb0bf6b8bb4a678bb (.llvm.15284836038652647814) () + 11
[...]

PR #467, opened 2024-03-28
humility hangs in hid_close() after failed USB transfer

jclulow:
This is a backport of commit 5ce9051e2f2e6501 from upstream hidapi, which adds a missing check for a failed USB transfer.


jclulow:
See also: #466

bcantrill (approved):
Looks great -- thank you for your service and sacrifice!

PR #469, opened 2024-04-02
humility-core: allow enums with signed variants

### LINKED ISSUE #468: Humility assumes enum discriminants are u64, can't parse recent output
The `HubrisEnumVariant` struct assumes that any enum discriminant can fit in a u64. This isn't a valid assumption, and in particular it isn't valid for some common Rust types like [Ordering](https://doc.rust-lang.org/std/cmp/enum.Ordering.html):

```rust
#[repr(i8)]
pub enum Ordering {
    /// An ordering where a compared value is less than another.
    Less = -1,
    /// An ordering where a compared value is equal to another.
    Equal = 0,
    /// An ordering where a compared value is greater than another.
    Greater = 1,
}
```

Enum variants can also be u128s or i128s. There is no single primitive type that is capable of storing all possible enum discriminators. The `tag` field probably needs to become some enum that supports different representations.

This has existed for a while, but I've noticed that more recent versions of rustc have started including such things in the DWARF, and Humility is thus refusing to load archives generated by rustc 1.75.0. This is currently blocking upgrading the toolchain version in Hubris to get some other fixes.

cbiffle:
Humility currently assumes that any enum discriminant can be represented as a u64. This isn't at all true, and in fact fails for common Rust types. See #468 for more details.

This fixes it to handle basic cases of signed enum variants. It's entirely possible that _some_ places in Humility still won't handle them correctly, because logic for loading the enum discriminator isn't reliably centralized in one place, and I may have missed cases.

Fixes #468.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1195,7 +1195,7 @@ dependencies = [
 
 [[package]]
 name = "humility"
-version = "0.11.1"
+version = "0.11.2"
 dependencies = [
  "anyhow",
  "bitfield",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ name = "humility"
 #
 # Be sure to check in and push all of the files that change.  Happy versioning!
 #
-version = "0.11.1"
+version = "0.11.2"
 authors = ["Bryan Cantrill <bryan@oxide.computer>"]
 edition = "2018"
 license = "MPL-2.0"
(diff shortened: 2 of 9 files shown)


bcantrill (approved):
Thanks for doing this, and my apologies for the assumption that Humility was making.  The change looks good -- or at least better than it was with respect to the great disappointment that is the complexity of reality. (Is it worth opening an issue to track the `TODO`s?)

cbiffle:
Good point, filed #471 as a follow-up listing those cases.

PR #472, opened 2024-04-09
hiffy: differentiate dead codes from error codes

hawkw:
When invoking an IPC using `humility-hiffy`, the result is represented as a `u32`. Calling code typically will then attempt to turn that `u32` back into the IPC interface's error code. However, if the server task that Hiffy called into has died, the `u32` will instead be a dead code, which cannot be translated into a named error code using the Idol interface (or similar). Therefore, this branch updates `humility hiffy` to return an `IpcError` enum instead, which is either an error code or a dead code. This forces callers to differentiate between these cases, and not print out random integers when the server task has died.


bcantrill (approved):
Looks good -- please bump the micro version as well!

humility-dump-agent/src/hiffy.rs
```diff
@@ -95,13 +102,15 @@ impl<'a> DumpAgent for HiffyDumpAgent<'a> {
...
-            if let Err(err) = result {
-                bail!(
-                    "failed to add segment at address {:#x} for length {}: {}",
-                    *base,
-                    *size,
-                    op.strerror(*err)
-                );
+            match result {
+                Err(IpcError::Error(err)) => {
+                    bail!("failed to add segment at address {base:#x} for length {size}: {}", op.strerror(*err))
```
mkeeter, line 107: Nit: manually split the string to be <= 80 characters here and below

cmd/validate/src/lib.rs
```diff
@@ -248,6 +248,7 @@ fn validate(context: &mut ExecutionContext) -> Result<()> {
                     Err(anyhow!("unexpected error type {:?}", op.error))
                 }
             }
+            Err(IpcError::ServerDied(e)) => Ok(e.to_string().red()),
```
mkeeter, line 251: This will be a bit cryptic, and I don't think the specific dead code is that helpful – maybe just `"server died".red()` instead?

humility-dump-agent/src/hiffy.rs
```diff
@@ -161,8 +170,12 @@ impl<'a> DumpAgent for HiffyDumpAgent<'a> {
 
         let results = self.run(ops.as_slice())?;
 
-        if let Err(err) = results[rindex] {
-            bail!("failed to take dump: {}", op.strerror(err));
+        match results[rindex] {
+            Err(IpcError::Error(err)) => {
+                bail!("failed to take dump: {}", op.strerror(err))
+            }
+            Err(e) => bail!("failed to take dump: {e}"),
```
mkeeter, line 177: Same comment as above: printing the raw dead code is kinda cryptic here.

(also, I'd vote to spell out `Err(IpcError::ServerDied(..)) => ..` for future-proofing)

humility-hiffy/src/lib.rs
```diff
@@ -56,11 +57,22 @@ pub struct HiffyFunction {
...
-        match self.errmap.get(&code) {
-            Some(name) => name.clone(),
-            None => format!("<Unknown {} error: {}>", self.name, code),
+    pub fn strerror(&self, err: impl Into<IpcError>) -> String {
+        match err.into() {
+            IpcError::Error(code) => match self.errmap.get(&code) {
+                Some(name) => name.clone(),
+                None => format!("<Unknown {} error: {}>", self.name, code),
+            },
+            e => format!("<{e}>"),
```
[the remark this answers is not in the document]
mkeeter: (printing in hex because that makes "this is a dead code" glaringly obvious 😄)

humility-hiffy/src/lib.rs
```diff
@@ -884,7 +896,7 @@ impl<'a> HiffyContext<'a> {
                     }
                 })
             }
-            Err(e) => {
+            Err(IpcError::Error(e)) => {
```
mkeeter, line 899: Can we use `op.strerror` here?

humility-hiffy/src/lib.rs
mkeeter, line 75:
```suggestion
            IpcError::ServerDied(gen) => format!("<server died: {gen}>"),
```

cmd/rendmp/src/lib.rs
```diff
@@ -2274,12 +2274,13 @@ fn rendmp(context: &mut ExecutionContext) -> Result<()> {
         }
     };
 
-    let word_result = |result: &Result<Vec<u8>, u32>, what| -> Result<u32> {
+    let word_result = |result: &Result<Vec<u8>, IpcError>,
+                       what|
```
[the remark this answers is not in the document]
hawkw: `what|`

cmd/validate/src/lib.rs
hawkw: gah, whoops, you're right, this was supposed to call `to_string` on the
`IpcError` type, not on the actual ID. will fix.

humility-hiffy/src/lib.rs
hawkw: Yes, I believe we can --- I didn't do that because this code didn't use it
previously. I'll change it to.

humility-dump-agent/src/hiffy.rs
hawkw: The reason this doesn't do that is because `IdolOperation::strerror` lives in
the `humility-idol` crate, which is a dependency of `humility-hiffy`, which is
where `IpcError` is defined. Although, now that I've actually written that down,
there's no reason that `IpcError` has to be in `humility-hiffy`...

hawkw: Done!

mkeeter (approved):

PR #476, opened 2024-04-16
Support ringbufs with `count: u16`

### LINKED ISSUE #475: Ringbuf decoding failures on a recent image
On `master` (df09cf85a3d70a1d0bb2b7ba688f4fb36a2b11cc), Humility can't decode ringbufs from a recent dump:
```
➜  ~ humility -d hubris.core.gimlet_seq.12 ringbuf
humility: attached to dump
humility: ring buffer drv_gimlet_seq_server::__RINGBUF in gimlet_seq:
humility: ringbuf dump failed: member missing from struct: cell
humility: ring buffer drv_gimlet_seq_server::vcore::__RINGBUF in gimlet_seq:
humility: ringbuf dump failed: not a u32: Base(U16(0))
humility: ring buffer drv_oxide_vpd::__RINGBUF in gimlet_seq:
humility: ringbuf dump failed: not a u32: Base(U16(0))
humility: ring buffer drv_packrat_vpd_loader::__RINGBUF in gimlet_seq:
humility: ringbuf dump failed: not a u32: Base(U16(0))
```

Here's the dump:
[hubris.core.gimlet_seq.16.zip](https://github.com/oxidecomputer/humility/files/14984367/hubris.core.gimlet_seq.16.zip)

mkeeter:
This is a small fix, followed by a zillion lines of new test suite output (because I added a dump with a `u16`-flavored ringbuf).

Fixes #475; bumps the Humility version to 0.11.4


mkeeter:
@hawkw I'm not sure if the `counters ipc` tests should be failing on the new archive, e.g.
```
$ humility -d tests/cmd/cores/hubris.core.u16-ringbuf counters ipc
humility: attached to dump
humility counters failed: read of 4704 bytes from invalid address: 0x24000490
```
Is that expected for a single-task dump?

humility-doppel/src/lib.rs
```diff
@@ -318,10 +318,29 @@ impl humility::reflect::Load for TaskId {
...
 
+#[derive(Clone, Debug)]
+pub struct RingbufCount(pub u32);
+
+impl humility::reflect::Load for RingbufCount {
+    fn from_value(v: &Value) -> Result<Self> {
+        match v.as_base()? {
+            Base::U16(v) => Ok(Self(u32::from(*v))),
+            Base::U32(v) => Ok(Self(*v)),
+            _ => bail!("not a u32 or u16: {v:?}"),
```
hawkw, line 333: Take it or leave it: should we just go ahead and support any integer here, as well? Just in case we start using `u8` or `u64` counters later?[^1]

[^1]: This is probably unlikely, but...be conservative in what you admit, but liberal in what you accept?

humility-doppel/src/lib.rs
```diff
@@ -318,10 +318,29 @@ impl humility::reflect::Load for TaskId {
 pub struct RingbufEntry {
     pub line: u16,
     pub generation: u16,
-    pub count: u32,
+    pub count: RingbufCount,
     pub payload: Value,
 }
 
+#[derive(Clone, Debug)]
+pub struct RingbufCount(pub u32);
```
hawkw, line 326: Note that as of https://github.com/oxidecomputer/hubris/pull/1733, ringbuf counts can also be `()` if de-duplication is disabled. This shouldn't currently occur in the wild *yet*, as we aren't actually disabling deduplication anywhere, but we ought to start handling the case where there are no counts, too. I'm fine with doing that separately if you prefer not to do it in this PR.

hawkw (approved):
I was planning on fixing this this morning — thanks for taking care of it for me! I had a couple minor comments. 

Regarding support for single-task dumps in `counters ipc`, good catch! I'll go ahead and fix that separately?

hawkw:
> Ah, it looks like we load the full task table:
> 
> ```rust
>         let task_table = {
>             let (base, task_count) = hubris.task_table(core)?;
>             let task_t = hubris.lookup_struct_byname("Task")?.clone();
>             humility_doppel::Task::load_tcbs(
>                 hubris,
>                 core,
>                 base,
>                 task_count as usize,
>                 &task_t,
>             )?
>         };
> ```
> 
> which isn't populated for single-task dumps.
> 
> We could make this command more flexible, only loading the task table if available (and using a dummy value for generation otherwise), but that shouldn't block this PR.

#477 fixes this; it depends on this branch in order to update the tests for the dump you've added. So, once we get this merged, we can merge the fix for single-task dumps as well.

humility-doppel/src/lib.rs
mkeeter: Sure, I bumped the internal size up to a `u64` and made it accept any unsigned integer (bdd39928cc7bae70e192085f844f1b7d1b443ab7)

humility-doppel/src/lib.rs
mkeeter: Switched to an `Option<u64>` (also in bdd39928cc7bae70e192085f844f1b7d1b443ab7).  We should get a unit test for this at some point, but this is fine for now IMHO

PR #486, opened 2024-05-13
idol: assume missing args structs are ZST

cbiffle:
We've got a range of versions of the idolatry code generator in the wild where rustc doesn't emit the foo_ARGS structs for arg-less functions, particularly when using encoding other than zerocopy. Humility won't send arg-less IPCs to firmware in this range of versions.

This is the root cause of https://github.com/oxidecomputer/hubris/issues/1777

I'm also patching Idolatry to make it more likely to produce DWARF, but I think it's best for Humility to tolerate such unused zero-size structs not being omitted.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1195,7 +1195,7 @@ dependencies = [
 
 [[package]]
 name = "humility"
-version = "0.11.5"
+version = "0.11.6"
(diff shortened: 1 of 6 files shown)


bcantrill (approved):
Looks good to me

PR #490, opened 2024-05-24
Add diagrammatic `humility net counters` output

mkeeter:
Screenshot 2024-05-24 at 3 06 54 PM

### DIFF
--- a/cmd/net/src/lib.rs
+++ b/cmd/net/src/lib.rs
@@ -60,6 +60,8 @@ enum NetCommand {
     ///
     /// This is a destructive operation that clears counter values, resetting
     /// them to zero.
+    ///
+    /// If no flags are provided, both the table and diagram are shown.
     Counters {
         /// Show a table of raw counter values
         #[clap(short, long)]
--- a/cmd/net/src/lib.rs
+++ b/cmd/net/src/lib.rs
@@ -536,12 +536,17 @@ fn net_counters_diagram(s: &Struct) -> Result<()> {
         v_media_rx[port] = value(v_rx, "media_good");
     }
 
-    let mac =
-        |i: u16| if v_mac_valid { i.to_string() } else { "--".to_string() };
+    let mac = |i: u16| {
+        if v_mac_valid {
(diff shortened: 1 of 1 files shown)


cmd/net/src/lib.rs
```diff
@@ -57,7 +57,17 @@ enum NetCommand {
...
     /// Print the counters
-    Counters,
+    ///
+    /// This is a destructive operation that clears counter values, resetting
+    /// them to zero.
+    Counters {
+        /// Show a table of raw counter values
+        #[clap(short, long)]
+        table: bool,
+        /// Show a diagram of good packet counts in context
```
hawkw, line 67: very nitpicky: do you think the help text should state that both are displayed if neither flags are present?

cmd/net/src/lib.rs
```diff
@@ -473,6 +503,89 @@ fn net_counters(context: &mut ExecutionContext) -> Result<()> {
...
+        ksz_rx[2].to_string().green(),
+        ksz_tx[2].to_string().green(),
+        ksz_tx[1].to_string().green(),
+        v_media_rx[1].to_string().green(),
+        mac(v_mac_tx[1]),
+        ksz_rx[1].to_string().green(),
+        v_media_tx[1].to_string().green(),
+        mac(v_mac_rx[1]),
+        "MEDIA".dimmed(),
+        "MAC".dimmed()
```
hawkw, line 584: nit, take it or leave it: this might be a lot easier to figure out if we used named parameters in the format placeholders? OTTOH the Unicode art would be less well aligned, so I dunno...

hawkw (approved):
overall, this looks good to me! it's a little bit hard to parse the format string that relies on positional ordering, so i might prefer to use named params there, but, take it or leave it.

cmd/net/src/lib.rs
mkeeter: Sure, done in e89180467001c17d22745957982a6fb3068f0fb4

(also, TIL that `-h` and `--help` are different in `clap`)

PR #491, opened 2024-06-03
Add `read_8` backend for `humility sensors`

mkeeter:
It's silly that `humility sensors` requires `udprpc` to iterate over sensors, when the data is _right there_ in RAM.

This PR adds a backend that reads the raw memory and imitates the `Sensor.get` API locally to decode the values.


cmd/sensors/src/lib.rs
```diff
@@ -52,6 +53,12 @@ struct SensorsArgs {
     #[clap(long, short)]
     list: bool,
 
+    /// use the given backend to read sensor data
+    ///
+    /// [default: `hiffy` if present, otherwise `readmem`]
+    #[clap(long, short, value_enum)]
+    backend: Option<Backend>,
```
hawkw, line 60: i'm not sure if i get the default behavior around this: do you mean that if you pass `--backend` without an argument, it uses `hiffy`, and otherwise, it uses RAM reading? or just, it uses RAM-reading by default?

because, we could express the latter by writing
```suggestion
    #[clap(long, short, value_enum, default_value_t=Backend::Readmem)]
    backend: Backend,
```
avoiding the need to make this an `Option`?

cmd/sensors/src/lib.rs
```diff
@@ -302,6 +325,179 @@ fn print(
...
+        hubris: &HubrisArchive,
+        sensors: &[(usize, HubrisSensor)],
+    ) -> Result<Self> {
+        let find_var = |v| {
+            hubris
+                .qualified_variables()
+                .find(|&(n, _v)| n == format!("task_sensor::main::{v}"))
+                .ok_or_else(|| anyhow!("could not find DATA_VALUE"))
+                .map(|(_name, v)| *v)
+        };
```
hawkw, line 364: very small nit: it's a bit confusing to me that there are two different things named `v` here...

cmd/sensors/src/lib.rs
```diff
@@ -302,6 +325,179 @@ fn print(
...
+impl RamSensorReader {
+    fn new(
+        hubris: &HubrisArchive,
+        sensors: &[(usize, HubrisSensor)],
+    ) -> Result<Self> {
+        let find_var = |v| {
+            hubris
+                .qualified_variables()
+                .find(|&(n, _v)| n == format!("task_sensor::main::{v}"))
+                .ok_or_else(|| anyhow!("could not find DATA_VALUE"))
```
hawkw, line 362: this should probably be 
```suggestion
                .ok_or_else(|| anyhow!("could not find {v}"))
```
right? presently, it will _always_ say that it could not find `DATA_VALUE` even if it was actually `LAST_READING` or `NERRORS` that couldn't be found...

cmd/sensors/src/lib.rs
```diff
@@ -302,6 +325,179 @@ fn print(
...
+        f: F,
+    ) -> Result<Vec<T>> {
+        let vs = humility::reflect::load_value(
+            hubris,
+            buf,
+            hubris.lookup_type(var.goff)?,
+            0,
+        )?;
+        let Value::Struct(vs) = vs else {
+            bail!("expected a MaybeUnint struct, not {vs:?}");
```
hawkw, line 394: typo:
```suggestion
            bail!("expected a MaybeUninit struct, not {vs:?}");
```

hawkw (commented):
I had a couple smallish nits, and noticed that there's an error message that's potentially incorrect. Besides that, this looks good to me!

cmd/sensors/src/lib.rs
mkeeter: The comment here was unclear ("if present" meant "if hiffy was present on the target", not "if the `--backend` flag is present in the CLI").

The intention was to normally use hiffy with an automatic fallback to RAM reading (making the default case robust), but you could override it if you want to force one or the other (hence the `Option`).

Let me know if 019575ef3dee0c0b9787019f0ce18bf5beb8648d is more obvious

cmd/sensors/src/lib.rs
mkeeter: Fixed in 019575ef3dee0c0b9787019f0ce18bf5beb8648d

cmd/sensors/src/lib.rs
mkeeter: Good catch, fixed in 019575ef3dee0c0b9787019f0ce18bf5beb8648d

cmd/sensors/src/lib.rs
```diff
@@ -302,6 +327,179 @@ fn print(
...
+        var: HubrisVariable,
+        f: F,
+    ) -> Result<Vec<T>> {
+        let vs = humility::reflect::load_value(
+            hubris,
+            buf,
+            hubris.lookup_type(var.goff)?,
+            0,
+        )?;
+        let Value::Struct(vs) = vs else {
```
cbiffle, line 395: So it seems like a _lot_ of this code could be simplified by describing the structure you want as a `struct` and applying `derive(Load)`. You could write the struct inline here. It'd also (IMO) make it clearer to a reader what things you're depending on.

I did this recently in the host boot reporting code if you want an example: https://github.com/oxidecomputer/humility/pull/492/files#diff-5290acdb0fe5054172bdefac251ae764fa71a2e5a7e938f1b43af61cbd7e1f5fR88

cmd/sensors/src/lib.rs
hawkw: yeah, i think that makes substantially more sense. thanks!

cmd/sensors/src/lib.rs
```diff
@@ -681,12 +677,18 @@ fn sensors(context: &mut ExecutionContext) -> Result<()> {
 
     let mut reader: Box<dyn SensorReader> = match subargs.backend {
         Some(Backend::Hiffy) => {
+            if core.is_dump() {
+                bail!("cannot use hiffy backend on dump");
+            }
```
hawkw, line 682: nit: perhaps the CLI help text should also mention this?

mkeeter: Sure, done in feea70d58273a78d0d167b05ba2c69a0b5a17244

cmd/sensors/src/lib.rs
mkeeter: Done in ff6df5c20fcfb55be866017c89011348c1d189e9 ; it's not quite a LOC increase, because I had to add `doppel::MaybeUninit<T>`, but now that's around for the future.

PR #497, opened 2024-07-01
add support for new IPCC panic data format

bcantrill:


cmd/host/src/lib.rs
```diff
@@ -249,103 +219,62 @@ fn host_last_panic(hubris: &HubrisArchive, core: &mut dyn Core) -> Result<()> {
...
+
+            for (r, v) in chunk {
+                let label: String = format!("{r}").chars().take(3).collect();
+                print!(" {label:3} {v:#18x}");
+            }
+            println!();
         }
     }
 
+    println!("message:   {}", data.message.unwrap_or("<none>".to_string()));
```
hawkw, line 271: very tiny nit that doesn't actually matter in humility: since constructing the `unwrap_or` value is relatively expensive (allocates), maybe prefer lazily constructing it only if `data.message` actually is `None`?
```suggestion
    println!("message:   {}", data.message.unwrap_or_else(|| "<none>".to_string()));
```

README.md
```diff
@@ -1013,38 +1012,46 @@ It is only functional on a Gimlet SP image.
...
+           rax                0x0 rbx 0xfffffcf931453280 rbp 0xfffff78809c74b60
+           r10                0x1 r11 0xfffffffff7c9c510 r12 0xfffffcf931da7210
+           r13 0xfffffcf931453370 r14                0x0 fsb 0xfffffffffc2604e8
+           gsb 0xfffff78809c74a90 es                 0x0 fs                 0x0
+           gs                 0x0 tra                0x6 err                0x0
+           rip 0xfffffffffc029d70 cs                0x30 rfl            0x10246
+           rsp 0xfffff78809c74ad8 ss                0x38
+message:   BAD TRAP: type=6 (#ud Invalid opcode) rp=fffff78809c749e0 addr=fffffcf853a0f940
+stack:
+           die+0x105                                (0xfffffffffbc5fb45)
```
hawkw, line 1039: It's nice that the stack trace lines up with the registers, but I'll admit that there's a part of me that wonders if we should indent this less so that we can fit longer stack frame names in a smallish terminal...on the other hand, that's maybe only worth worrying about when there's enough Rust code in the kernel that there actually are function names in excess of 40 characters... :P

hawkw (approved):
Very straightforward, looks good to me!

cmd/host/src/lib.rs
bcantrill: Fixed

README.md
bcantrill: I left the indent (I do like it not crammed against the left side), but I gave it a bit more room to fill up the 80 columns.

cmd/host/src/lib.rs
```diff
@@ -234,7 +234,9 @@ fn print_panic(d: Vec<u8>) -> Result<()> {
     println!("thread:    {:#x}", data.thread);
 
     if let Some(time) = data.time {
-        let dt: DateTime<Utc> = time.system_time.into();
+        let dt = DateTime::from_timestamp(time.sec as i64, time.nsec)
+            .ok_or_else(|| anyhow!("invalid timestamp"))?;
```
hawkw, line 238: not a big deal, but, in case this ever actually happens, would it maybe be better to just print "invalid timestamp" in the output and keep going so that we can display the rest of the panic data?

currently, if someone is on Windows and tries to read a panic with a timestamp that Windows can't understand, we'll bail out without printing registers, stack, etc. it seems like it might be nicer to just say "hey, we couldn't represent this timestamp as RFC3339 on your system for some reason, but here's the raw seconds/nanoseconds, and here's the rest of your dump"...

bcantrill: Good call; will fix!

Issue #498, opened 2024-07-25
"humility rebootleby" may have missed a memo?


jclulow:
It seems like the structure of **bootleby** archives has changed some time between **v1.1.0** and **v1.2.0**:

```
$ unzip -t staging/dev/gimlet/bootleby-v1.1.0-staging-dev-gimlet-onlydev.zip
Archive:  staging/dev/gimlet/bootleby-v1.1.0-staging-dev-gimlet-onlydev.zip
    testing: bootleby.bin             OK
    testing: cfpa.bin                 OK
    testing: cmpa.bin                 OK
No errors detected in compressed data of staging/dev/gimlet/bootleby-v1.1.0-staging-dev-gimlet-onlydev.zip.
```

```
$ unzip -t staging/dev/gimlet/bootleby-v1.2.0-staging-dev-gimlet.zip
Archive:  staging/dev/gimlet/bootleby-v1.2.0-staging-dev-gimlet.zip
hubris build archive v0
    testing: app.toml                 OK
    testing: elf/kernel               OK
    testing: git-rev                  OK
    testing: build_cfg/board_config.toml   OK
    testing: build_cfg/build_config.toml   OK
    testing: build_cfg/keys.toml      OK
    testing: build_cfg/mfg_cfg.toml   OK
    testing: cfpa.bin                 OK
    testing: cmpa.bin                 OK
    testing: img/final.bin            OK
    testing: img/final.elf            OK
No errors detected in compressed data of staging/dev/gimlet/bootleby-v1.2.0-staging-dev-gimlet.zip.
```

It does not appear that **humility rebootleby** has been updated to handle the new format:

```
humility: attached to 1fc9:0143:0NCTG3HM1MC5J via CMSIS-DAP
humility rebootleby failed: can't find bootleby.bin in bundle
```

This is with...

```
$ humility --version
humility 0.11.9
```

PR #499, opened 2024-07-25
Support newer bootleby format

labbott:
Newer bootleby looks like a hubris archive

### DIFF
--- a/cmd/rebootleby/src/lib.rs
+++ b/cmd/rebootleby/src/lib.rs
@@ -42,10 +42,17 @@ fn rebootleby(context: &mut ExecutionContext) -> Result<()> {
     let mut zip =
         ZipArchive::new(bundle_reader).context("opening bundle file as ZIP")?;
 
-    let img_bootleby = {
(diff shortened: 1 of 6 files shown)


labbott:
Fixes #498

cmd/rebootleby/src/lib.rs
```diff
@@ -42,13 +42,20 @@ fn rebootleby(context: &mut ExecutionContext) -> Result<()> {
     let mut zip =
         ZipArchive::new(bundle_reader).context("opening bundle file as ZIP")?;
 
-    let img_bootleby = {
+    let img_bootleby = if zip.file_names().any(|x| x == "bootleby.bin") {
```
mkeeter, line 45: You could probably do something like this (untested)
```rust
let img_bootleby = {
    let mut entry = zip
        .by_name("bootleby.bin")
        .or_else("img/final.bin")
        .context("can't find bootleby.bin or img/final.bin in bundle")?;
     let mut data = vec![];
    entry.read_to_end(&mut data).context("reading bootleby binary")?;
    data
};
```

PR #510, opened 2024-10-04
Don't require gdb script

labbott:
Having the gdb script in the archive reduces reproducibility as it contains local paths. Having humility run gdb is probably not scriptly necessary anyway so make a best effort and then spit out an error.


cmd/gdb/src/lib.rs
```diff
@@ -75,12 +79,23 @@ fn gdb(context: &mut ExecutionContext) -> Result<()> {
...
+    let gdb_script_path = work_dir.path().join("script.gdb");
+    // We moved the gdb file out of the hubris archive because it was
+    // preventing reproducible builds. An extraction is non fatal
+    let _ = hubris.extract_file_to("debug/script.gdb", &gdb_script_path);
+
+    if let Some(path) = subargs.gdb_script {
+        // It's possible we overwrite the script in the old archive but
+        // that seems like behavior someone would want when actively
+        // passing in a path!
+        std::fs::copy(path, &gdb_script_path)?;
```
mkeeter, line 92: Small suggestion to make things explicit:
```suggestion
    let r = hubris.extract_file_to("debug/script.gdb", &gdb_script_path);

    if let Some(path) = subargs.gdb_script {
        // It's possible we overwrite the script in the old archive but
        // that seems like behavior someone would want when actively
        // passing in a path; print a warning to that effect
        if r.is_okay() {
            warn!("--gdb_script is overriding GDB script in archive");
        }
        std::fs::copy(path, &gdb_script_path)?;
```

mkeeter (approved):

PR #518, opened 2024-10-29
Support `qspi --diffwrite FILE --addr ADDR`

mkeeter:
Previously, `--diffwrite` ignored `--addr` and always wrote from sector 0.

This PR lets us use the `--addr` argument as a starting (sector-aligned) address.

In addition, `qspi --writefile` now bails with an error if `--addr` is provided, because it's _not_ used.

### DIFF
--- a/cmd/qspi/src/lib.rs
+++ b/cmd/qspi/src/lib.rs
@@ -149,6 +149,7 @@ struct QspiArgs {
     /// specify flash address in bytes
     #[clap(long, short, value_name = "address",
         parse(try_from_str = parse_int::parse),
+        conflicts_with("writefile"),
     )]
     addr: Option<usize>,
 
@@ -251,6 +252,7 @@ fn deltas(
     device: &QspiDevice,
     filename: &str,
     compare: &[(u32, Vec<u8>)],
+    base_addr: u32,
     mut diff: impl FnMut(u32, &[u8]) -> Result<()>,
 ) -> Result<()> {
     let filelen = fs::metadata(filename)?.len() as u32;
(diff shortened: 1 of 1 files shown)


nathanaelhuffman (approved):

cmd/qspi/src/lib.rs
```diff
@@ -576,6 +578,9 @@ fn qspi(context: &mut ExecutionContext) -> Result<()> {
         ops.push(Op::Call(qspi_page_program.id));
         (Some(arr), qspi_page_program)
     } else if let Some(filename) = subargs.writefile {
+        if subargs.addr.is_some() {
+            bail!("--addr is not implemented when using --writefile");
```
hawkw, line 582: Nit, take it or leave it: couldn't this alternatively be done by putting `#[arg(..., conflicts_with = "addr")]` on the `writefile` argument?

cmd/qspi/src/lib.rs
```diff
@@ -852,7 +857,13 @@ fn qspi(context: &mut ExecutionContext) -> Result<()> {
         // We are going to hash the contents to find the differences, and
         // then erase/flash the different sectors.
         //
-        let mut address = 0u32;
+        let base_addr = subargs.addr.unwrap_or(0) as u32;
+        if base_addr % SECTOR_SIZE != 0 {
+            bail!(
+                "base address {base_addr:#x} must be divisible by sector size \
+                 ({SECTOR_SIZE})"
+            );
```
hawkw, line 865: super nitpicky, feel free to ignore: would it maybe be helpful to explicitly indicate here that the base address came from the `--addr` argument? since, if it wasn't provided by the CLI, it will be 0, which will always be well aligned, it might be useful-ish to tell the user where the misaligned address came from?

hawkw (approved):
looks good to me! i had a couple teensy CLI UX suggestions, but no blockers.

cmd/qspi/src/lib.rs
mkeeter: Sure, done!

cmd/qspi/src/lib.rs
mkeeter: Done!

PR #520, opened 2024-11-18
Add subcommand to "hydrate" a task dump

mkeeter:
See https://github.com/oxidecomputer/management-gateway-service/pull/316
and https://github.com/oxidecomputer/hubris/pull/1921 for details

This PR adds a new `humility hydrate` subcommand, which takes a "dry" task dump and staples it to a Hubris archive, creating a normal Hubris core file.

### DIFF
--- a/cmd/hydrate/src/lib.rs
+++ b/cmd/hydrate/src/lib.rs
@@ -117,51 +117,29 @@ fn run(context: &mut ExecutionContext) -> Result<()> {
     let mut z = zip::ZipArchive::new(f)?;
 
     let mut s = String::new();
-    z.by_name("meta.json")?.read_to_string(&mut s)?;
+    z.by_name("dump.json")?.read_to_string(&mut s)?;
+
     #[derive(serde::Deserialize)]
-    struct Meta {
-        version: u64,
+    struct DumpInfo {
+        format: u64,
+        task_index: u16,
+        crash_time: u64,
+        board_name: String,
+        git_commit: String,
+        archive_id: String,
+        fw_version: Option<String>,
     }
-    let meta: Meta = serde_json::from_str(&s)?;
-    if meta.version != 1 {
-        bail!(
-            "invalid version in `meta.json`: expected 1, got {}",
-            meta.version
+    let info: DumpInfo = serde_json::from_str(&s)?;
+    if info.format != 1 {
+        humility_log::warn!(
+            "unexpected format in `dump.json`: expected 1, got {}",
+            info.format
         );
     }
 
-    let mut s = String::new();
-    z.by_name("TASK_INDEX")?.read_to_string(&mut s)?;
-    let task_id: u16 = s.trim().parse()?;
-
-    let mut s = String::new();
-    z.by_name("TIMESTAMP")?.read_to_string(&mut s)?;
-    let timestamp: u64 = s.trim().parse()?;
-
-    let mut bord = String::new();
-    z.by_name("BORD")?.read_to_string(&mut bord)?;
-    bord.pop(); // remove newline
-
-    let mut gitc = String::new();
-    z.by_name("GITC")?.read_to_string(&mut gitc)?;
-    gitc.pop(); // remove newline
-
-    let vers = match z.by_name("VERS") {
-        Ok(mut v) => {
-            let mut s = String::new();
-            v.read_to_string(&mut s)?;
-            s.pop();
-            Some(s)
-        }
-        Err(zip::result::ZipError::FileNotFound) => None,
-        Err(e) => return Err(e.into()),
-    };
-
-    let mut s = String::new();
-    z.by_name("ARCHIVE_ID")?.read_to_string(&mut s)?;
     let mut archive_id = [0u8; 8];
     for (i, b) in archive_id.iter_mut().enumerate() {
-        *b = u8::from_str_radix(&s[i * 2..][..2], 16)?;
+        *b = u8::from_str_radix(&info.archive_id[i * 2..][..2], 16)?;
     }
 
     // We have to collect filenames separately, because we can't iterate over
@@ -182,12 +160,12 @@ fn run(context: &mut ExecutionContext) -> Result<()> {
     }
 
     msg!("read dehydrated crash dump");
-    msg!("  task index: {task_id}");
-    msg!("  crash time: {timestamp}");
+    msg!("  task index: {}", info.task_index);
+    msg!("  crash time: {}", info.crash_time);
     msg!("  archive id: {archive_id:02x?}");
-    msg!("  board:      {bord}");
-    msg!("  git commit: {gitc}");
-    msg!("  version:    {}", vers.as_deref().unwrap_or("[missing]"));
+    msg!("  board:      {}", info.board_name);
+    msg!("  git commit: {}", info.git_commit);
+    msg!("  version:    {}", info.fw_version.as_deref().unwrap_or("[missing]"));
     msg!("  {} memory regions:", mem.len());
     for (k, v) in &mem {
         msg!("    {k:#08x}: {} bytes", v.len());
@@ -212,9 +190,9 @@ fn run(context: &mut ExecutionContext) -> Result<()> {
         &mut core,
         Some(humpty::DumpTask {
             magic: humpty::DUMP_TASK_MAGIC,
-            id: task_id,
+            id: info.task_index,
             pad: 0u32,
-            time: timestamp,
+            time: info.crash_time,
         }),
         subargs.out.as_deref(),
         Some(std::time::Instant::now()),
--- a/cmd/hydrate/src/lib.rs
+++ b/cmd/hydrate/src/lib.rs
@@ -10,6 +10,12 @@
 //! ```console
 //! $ humility -a build-grapefruit-image-default.zip hydrate hubris.dry.0
 //! ```
+//!
+//! The archive is required, because the raw memory dump does not contain debug
+//! information.
+//!
+//! By default, this writes to `hubris.core.TASK_NAME.N` (where `N` is the
+//! lowest available value); use `--out` to specify a different path name.
 
 use anyhow::{anyhow, bail, Result};
 use clap::{ArgGroup, IntoApp, Parser};
@@ -26,6 +32,10 @@ use std::{collections::BTreeMap, io::Read, path::PathBuf};
     group = ArgGroup::new("target").multiple(false)
 )]
 struct Args {
+    /// Path to write the resulting dump
+    #[clap(short, long)]
+    out: Option<PathBuf>,
+
     /// Path to the raw memory dump
     file: PathBuf,
 }
@@ -154,6 +164,8 @@ fn run(context: &mut ExecutionContext) -> Result<()> {
         *b = u8::from_str_radix(&s[i * 2..][..2], 16)?;
     }
 
+    // We have to collect filenames separately, because we can't iterate over
+    // `file_names` and read them with `ZipArchive::by_name` simultaneously.
     let mut mem = BTreeMap::new();
     let mem_files = z
         .file_names()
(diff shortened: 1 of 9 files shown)


cmd/hydrate/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+        data.copy_from_slice(&mem[offset..end]);
+        Ok(())
+    }
+}
+
+fn run(context: &mut ExecutionContext) -> Result<()> {
+    let Subcommand::Other(subargs) = context.cli.cmd.as_ref().unwrap();
+    let subargs = Args::try_parse_from(subargs)?;
+    let f = std::fs::File::open(&subargs.file)?;
+    let mut z = zip::ZipArchive::new(f)?;
```
labbott, line 107: How well documented/stable is this format and how mad are we going to be if we need to change it in the future?

mkeeter: I don't expect it to change much, because there's only so many ways to represent a bunch of bytes.  Still, I added `meta.json` containing a `version` value.

cmd/hydrate/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+    // compare archive ID
+    if context.cli.archive.is_none() {
+        bail!("must provide an archive");
+    }
+    let archive = context.archive.as_ref().unwrap();
+    let expected_id = archive
+        .image_id()
+        .ok_or_else(|| anyhow!("missing image ID in archive"))?;
+    if archive_id != expected_id {
+        bail!("image ID mismatch: archive ID is {expected_id:02x?}");
```
hawkw, line 193: should we also include the actual archive ID in this error message?

cmd/hydrate/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! ## `humility hydrate`
+//!
+//! `humility hydrate` combines a raw memory dump (obtained from MGS) and a
+//! Hubris archive, forming a proper Hubris dump.
+//!
+//! ```console
+//! $ humility -a build-grapefruit-image-default.zip hydrate hubris.dry.0
```
hawkw, line 11: this is probably not a big deal, but it might be nice if we actually stated somewhere explicitly that the archive is required for this command?

labbott (approved):

cmd/hydrate/src/lib.rs
hawkw: it occurs to me that if we want to make it a _little_ easier to evolve the format in the future, we could maybe stuff the paths to the various file entries we read out of the zip file in `meta.json`, so the format is a little bit more self-describing? i.e. instead of `z.by_name("TASK_INDEX")`, you would do `z.by_name(meta.task_index_path)` and so on. 

that way, if the paths change, we don't actually need to treat that as a totally new version that has to be handled separately. on the other hand, i'm not sure if changing these paths is actually a thing that will, like, ever reasonably happen, so maybe this isn't worth the effort, even if it's not that much effort.

cmd/hydrate/src/lib.rs
hawkw: also, maybe the docs should explicitly note that the result of this command is that a new core file for the dump is written to the filesystem in the current working directory? i realize that this is the same behavior as other dump commands, but it might be nice to write it down in the docs for this command so that it's obvious what it does...

cmd/hydrate/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+    z.by_name("TIMESTAMP")?.read_to_string(&mut s)?;
+    let timestamp: u64 = s.trim().parse()?;
+
+    let mut bord = String::new();
+    z.by_name("BORD")?.read_to_string(&mut bord)?;
+    bord.pop(); // remove newline
+
+    let mut gitc = String::new();
+    z.by_name("GITC")?.read_to_string(&mut gitc)?;
+    gitc.pop(); // remove newline
```
hawkw, line 137: nit, take it or leave it: perhaps we ought to make this explicitly remove _only_ a newline, rather than "whatever the last character is"? i realize these are always expected to end with a newline, but it might be nice to be forward compatible and handle any potential future changes where they're _not_ newline-terminated, rather than just happily corrupting the commit SHA or board name if the newline ever goes away?

```suggestion
    let bord = bord.trim_end_matches('\n'); // remove newline

    let mut gitc = String::new();
    z.by_name("GITC")?.read_to_string(&mut gitc)?;
    let gitc = gitc.trim_end_matches('\n');; // remove newline
```

cmd/hydrate/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    let vers = match z.by_name("VERS") {
+        Ok(mut v) => {
+            let mut s = String::new();
+            v.read_to_string(&mut s)?;
+            s.pop();
```
hawkw, line 143: similarly, perhaps we ought to `trim_end_matches` here?

cmd/hydrate/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+    let meta: Meta = serde_json::from_str(&s)?;
+    if meta.version != 1 {
+        bail!(
+            "invalid version in `meta.json`: expected 1, got {}",
+            meta.version
+        );
+    }
+
+    let mut s = String::new();
+    z.by_name("TASK_INDEX")?.read_to_string(&mut s)?;
```
hawkw, line 124: nit: looks like the `ZipError::FileNotFound` variant from the `zip` crate doesn't include the filename, so perhaps we ought to do something like:
```suggestion
    z.by_name("TASK_INDEX")
      .context("invalid dump: missing TASK_INDEX")?
      .read_to_string(&mut s)
      .context("failed to read TASK_INDEX from dump")?;
```
or similar, here and for every other time we read a file?

if that's too wordy, we could also have a helper like
```rust
fn read_zip_entry(z: &mut ZipArchive, file: &str) -> anyhow::Result<String> {
    let mut s = String::new();
    z.by_name(file)
      .with_context(|| format!("invalid dump: missing {file}"))?
      .read_to_string(&mut s)
      .with_context(|| format!("failed to read {file} from dump"))?;
    Ok(s)
}
```
and just call that everywhere we read something from the dry dump?

cmd/hydrate/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+    msg!("  {} memory regions:", mem.len());
+    for (k, v) in &mem {
+        msg!("    {k:#08x}: {} bytes", v.len());
+    }
+
+    // compare archive ID
+    if context.cli.archive.is_none() {
+        bail!("must provide an archive");
+    }
+    let archive = context.archive.as_ref().unwrap();
```
hawkw, line 188: nitpicky, feel free to ignore: why not golf these two down to just:
```suggestion
    let Some(archive) = context.archive.as_ref() else {
        bail!("must provide an archive");
    };
```

cmd/hydrate/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+    for (i, b) in archive_id.iter_mut().enumerate() {
+        *b = u8::from_str_radix(&s[i * 2..][..2], 16)?;
+    }
+
+    let mut mem = BTreeMap::new();
+    let mem_files = z
+        .file_names()
+        .filter(|f| f.starts_with("0x"))
+        .map(|s| s.to_owned())
+        .collect::<Vec<_>>();
```
hawkw, line 162: ~~nit, not a big deal: we should be able to drop the `collect` here, and instead just `for` loop over the iterator directly, right?~~

EDIT: oh, I bet this is necessary because the iterator borrows the `ZipArchive`, but calling `ZipArchive::by_name` inside the loop mutably borrows it, so we can't do that. Maybe worth a comment here?

hawkw (approved):
Overall, the command looks good to me! I had a bunch of fairly minor little suggestions, but no major issues.

cmd/hydrate/src/lib.rs
mkeeter: Yup, done

cmd/hydrate/src/lib.rs
mkeeter: Good idea, done

cmd/hydrate/src/lib.rs
mkeeter: That's not _exactly_ equivalent, because `context.cli.archive` and `context.archive` are different things!  However, I'm not sure why they'd ever differ, so this is probably a fine suggestion; done

cmd/hydrate/src/lib.rs
mkeeter: I decided to move the metadata from scattered files in the ZIP archive into the JSON file (renamed to `dump.json`), which lets us deserialize with `serde` (and makes the questions about trailing newlines moot)

cmd/hydrate/src/lib.rs
mkeeter: This is somewhat moot because I moved metadata into the JSON file, but I added a few `with_context(...)` annotations elsewhere.

cmd/hydrate/src/lib.rs
hawkw: Oh! I can't read --- I thought they were both `context.archive`. Whoops!

cmd/hydrate/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+//!
+//! ```console
+//! $ humility -a build-grapefruit-image-default.zip hydrate hubris.dry.0
+//! ```
+//!
+//! The archive is required, because the raw memory dump does not contain debug
+//! information.
+//!
+//! By default, this writes to `hubris.core.TASK_NAME.N` (where `N` is the
+//! lowest available value); use `--out` to specify a different path name.
```
hawkw, line 18: lovely, thank you for adding both of these!

cmd/hydrate/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+#[clap(
+    name = "hydrate", about = env!("CARGO_PKG_DESCRIPTION"),
+    group = ArgGroup::new("target").multiple(false)
+)]
+struct Args {
+    /// Path to write the resulting dump
+    #[clap(short, long)]
+    out: Option<PathBuf>,
+
+    /// Path to the raw memory dump
```
hawkw, line 39: turbo-nit: we _may_ want to add
```suggestion
    #[clap(short, long, value_hint = clap::ValueHint::FilePath)]
    out: Option<PathBuf>,

    /// Path to the raw memory dump
    #[clap(value_hint = clap::ValueHint::FilePath)]
```
to generate nicer completions, but...i don't believe we're currently generating completions for humility anyway, so it may not matter...

cmd/hydrate/src/lib.rs
hawkw: lovely!

cmd/hydrate/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+    #[derive(serde::Deserialize)]
+    struct DumpInfo {
+        format: u64,
+        task_index: u16,
+        crash_time: u64,
+        board_name: String,
+        git_commit: String,
+        archive_id: String,
+        fw_version: Option<String>,
+    }
```
hawkw, line 133: part of me wonders if it might be worth having a shared definition of this structure that could also be used by MGS instead of the [ad-hoc JSON here](https://github.com/oxidecomputer/management-gateway-service/pull/316/files#diff-eecb3109d05f85699dc42276b1a71b8ec4041db954cce5ba3a1496c386685bf0R1583-R1589), but... i don't really know what crate we would put that in that it would make sense to have as a dependency of both MGS and humility. so, maybe it's not worth the effort... :woman_shrugging:

hawkw (approved):
latest changes look great!

cmd/hydrate/src/lib.rs
mkeeter: I think our use of Clap is weird enough that completions would be a challenge, so I'm going to punt on this for now.

cmd/hydrate/src/lib.rs
mkeeter: I think the dependency should probably go Humility → something in `management-gateway`, but agree that it's not worth the effort / coupling yet...

PR #528, opened 2025-01-07
Allow building without usb/debug probes

labbott:
Humility was originally designed to be used with a variety of debug probes. It's since grown the ability to take dumps over the network. There are situations/environments where it's useful to run in network only mode. Allow building without debug probes with `cargo build --no-default-features`.

To make this work with cargo features, the probes are pulled into a separate crate and the humility binary is moved to `humilty-bin`


humility-cmd/Cargo.toml
```diff
@@ -10,3 +10,9 @@ clap.workspace = true
 humility.workspace = true
 humility-net-core.workspace = true
 humility-cli.workspace = true
+
+humility-probes-core = {workspace = true, optional = true }
```
mkeeter, line 14: Nit:
```suggestion
humility-probes-core = { workspace = true, optional = true }
```

mkeeter (approved):

jclulow:
Is there (can there be?) an indication in the `humility --version` output that it's the not-USB version?  I'd like to package this up for Helios in such a way that you can choose whether you get the USB version or not via the use of a [pkg(7) variant](https://man.omnios.org/man7/pkg#FACETS_AND_VARIANTS), and it would be good to be able to tell through the version output which flavour you happen to have installed.

labbott:
> Is there (can there be?) an indication in the `humility --version` output that it's the not-USB version? I'd like to package this up for Helios in such a way that you can choose whether you get the USB version or not via the use of a [pkg(7) variant](https://man.omnios.org/man7/pkg#FACETS_AND_VARIANTS), and it would be good to be able to tell through the version output which flavour you happen to have installed.

would something like the following work?

```
$ ./target/debug/humility --version
humility 0.12.0 probeless
```

PR #529, opened 2025-01-15
Remove hardware tracing support

labbott:
We never really got ITM/ETM working as well as we wanted and keeping this support is making it harder to update probe-rs. Just drop it for now and an eager person can bring it back later.


mkeeter (approved):

bcantrill (approved):
I would say thanks for the memories, but they aren't all that fond to be honest -- so good riddance!

PR #533, opened 2025-02-26
Update `ipcc-data` URL

mkeeter:


labbott (approved):

Cargo.toml
```diff
@@ -95,7 +95,7 @@ tlvc = { git = "https://github.com/oxidecomputer/tlvc" }
 tlvc-text = {git = "https://github.com/oxidecomputer/tlvc"}
 vsc7448-info = { git = "https://github.com/oxidecomputer/vsc7448.git" }
 vsc7448-types = { git = "https://github.com/oxidecomputer/vsc7448.git" }
-ipcc-data = { git = "https://github.com/oxidecomputer/ipcc-data" }
+ipcc-data = { git = "https://github.com/oxidecomputer/ipcc-rs" }
```
labbott, line 98: Do we want to consider pinning this to a specific commit, especially in a shared repo?

mkeeter: It's implicitly pinned, because we commit `Cargo.lock` (unlike `sprockets`), so I think it's fine!

PR #540, opened 2025-04-09
fix "humility pmbus -w OPERATION.OnOffState" on lm5066

bcantrill:


cmd/pmbus/src/lib.rs
```diff
@@ -1312,16 +1324,29 @@ fn writes(
...
 
-        ndx += 1;
-
-        let getmode = || mode;
+        let getmode = || match mode {
+            Some(mode) => mode,
+            None => {
+                panic!("unexpected call to get VOutMode");
+            }
+        };
```
mkeeter, line 1349:
```suggestion
        let getmode = || mode.expect("unexpected call to get VOutMode");
```

mkeeter (approved):
One small suggestion, but this looks fine otherwise

PR #542, opened 2025-04-29
"humility rendmp": add Gen3 support

bcantrill:


cmd/rendmp/src/lib.rs
```diff
@@ -483,10 +550,6 @@ impl RendmpDevice {
     ) -> Result<Vec<Option<RendmpBankStatus>>> {
         let mut rval = vec![];
 
-        if status.len() != 8 {
```
mkeeter, line 486: Should we `assert_eq!(status.len(), self.bank_status_len())` here?

mkeeter (approved):

PR #550, opened 2025-06-17
Remove `bankerase` command

labbott:
This was a very old command that was added for some early LPC55 work that mostly serves to confuse people. Remove it.

Fixes humility#387


mkeeter (approved):

hawkw:
(when i try to actually leave a review, I keep getting an error page from GitHub)

PR #552, opened 2025-07-15
Add special-case handling for measurement token handoff

mkeeter:
(opening this as a draft while we wait for https://github.com/oxidecomputer/lpc55_support/pull/94)

Without this change, the SP's "multiple seconds of resets" in https://github.com/oxidecomputer/hubris/pull/2138 cause Humility to time out.  We could increase Humility's timeout, but having an explicit `SKIP` token seems like a cleaner solution; with the SWD dongle attached, the SP is definitely not getting measured!


cmd/flash/src/lib.rs
```diff
@@ -403,7 +403,22 @@ fn flashcmd(context: &mut ExecutionContext) -> Result<()> {
...
     }
 
-    core.reset()?;
+    // If this image uses handoff to send a measurement token between the RoT
+    // and SP, this won't work with a debugger physically attached.  To prevent
+    // the SP from resetting itself, we write a different token which skips this
+    // reboot loop.  The memory address and token values are hard-coded based on
+    // the `measurement-token` crate in Hubris.
+    if hubris.manifest.features.iter().any(|s| s == "measurement-handoff") {
+        core.reset_and_halt(std::time::Duration::from_millis(25))?;
```
labbott, line 412: Do we want to make this time a constant in `measurement-token` too?

PR #556, opened 2025-08-08
Bump gimli, goblin, scroll, and edition

mkeeter:


humility-core/src/hubris.rs
hawkw, line 1: There are some Clippy warnings in here about hidden elided lifetimes, but they don't seem to be making the check fail: https://github.com/oxidecomputer/humility/actions/runs/16838991454/job/47705335164?pr=556#step:6:733

humility-core/src/hubris.rs
```diff
@@ -3933,11 +3933,8 @@ impl HubrisObjectLoader {
                 continue;
             }
 
-            let name = match elf.strtab.get(sym.st_name) {
-                Some(n) => n?,
-                None => {
-                    bail!("bad symbol in {}: {}", object, sym.st_name);
-                }
+            let Some(name) = elf.strtab.get_at(sym.st_name) else {
+                bail!("bad symbol in {}: {}", object, sym.st_name);
```
hawkw, line 3937: nit, take it or leave it:
```suggestion
                bail!("bad symbol in {object}: {}",, sym.st_name);
```

PR #559, opened 2025-08-12
Fix warnings from new `rustc` (skipping 1.88)

mkeeter:
This is similar to #558 , but skips 1.88 so we don't need to fix every single `format!` string.


labbott (approved):
LGTM, hopefully nothing is depending on us having a `rust-toolchain.toml`

PR #564, opened 2025-08-25
Let `humility map` work with programmer attached

mkeeter:
Previously, `humility map` would work in two configurations:

- If there was no debugger present
- If the debugger was connected to a system **running the same image as the archive**

This is annoying, so I'm fixing it: a command can now specify `CommandKind::Unattached { archive: Archive::Required }` to run with an `ArchiveCore`.


mkeeter:
CI fix: this adds `humility: attached to dump` messages to a bunch of unit tests, which isn't a big deal.

labbott (approved):

PR #570, opened 2025-10-02
Add option to configure debug probe speed

sdonnan:
By default `probe-rs` uses a hardcoded debug adapter speed. For adapters that cannot meet that exact speed they may fall back to a slower one rather than use an internal sane default. This can slow down operation on some adapters (like mculink).

This change adds and optional command line flag to provide a speed setting.

An example of the impact this setting has:

```
~/w/h/t/release (feature/set-probe-speed)> time ./humility flash -F
humility: attached to archive
humility: attaching with chip set to "STM32H753ZITx"
humility: archive appears to be already flashed; forcing re-flash
humility: flashing done

________________________________________________________
Executed in   48.68 secs

~/w/h/t/release (feature/set-probe-speed)> time ./humility --speed 8000 flash -F
humility: attached to archive
humility: attaching with chip set to "STM32H753ZITx"
humility: archive appears to be already flashed; forcing re-flash
humility: flashing done

________________________________________________________
Executed in   34.44 secs
```

Tested with stlinkv3, mculink, and another generic daplink device.

### DIFF
--- a/humility-probes-core/src/lib.rs
+++ b/humility-probes-core/src/lib.rs
@@ -97,7 +97,7 @@ fn open_probe<T: Into<DebugProbeSelector> + Clone>(
         && let Some(rcode) = err.downcast_ref::<rusb::Error>()
         && *rcode == rusb::Error::Busy
     {
-        bail!("USB link in use; is OpenOCD or  another debugger running?");
+        bail!("USB link in use; is OpenOCD or another debugger running?");
     }
 
     let mut probe = res?;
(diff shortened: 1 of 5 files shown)


labbott:
This looks super helpful, thanks for digging into this

humility-probes-core/src/lib.rs
```diff
@@ -63,51 +63,67 @@ fn get_usb_probe(index: Option<usize>) -> Result<DebugProbeInfo> {
...
         }
     }
 
-    res.map_err(|e| e.into())
+    if let Err(DebugProbeError::Usb(Some(ref err))) = res
+        && let Some(rcode) = err.downcast_ref::<rusb::Error>()
+        && *rcode == rusb::Error::Busy
+    {
+        bail!(
+            "USB link in use; is OpenOCD or \
```
mkeeter, line 101: Nit: this could fit on a single line

PR #572, opened 2025-10-02
Remove text output files in `cores` directory

mkeeter:
`o`, `o1`, and `o2` aren't Hubris cores; they look like outputs of `humility readvar`, so they were presumably committed by accident.


hawkw (approved):
o

PR #573, opened 2025-10-08
Ignore archive with rebootleby

labbott:
The bootleby archive looks kind of like a hubris archive but it can't actually be fully loaded with `-a`. This isn't a problem because rebootleby takes the archive separately but it can create confusion if you do attempt to pass it in. Just ignore the archive.


labbott:
This PR sponsored by a brief 15 minute freakout that we had broken `rebootleby`

PR #577, opened 2025-10-20
Use measurement token in auxflash reset as well

mkeeter:
This moves "reset with measurement token handoff" into a new `Core::reset_with_handoff`, which is implemented by default on all `Core` types.  Then, we use this function in both post-flashing and auxflash reset.

I also deleted `EepromHandler::reset` because it was unused.


labbott (approved):
LGTM, takes care of all the places we seem to be calling `reset`

PR #581, opened 2025-10-29
Increment `halted` in `reset_and_halt`

mkeeter:
Otherwise, we can underflow when `run` is called.


mkeeter:
Fixed with https://github.com/oxidecomputer/humility/pull/582 instead.

PR #591, opened 2025-12-10
RAA229620A phase check and open pin support

cbiffle:
This appears to work on Cosmo.

### DIFF
--- a/cmd/rendmp/src/lib.rs
+++ b/cmd/rendmp/src/lib.rs
@@ -216,7 +216,7 @@ struct RendmpArgs {
     #[clap(long, group = "subcommand")]
     open_pin: bool,
 
-    /// checks for open pins on the given converter
+    /// performs a phase check on the converter
     #[clap(long, group = "subcommand")]
     phase_check: bool,
 
@@ -959,6 +959,7 @@ fn rendmp_ingest(subargs: &RendmpArgs) -> Result<()> {
 enum SupportedDevice {
     ISL68224,
     RAA229618,
+    RAA229620A,
 }
 
 impl SupportedDevice {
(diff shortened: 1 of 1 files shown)


cmd/rendmp/src/lib.rs
```diff
@@ -1011,20 +1022,17 @@ fn check_addr(
...
                 && rails.iter().any(|r| r.name == *rail)
             {
-                let dev = match d.device.as_str() {
-                    ISL_DEV_NAME => SupportedDevice::ISL68224,
-                    RAA_DEV_NAME => SupportedDevice::RAA229618,
-                    _ => {
-                        bail!("rail {rail} is not a supported device");
-                    }
+                let Ok(dev) = SupportedDevice::from_str(&d.device) else {
+                    let supported = SupportedDevice::VARIANTS.join(", ");
```
labbott: TIL `EnumVariantNames`

labbott (approved):
LGTM

cmd/rendmp/src/lib.rs
```diff
@@ -1654,19 +1664,17 @@ fn rendmp_phase_check<'a>(
...
-                "cannot check phases with DIMMs present: if a checked \
-                phase powers a DIMM, an I2C hang can result; this can be \
-                overridden with --force-phase-check -- but use this carefully"
-            );
-        }
+    if subargs.force_phase_check {
+        warn!(
+            "DIMM presence check disabled, if this check hangs, remove the \
+            DIMMs and retry."
+        );
```
hawkw: this is confusing to me: this says "DIMM presence check disabled, if this check hangs, remove the DIMMS and retry". after thinking about it for a minute or two, i realized that "DIMM presence check" means the thing that would _skip_ the _phase check_ if DIMMs are present, but it *sounds* like it's saying, "the check i am about to perform has been disabled. if the check i am about to perform, which i just told you was disabled, hangs..." which feels weird.

cbiffle: Hmmmm, yes, I should reword this.

PR #610, opened 2026-04-01
Remove call functionality from `humility rpc`

mkeeter:
(staged on top of #608)

`humility rpc --ip=... --call ...` is _almost_ identical to `humility --ip=... hiffy --call ...`, and duplicates the calling implementation.  The **one** exception is being able to execute a `udprpc` call across every compatible SP.  I don't think we use that functionality anywhere (searching the whole `oxidecomputer` org finds no users of `humility rpc`), so I'd vote to delete it.

This leaves `humility rpc --listen` as the only subcommand, which is still useful (and not duplicated elsewhere).

### DIFF
--- a/cmd/discover/src/lib.rs
+++ b/cmd/discover/src/lib.rs
@@ -2,24 +2,23 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-//! ## `humility rpc`
+//! ## `humility discover`
 //!
-//! `humility rpc` lets you discover SPs on a network, instead of using a
+//! `humility discover` lets you discover SPs on a network, instead of using a
 //! physically attached debugger.  Once SPs are discovered, they may be used as
 //! a target by setting `HUMILITY_IP` or providing the `--ip` argument.
 //!
-//! You may need to configure an IPv6 network for `humility rpc` to work. On
+//! You may need to configure an IPv6 network for `humility discover` to work. On
 //! illumos, it looks like this:
 //!
 //! ```console
 //! $ pfexec ipadm create-addr -t -T addrconf e1000g0/addrconf
 //! ```
 //!
-//! To listen for compatible devices on your network, run `humility rpc
-//! --listen`
+//! To listen for compatible devices on your network, run `humility discover`
 //!
 //! ```console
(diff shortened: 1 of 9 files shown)


cmd/rpc/src/lib.rs
```diff
@@ -4,26 +4,9 @@
...
-//! Function calls are handled identically to the `humility hiffy` subcommand,
-//! except that an `--ip` address is required:
-//!
-//! ```console
-//! $ humility rpc --ip fe80::0c1d:9aff:fe64:b8c2%en0 -c UserLeds.led_on -aindex=0
-//! UserLeds.led_on() = ()
-//! ```
-//!
-//! Alternatively, you can set the `HUMILITY_RPC_IP` environmental variable.
+//! `humility rpc` lets you discover SPs on a network, instead of using a
```
hawkw: So, I kinda feel like, if that's now the only thing it does, maybe it shouldn't be called `humility rpc`...maybe `humility discover` or something? It could remain aliased as `humility rpc --listen` for backwards compatibility if there's scripts that rely on this. It just seems a bit possibly confusing for the help command to be like "hey look! It's `humility rpc`, a command which doesn't actually perform RPCs!"

mkeeter: Seems reasonable, I've renamed to `humility discover` (and added support for running without an archive).

cmd/discover/src/lib.rs
```diff
@@ -0,0 +1,336 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! ## `humility discover`
+//!
+//! `humility discover` lets you discover SPs on a network, instead of using a
+//! physically attached debugger.  Once SPs are discovered, they may be used as
+//! a target by setting `HUMILITY_IP` or providing the `--ip` argument.
```
labbott, line 9: Ohhhhh I feel like I've wanted this exact command before

labbott (approved):

cmd/discover/src/lib.rs
mkeeter: Surprise, it has existed all along!

mkeeter:
Whoops, merged this in the wrong order.

Issue #614, opened 2026-04-13
hiffy sync changes lead to false negatives


mkeeter:
https://github.com/oxidecomputer/humility/pull/598 changed the requirement for `hiffy` to be considered up: we must see `HIFFY_READY == 1` **and** have spotted the task in a syscall.

https://github.com/oxidecomputer/humility/blob/c556213e82a2590166adc0f79b3532e28e7a6456/humility-hiffy/src/lib.rs#L1284-L1318

The goal of that PR was to fix cases where `hiffy` had not yet started, but `HIFFY_READY` was 1 (either spuriously or due to previous RAM contents).  This could occur on images which run high-priority CPU-heavy workloads at startup, since those tasks may starve `hiffy` of runtime.  Waiting to see `hiffy` in a syscall gives us confidence that the task has actually booted; otherwise, Humility prints `HIF execution facility unavailable` and returns an error.

Unfortunately, we're now seeing failures at manufacturing time due to this change ([1](https://mfg-support.shared.oxide.computer/floor/activity/01KP3ET9HZ0ZNTWXV1VPXED70K), [2](https://mfg-support.shared.oxide.computer/floor/activity/01KP3KPZNT9TM8KNHY7VSTBKA4)).

[1] looks like a false positive due to this new check:

- The manufacturing software calls `SpRot.status()` repeatedly and waits for it to return successfully.  The mfg software ignores the `HIF execution facility unavailable` error at this point, because it's waiting for `hiffy` to start up.
- The SP is doing CPU-intensive work, but _occasionally_ lets `hiffy` run, so this call returns successfully — but the SP is still starving `hiffy`.  At this point, mfg software expects `hiffy` calls to succeed.
- Mfg software tries to do another `hiffy` call.  The CPU-intensive work is still running, so this call fails.  The mfg software _is not expecting this_, and bails out with an error.

In [2], it's more ambiguous who's at fault: the manufacturing software doesn't explicitly wait for `hiffy` to come up, so it could be a genuine case of other tasks preventing `hiffy` from starting up – or another false positive.

cc @Aaron-Hartwig @cbiffle

mkeeter:
I can trivially reproduce this on a Grapefruit by starting a host-flash hash running, then trying to call hiffy functions.  The system sees `hiffy` as runnable, rather than in a syscall:
```
humility: attached via ST-Link V3
system time = 355329
ID TASK                       GEN PRI STATE
[...]

PR #615, opened 2026-04-13
More laps when waiting for hiffy

mkeeter:
This is a (stochastic) fix for https://github.com/oxidecomputer/humility/issues/614

It gives us ~250 chances to catch `hiffy` in a syscall over 5 seconds (using our typical timeout), rather than 10 chances over ~50 ms (using a hard-coded loop count).  Testing on my desk, I can longer trigger the `HIF execution facility unavailable` failure by making `hiffy` calls while hashing the host flash (to load the SP CPU); previously, it happened about 50% of the time.

In the worst case, this could cause `humility hiffy` calls to take twice as long (if we almost hit the timeout waiting for the `hiffy` task, then almost hit it waiting for the operation to complete) – but I don't think that's a big issue here.

### DIFF
--- a/humility-hiffy/src/lib.rs
+++ b/humility-hiffy/src/lib.rs
@@ -1025,28 +1025,29 @@ impl<'a> HiffyContext<'a> {
         }
 
         let mut lap = 0;
-        const MAX_LAPS: u64 = 10;
 
-        let ready = loop {
+        let start = std::time::Instant::now();
+        let mut ready = false;
+        while start.elapsed().as_millis() < u128::from(self.timeout) {
             if !syscall_observed {
                 if has_task_started(self.hubris, core, hiffy_task.unwrap())? {
                     syscall_observed = true;
                 }
             } else if core.read_word_32(self.ready.addr)? == 1 {
-                break true;
+                ready = true;
+                break;
             }
 
             core.op_done()?;
 
             lap += 1;
 
-            if lap >= MAX_LAPS {
-                break false;
-            }
+            // sleep time is based on vibes
+            let sleep_time = Duration::from_millis(lap.min(20));
+            thread::sleep(sleep_time);
 
-            thread::sleep(Duration::from_millis(lap));
             core.op_start()?;
-        };
+        }
 
         if !ready {
             bail!("HIF execution facility unavailable");
--- a/humility-hiffy/src/lib.rs
+++ b/humility-hiffy/src/lib.rs
@@ -1028,7 +1028,8 @@ impl<'a> HiffyContext<'a> {
 
         let start = std::time::Instant::now();
         let mut ready = false;
-        while start.elapsed().as_millis() < u128::from(self.timeout) {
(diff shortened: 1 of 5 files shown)


Aaron-Hartwig (commented):
I'll leave the actual code approval here to @cbiffle, but I think this is reasonable.

>In the worst case, this could cause humility hiffy calls to take twice as long (if we almost hit the timeout waiting for the hiffy task, then almost hit it waiting for the operation to complete) – but I don't think that's a big issue here.

I agree that this is not a big issue.

Aaron-Hartwig (approved):

humility-hiffy/src/lib.rs
```diff
@@ -1025,28 +1025,29 @@ impl<'a> HiffyContext<'a> {
         }
 
         let mut lap = 0;
-        const MAX_LAPS: u64 = 10;
 
-        let ready = loop {
+        let start = std::time::Instant::now();
+        let mut ready = false;
+        while start.elapsed().as_millis() < u128::from(self.timeout) {
```
jamesmunns: nit:

```suggestion
        let timeout = Duration::from_millis(self.timeout.into();
        while start.elapsed() < timeout {
```

humility-hiffy/src/lib.rs
```diff
@@ -1025,28 +1025,29 @@ impl<'a> HiffyContext<'a> {
...
 
             core.op_done()?;
 
             lap += 1;
 
-            if lap >= MAX_LAPS {
-                break false;
-            }
+            // sleep time is based on vibes
+            let sleep_time = Duration::from_millis(lap.min(20));
```
jamesmunns, line 1047: Just checking, you want this to CAP at "every 20ms" as the longest interval? I always screw up "min == AT MOST", not doing this backoff-style?

humility-hiffy/src/lib.rs
jamesmunns, line 1057: Nit: Maybe consider some kind of logging on success so we can tell how long this *typically* takes?

jamesmunns (approved):
Overall LGTM, this is a bit of a rubber stamp. Some nits/"just checking" comments, take them or leave them.

humility-hiffy/src/lib.rs
jamesmunns: (It *might* be worth renaming that field to `timeout_ms` or such, to indicate unit, or just store a `Duration` directly, but that's even more of a nit).

humility-hiffy/src/lib.rs
mkeeter: Yes, my intent is delay more and more over the first 20 laps (from 0 → 20 ms), then cap the delay at 20 ms (so we get a bunch of samples over the 5-second default timeout).

humility-hiffy/src/lib.rs
mkeeter: I think that would be useful once we switch over to the `log` crate, so we could put it at `debug` or `trace` level, but I don't want to spam the console right now.

PR #616, opened 2026-04-14
Add support for the `hiffy` network backend

mkeeter:
This is the Humility side of https://github.com/oxidecomputer/hubris/pull/2466

It adds a third backend to execute `hiffy` programs over the network; the three backends are now

- Debugger
  - Reads and writes to RAM directly
  - Bytecode is executed by `task-hiffy` on the SP
- Network-attached `udprpc`
  - Bytecode is executed on the host
  - Function calls in bytecode are handled by `udprpc` over the network
- Network-attached `hiffy` (new!)
  - Writes to `hiffy` arrays are done over the network using a new socket
  - Reads are done using existing dump infrastructure to read from RAM
  - Bytecode is executed by `task-hiffy` on the SP

We abstract over the debugger and network-attached `hiffy` with a new `enum HiffyWrite` to write to `hiffy` arrays, instead of calling `HubrisCore` functions directly.  Network-attached `udprpc` remains a special case, because it can only handle a subset of HIF programs (ones that only use `Send`-flavored functions).

### DIFF
--- a/humility-core/src/core.rs
+++ b/humility-core/src/core.rs
@@ -126,6 +126,7 @@ pub trait Core {
 pub enum NetAgent {
     UdpRpc,
     DumpAgent,
+    Hiffy,
 }
 
 pub fn attach_dump(
--- a/humility-doppel/src/lib.rs
+++ b/humility-doppel/src/lib.rs
@@ -38,7 +38,7 @@ use humility::reflect::{self, Base, Load, Ptr, Value};
 use indexmap::IndexMap;
 use std::convert::TryInto;
 use std::fmt;
-use zerocopy::{AsBytes, LittleEndian, U16, U64};
+use zerocopy::{AsBytes, LittleEndian, U16, U32, U64};
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Load)]
 pub struct TaskDesc {
@@ -428,6 +428,20 @@ pub struct RpcHeader {
     pub op: U16<LittleEndian>,
     pub nreply: U16<LittleEndian>,
     pub nbytes: U16<LittleEndian>,
+}
+
+/// Double of the RPC types from `hiffy` (with the `net` feature enabled)
+pub mod hiffy {
+    use super::*;
+
+    #[derive(Copy, Clone, Debug, AsBytes)]
+    #[repr(C)]
+    pub struct RpcHeader {
+        pub image_id: U64<LittleEndian>,
+        pub version: U16<LittleEndian>,
+        pub operation: U16<LittleEndian>,
+        pub arg: U32<LittleEndian>,
+    }
 }
 
 impl humility::reflect::Load for CountedRingbuf {
--- a/humility-hiffy/src/lib.rs
+++ b/humility-hiffy/src/lib.rs
@@ -197,6 +197,11 @@ impl HiffyFunctions {
     }
 }
 
+/// Error indicating that the hiffy backend does not support data arguments
+#[derive(thiserror::Error, Debug)]
+#[error("hiffy backend does not support data arguments")]
+pub struct DataNotSupported;
+
 impl<'a> HiffyContext<'a> {
     fn variable(
         hubris: &'a HubrisArchive,
@@ -450,12 +455,12 @@ impl<'a> HiffyContext<'a> {
     }
 
     /// Returns the size of the `HIFFY_DATA` array, or 0 if unsupported
-    pub fn data_size(&self) -> usize {
+    pub fn data_size(&self) -> Result<usize, DataNotSupported> {
         match &self.hiffy {
             HiffyImpl::Debugger(vars) | HiffyImpl::NetHiffy { vars, .. } => {
-                vars.data.size
+                Ok(vars.data.size)
             }
-            HiffyImpl::NetUdpRpc { .. } => 0, // not supported
+            HiffyImpl::NetUdpRpc { .. } => Err(DataNotSupported),
         }
     }
 
@@ -1049,13 +1054,14 @@ impl<'a> HiffyContext<'a> {
             HiffyImpl::Debugger(vars) => (vars, HiffyWrite::Debugger(vars)),
         };
 
+        let data_size = vars.data.size;
         if let Some(data) = data
-            && data.len() > self.data_size()
+            && data.len() > data_size
         {
             bail!(
                 "data size ({}) exceeds maximum data size ({})",
                 data.len(),
-                self.data_size()
+                data_size,
             );
         }
 
--- a/humility-hiffy/src/lib.rs
+++ b/humility-hiffy/src/lib.rs
@@ -281,7 +281,7 @@ impl<'a> HiffyContext<'a> {
                     vars: HiffyVars::new(hubris)?,
                     ops: HiffyNetOps {
                         write_text: get_tag("WriteHiffyText")?,
-                        write_data: get_tag("HiffyWriteData")?,
+                        write_data: get_tag("WriteHiffyData")?,
                         kick: get_tag("HiffyKick")?,
                     },
                     errs,
--- a/humility-hiffy/src/lib.rs
+++ b/humility-hiffy/src/lib.rs
@@ -44,9 +44,9 @@ pub struct HiffyContext<'a> {
 // Constants when running with the `NetUdpRpc` impl, which runs the program on
 // the host computer.  These values are much larger than anything we'd see on
 // the embedded system, so we can run any program.
-const HIFFY_NET_TEXT_SIZE: usize = 65536;
-const HIFFY_NET_RSTACK_SIZE: usize = 65536;
-const HIFFY_NET_SCRATCH_SIZE: usize = 65536;
+const HIFFY_HOST_NET_TEXT_SIZE: usize = 65536;
+const HIFFY_HOST_NET_RSTACK_SIZE: usize = 65536;
+const HIFFY_HOST_NET_SCRATCH_SIZE: usize = 65536;
 
 /// Variables used when interacting with the `hiffy` task
 struct HiffyVars<'a> {
@@ -341,14 +341,13 @@ impl<'a> HiffyContext<'a> {
         }
 
         let scratch_size = if matches!(hiffy, HiffyImpl::NetUdpRpc { .. }) {
-            HIFFY_NET_SCRATCH_SIZE
+            HIFFY_HOST_NET_SCRATCH_SIZE
         } else {
             // Get the size of the HIFFY_SCRATCH variable, falling back to 256
             // bytes for older images (which used a fixed-size stack array)
             Self::variable(hubris, "HIFFY_SCRATCH", false)
                 .map(|scratch| -> Result<usize> {
-                    let mut buf: Vec<u8> = vec![];
-                    buf.resize_with(scratch.size, Default::default);
+                    let mut buf = vec![0u8; scratch.size];
 
                     core.op_start()?;
                     core.read_8(scratch.addr, buf.as_mut_slice())?;
@@ -449,6 +448,7 @@ impl<'a> HiffyContext<'a> {
         })
     }
 
+    /// Returns the size of the `HIFFY_DATA` array, or 0 if unsupported
     pub fn data_size(&self) -> usize {
         match &self.hiffy {
             HiffyImpl::Debugger(vars) | HiffyImpl::NetHiffy { vars, .. } => {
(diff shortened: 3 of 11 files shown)


humility-hiffy/src/lib.rs
```diff
@@ -205,7 +249,57 @@ impl<'a> HiffyContext<'a> {
...
+                        .cloned()
+                };
+                HiffyImpl::NetHiffy {
+                    vars: HiffyVars::new(hubris)?,
+                    ops: HiffyNetOps {
+                        write_text: get_tag("WriteHiffyText")?,
+                        write_data: get_tag("HiffyWriteData")?,
+                        kick: get_tag("HiffyKick")?,
+                    },
+                    errs,
```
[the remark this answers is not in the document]
mkeeter: nah

humility-hiffy/src/lib.rs
```diff
@@ -939,28 +1019,43 @@ impl<'a> HiffyContext<'a> {
...
+            }
+            HiffyImpl::NetHiffy { vars, ops, errs } => {
+                let image_id = u64::from_le_bytes(
+                    self.hubris.image_id().unwrap().try_into().unwrap(),
                 );
+                let buf_size = self
+                    .hubris
+                    .manifest
+                    .get_socket_by_task("hiffy")
+                    .unwrap()
```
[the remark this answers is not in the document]
mkeeter: nah, we already know there's a socket at this point, but I'll change it to `expect(..)`

humility-hiffy/src/lib.rs
```diff
@@ -205,7 +249,57 @@ impl<'a> HiffyContext<'a> {
         core: &mut dyn Core,
         timeout: u32,
     ) -> Result<HiffyContext<'a>> {
-        if !core.is_net() {
+        let hiffy = if core.is_net() {
+            if let Some(hiffy_task) = hubris.lookup_task("hiffy")
+                && hubris.does_task_have_feature(hiffy_task, "net").unwrap()
```
lzrd: Does/Can this error case give useful guidance to the user? e.g. "hiffy_task is not configured with net feature".
(Looking for other unwraps() that cold be more helpful in common failure cases...)

humility-hiffy/src/lib.rs
```diff
@@ -939,28 +1019,43 @@ impl<'a> HiffyContext<'a> {
...
+                return self.perform_rpc(core, ops);
+            }
+            HiffyImpl::NetHiffy { vars, ops, errs } => {
+                let image_id = u64::from_le_bytes(
+                    self.hubris.image_id().unwrap().try_into().unwrap(),
                 );
+                let buf_size = self
+                    .hubris
+                    .manifest
+                    .get_socket_by_task("hiffy")
```
lzrd, line 1042: nit: (symbolic constants vs. literals) `HIFFY_TASK_NAME`? I know we're not going to change this and the function is `get_socket_by_task()`, but if there are enough "hiffy" names and symbols in different contexts it helps to know what kind of hiffy we are talking about.

lzrd (commented):
Having a more helpful error message for archives where `hiffy` `net` feature is not configured is my only suggestion.

humility-hiffy/src/lib.rs
mkeeter: That `unwrap` can't fail once we've found `hiffy_task`.  I've rewritten this to be more clear.

humility-hiffy/src/lib.rs
mkeeter: I'm ambivalent – `HIFFY_TASK_NAME` is marginally more clear, but replacing `"hiffy"` with `HIFFY_TASK_NAME` everywhere in this file just makes a bunch of lines slightly longer (forcing one large block to reflow to the next indent level).

labbott (commented):
I think this looks okay, I'm going to give it another pass in a bit

humility-core/src/core.rs
```diff
@@ -126,6 +126,7 @@ pub trait Core {
 pub enum NetAgent {
```
jamesmunns, line 126: Should we make this `non_exhaustive` to make additions like this not a breaking change (after this one)? Do we care about breaking changes?

humility-doppel/src/lib.rs
```diff
@@ -429,6 +429,20 @@ pub struct RpcHeader {
     pub nbytes: U16<LittleEndian>,
 }
 
+/// Double of the RPC types from `hiffy` (with the `net` feature enabled)
```
jamesmunns, line 432: Is it worth it to extract this into a shared crate somehow to avoid the risk of unsync'd changes?

humility-hiffy/src/lib.rs
```diff
@@ -34,6 +34,22 @@ enum State {
...
+    timeout: u32,
+    state: State,
+    functions: HiffyFunctions,
+    hiffy: HiffyImpl<'a>,
+}
+
+// Constants when running with the `NetUdpRpc` impl, which runs the program on
+// the host computer.  These values are much larger than anything we'd see on
+// the embedded system, so we can run any program.
+const HIFFY_NET_TEXT_SIZE: usize = 65536;
```
jamesmunns: maybe note the "HOST" part in the name here somewhere to make it clear we're doing so/don't misuse these consts?

humility-hiffy/src/lib.rs
```diff
@@ -344,59 +441,47 @@ impl<'a> HiffyContext<'a> {
...
         })
     }
 
     pub fn data_size(&self) -> usize {
-        self.data.size
+        match &self.hiffy {
+            HiffyImpl::Debugger(vars) | HiffyImpl::NetHiffy { vars, .. } => {
+                vars.data.size
+            }
+            HiffyImpl::NetUdpRpc { .. } => 0, // not supported
```
jamesmunns: Maybe add a function doc comment to note that `0` is a sentinel for "not supported" (or return an `Option` if more invasive changes are palatable?)

humility-hiffy/src/lib.rs
```diff
@@ -205,7 +249,60 @@ impl<'a> HiffyContext<'a> {
         core: &mut dyn Core,
         timeout: u32,
     ) -> Result<HiffyContext<'a>> {
-        if !core.is_net() {
+        let hiffy = if core.is_net() {
```
jamesmunns, line 257: nit: Maybe worth breaking out into a helper function to simplify the control flow/reduce rightward drift (mostly to allow early returns)?

humility-hiffy/src/lib.rs
```diff
@@ -243,29 +340,29 @@ impl<'a> HiffyContext<'a> {
...
-                256
-            }
+        let scratch_size = if matches!(hiffy, HiffyImpl::NetUdpRpc { .. }) {
+            HIFFY_NET_SCRATCH_SIZE
+        } else {
+            // Get the size of the HIFFY_SCRATCH variable, falling back to 256
+            // bytes for older images (which used a fixed-size stack array)
+            Self::variable(hubris, "HIFFY_SCRATCH", false)
+                .map(|scratch| -> Result<usize> {
+                    let mut buf: Vec<u8> = vec![];
```
jamesmunns: Nit: I think you can just do `let mut buf = vec![0u8; scratch.size];`?

edit: I just realized this is old code, feel free to ignore :)

humility-hiffy/src/lib.rs
```diff
@@ -1574,3 +1683,108 @@ fn check_leases(
...
+                    Var::Data => ops.write_data,
+                };
+                // Chosen to be smaller than packet size
+                let Some(chunk_size) = buf_size
+                    .checked_sub(std::mem::size_of::<hiffy::RpcHeader>())
+                else {
+                    bail!("buffer size {buf_size} is smaller than `RpcHeader`");
+                };
+                for (i, chunk) in data.chunks(chunk_size).enumerate() {
+                    let offset = u32::try_from(i * chunk_size).unwrap();
```
jamesmunns: Return "out of bounds" Err instead of unwrap?

humility-hiffy/src/lib.rs
```diff
@@ -1574,3 +1683,108 @@ fn check_leases(
     }
     Ok(())
 }
+
+/// Abstraction for writing hiffy values over multiple transports
+enum HiffyWrite<'a, 'b> {
```
jamesmunns, line 1695: Nit: It *might* be worth making each of the variants a specific struct, e.g. `Debugger(DebuggerWrite)`, to be able to implement methods on each subtype. That would make the top-level dispatch functions a little smaller.

jamesmunns (approved):
Mostly pedantic notes below, I didn't spot anything disagreeable!

One "bigger picture" note here: A lot of the code is somewhat verbose because we're matching on different `HiffyImpl`s, or similar cases. If this was more abstract, I'd say it'd be worth trait-ify-ing, but since we're deciding at runtime, that doesn't make sense unless we want to go down the `dyn Trait` route, which again doesn't make sense because all the impls live in this crate anyway.

HOWEVER, it might be worth *informally* defining an impl API, so we can have top level dispatch functions like:

```rust
pub fn text_size(&self) -> usize {
  match &self.hiffy {
    HiffyImpl::Debugger(dbg) => dbg.text_size(),
    HiffyImpl::NetHiffy(net) => net.text_size(),
    HiffyImpl::NetUdpRpc(..) => HIFFY_NET_TEXT_SIZE,
  }
}
```

then have:

```rust
impl HiffyDebuggerImpl {
  fn text_size(&self) -> usize {
    self.vars.text.size
  }
}
```

This way, we can more easily extract the "backend specific" behavior, and make the high level flow more clear, as well as make some of the helper "backend" functions a little easier to refactor.

If NetUdpRpc is the Last Ever™️ impl, it might not be worth it to refactor, but I think making the "backend contract" even informally defined probably has some documentation value.

humility-core/src/core.rs
mkeeter: I don't think so – this is only ever used as an input, so it would be weird for outside the net core to match on it.  

In general, we (Hubris / Humility) don't care _much_ about breaking changes to quasi-internal Rust APIs, where no one is consuming them besides us and blast radius is a single repository (no one is using `NetAgent` outside of the `humility` repo).  This could change as Humility gets library-ified, but even then I suspect we'll err on the side of breaking things freely.

humility-doppel/src/lib.rs
mkeeter: In theory, that would be the most robust option – but in practice, we've got a few of these internal message types and they're primitive enough to never change.  There's also the fact that if they _did_ change, we'd want to support both versions in Humility, because Humility wants to talk to every Hubris image.

humility-hiffy/src/lib.rs
mkeeter: I'll add a doc comment; 0 is both a sentinel for "not support" and _technically_ a valid response (you can access 0 bytes of `HIFFY_DATA`), so I don't want to return an `Option` here.

jamesmunns: I'm okay with that either way, but wouldn't that be BETTER to use a Option then? To disambiguate between "does not support" and "does support and you can access zero bytes"?

Not too fussed about it though :)

humility-hiffy/src/lib.rs
mkeeter: Will do!

humility-hiffy/src/lib.rs
mkeeter: Fixed, but I also opened https://github.com/oxidecomputer/humility/issues/624 because this isn't strictly necessary

humility-hiffy/src/lib.rs
mkeeter: Fixed!

humility-hiffy/src/lib.rs
mkeeter: I was going to argue that there's not a user-visible difference between "data not supported" and "supports data of size 0", but then I managed to panic `humility auxflash write`:

https://github.com/oxidecomputer/humility/blob/1e40779957d94e57233521ff06733c4ade128492/cmd/auxflash/src/lib.rs#L330

I've now switched `data_size` to return a `Result<usize, DataNotSupported>` (slowly sneaking `thiserror`-based types into Humility as I updated things).

mkeeter:
@jamesmunns `NetHiffy` is going to be the last backend for the forseeable future, so I'm going to punt on more substantial refactorings (like making `enum` variants into their own types).  We can always revisit if someone comes up with a new backend, but I'm not sure what that would be!

Issue #595, opened 2026-02-05
`spd` command is currently very Gimlet-specific


cbiffle:
Currently, the `spd` command relies on the existence of a `gimlet_bufs` field inside Packrat, which I found surprising. This doesn't work on Cosmo.

Now, granted, the entire `spd` command won't work on Cosmo _per se,_ but this is relevant because `rendmp` uses `spd` internally, to check for DIMM presence. This is a Gimlet-ism we're going to have to clean up.

mkeeter:
Done in #617

Issue #568, opened 2025-09-22
`humility spd` may need updates to work with servers other than Gimlet


cbiffle:

mkeeter:
Done in #617

PR #617, opened 2026-04-15
Fix SPD on Gimlet and Cosmo

mkeeter:
This is a fix for https://github.com/oxidecomputer/humility/issues/596, https://github.com/oxidecomputer/humility/issues/595, and https://github.com/oxidecomputer/humility/issues/568.

In https://github.com/oxidecomputer/hubris/pull/2094, we moved the SPD data to a parameterized `struct SpdData<const DIMM_COUNT: usize, const DATA_SIZE: usize>`.  This is because Cosmo has a different number of DIMMS (16 → 12), and their SPD data are a different size (512 → 1024 bytes).

Unfortunately, this broke Humility, which was expecting an array named `spd_data` and instead found a `struct`.

Cosmo is _also_ broken because DDR5 has a different SPD layout than DDR4; even if we could read the `struct`, we wouldn't know how to parse it.

This PR fixes both things, flexibly decoding the `SpdData` and printing both DDR4 and DDR5 SPDs.  (This required updating [the `spd` crate](https://github.com/oxidecomputer/spd) with DDR5 info) 

To avoid a giant diff, this doesn't add new dumps for the test suite, but I'm planning to add them in a subsequent PR.

Here's an example of Cosmo output:
```
     Running `target/debug/humility -d /Users/mjk/Desktop/hubris.core.cosmo spd`
humility: attached to dump
ADDR MANUFACTURER              PART                           WEEK YEAR
   0 Micron Technology         MTC40F2047S1RC64BB1 USFF          2 2025
   1 Micron Technology         MTC40F2047S1RC64BB1 USFF          2 2025
   2 Micron Technology         MTC40F2047S1RC64BB1 USFF          2 2025
   3 Micron Technology         MTC40F2047S1RC64BB1 USFF          2 2025
   4 Micron Technology         MTC40F2047S1RC64BB1 USFF          2 2025
   5 Micron Technology         MTC40F2047S1RC64BB1 USFF          2 2025
   6 Micron Technology         MTC40F2047S1RC64BB1 USFF          2 2025
   7 Micron Technology         MTC40F2047S1RC64BB1 USFF          2 2025
   8 Micron Technology         MTC40F2047S1RC64BB1 USFF          2 2025
   9 Micron Technology         MTC40F2047S1RC64BB1 USFF          2 2025
  10 Micron Technology         MTC40F2047S1RC64BB1 USFF          2 2025
  11 Micron Technology         MTC40F2047S1RC64BB1 USFF          2 2025
```

### DIFF
--- a/cmd/spd/src/lib.rs
+++ b/cmd/spd/src/lib.rs
@@ -135,7 +135,7 @@ struct SpdArgs {
     )]
     timeout: u32,
 
-    /// verbose output (including raw 512-byte SPD)
+    /// verbose output (including raw SPD data)
     #[clap(long, short)]
     verbose: bool,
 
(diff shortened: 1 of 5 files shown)


cmd/spd/src/lib.rs
```diff
@@ -423,14 +523,14 @@ fn spd(context: &mut ExecutionContext) -> Result<()> {
     // First, we want to have all SPDs on the specified bus flip to
     // their 0 page
     //
-    set_page(&mut ops, &i2c_write, 0);
+    set_page(&mut ops, &i2c_write, spd::ee1004::Page::Page0);
```
rmustacc, line 543: What does it mean to be using an `ee1004` constant when the type of memory will vary here and `ee1004` is only really specific to DDR4. The value is the same, but it's a bit confusing to me at first glance. The same for a constant down below.

cmd/spd/src/lib.rs
```diff
@@ -447,14 +547,15 @@ fn spd(context: &mut ExecutionContext) -> Result<()> {
...
+    for addr in 0..spd::ee1004::MAX_DEVICES {
         if results[addr as usize + 1].is_ok() {
             let mut ops = base.clone();
 
             //
             // Issue the read for the bottom 128 bytes from the 0 page
             //
-            let dev = spd::Function::Memory(addr).to_device_code().unwrap();
+            let dev =
+                spd::ee1004::Function::Memory(addr).to_device_code().unwrap();
```
rmustacc, line 575: This is probably a me thing again, but my read is that this is supposed to cover both DDR4 and DDR5 and the function naming while it may overlap is pretty different, especially as we nominally have to talk to the hub and change the page in a different way. Is this section still just DDR4 specific in a way that I'm not realizing?

rmustacc (commented):
You'll want to have some folks look a bit closer at some of the specifics. I just have a few comments that mostly stem from my own confusion I think.

cmd/spd/src/lib.rs
mkeeter: This section is only mean to cover DDR4, since we can only talk directly to the SPDs on Gimlet.  I'll move it into a separate block to make it more obvious.

cmd/spd/src/lib.rs
mkeeter: Same as below; this is now in a function named `dump_ddr4_over_i2c` and warns you if you run it on non-Gimlet hardware.

cmd/spd/src/lib.rs
rmustacc: Thanks, that clears up some of the confusion I had as it looked like we were going to try regardless originally.

cmd/spd/src/lib.rs
```diff
@@ -287,20 +341,10 @@ fn set_page(ops: &mut Vec<Op>, i2c_write: &HiffyFunction, page: u8) {
...
-#[derive(Load, Debug)]
-struct PackratStaticBufs {
-    gimlet_bufs: GimletStaticBufs,
-}
-
 fn spd_lookup(
     hubris: &HubrisArchive,
     core: &mut dyn humility::core::Core,
-) -> Result<Option<Vec<u8>>> {
+) -> Result<Option<Vec<Vec<u8>>>> {
```
labbott: Channelling jgallagher a bit, this type is confusing to follow. Can you either comment or add some type aliases to make it clear what this is returning?

cmd/spd/src/lib.rs
```diff
@@ -387,12 +484,29 @@ fn spd(context: &mut ExecutionContext) -> Result<()> {
...
+    core: &mut dyn humility::core::Core,
+    subargs: &SpdArgs,
+) -> Result<()> {
+    // Warn the user that we probably can't talk to DDR4s on non-Gimlet hardware
+    if !hubris.manifest.target.as_ref().is_some_and(|t| t.contains("gimlet")) {
+        humility::warn!(
+            "trying to talk to DDR4 SPDs on an invalid target `{}`",
+            hubris.manifest.target.as_deref().unwrap_or("<unknown>")
+        );
+    };
```
labbott, line 520: Who do we expect would see this message and are they going to know what to do when they see it?

labbott (approved):
No real issues, mostly clarification for the next person who needs to look at it

cmd/spd/src/lib.rs
mkeeter: I'll wrap the inner `Vec<u8>` in a `struct SpdData<Vec<u8>>`.

cmd/spd/src/lib.rs
mkeeter: This message will be seen if someone runs `humility spd` with an attached non-Gimlet SP, and manually specifies an I2C bus or controller.  I think anyone doing that will know what to do when they see this error message (rethink their choices), since they should know that (for example) Cosmo does not have DDR4 DIMMs.

Issue #202, opened 2022-08-04
Humility can't find some CMSIS-DAP probes on illumos


bcantrill:
On illumos and with some CMSIS-DAP probes (notably, NXP's MCU Link), Humility will fail to find the probe:

```console
# humility probe
humility probe failed: Probe could not be created

Caused by:
    Probe was not found.
```

Running with `-v` sheds a little more light -- but only a little:

```console
# humility -v probe
[2022-07-21T21:40:23Z DEBUG probe_rs::probe::cmsisdap::tools] Searching for CMSIS-DAP probes using libusb
[2022-07-21T21:40:23Z TRACE probe_rs::probe::cmsisdap::tools] MCU-LINK r0FF CMSIS-DAP V0.078: CMSIS-DAP device with 4 interfaces
[2022-07-21T21:40:23Z TRACE probe_rs::probe::cmsisdap::tools]   Interface 0: MCU-LINK CMSIS-DAP V0.078
[2022-07-21T21:40:23Z TRACE probe_rs::probe::cmsisdap::tools] Will use interface number 0 for CMSIS-DAPv1
[2022-07-21T21:40:23Z DEBUG probe_rs::probe::cmsisdap::tools] Found 1 CMSIS-DAP probes using libusb, searching HID
[2022-07-21T21:40:23Z DEBUG probe_rs::probe::cmsisdap::tools] Found 1 CMSIS-DAP probes total
[2022-07-21T21:40:23Z DEBUG jaylink] libusb 1.0.25.11692
[2022-07-21T21:40:23Z DEBUG jaylink] libusb has capability API: true
[2022-07-21T21:40:23Z DEBUG jaylink] libusb has HID access: false
[2022-07-21T21:40:23Z DEBUG jaylink] libusb has hotplug support: false
[2022-07-21T21:40:23Z DEBUG jaylink] libusb can detach kernel driver: false
[2022-07-21T21:40:23Z TRACE probe_rs::probe::cmsisdap::tools] Attempting to open device matching 1fc9:0143:XAEPYT3E3JX0D
[2022-07-21T21:40:23Z TRACE probe_rs::probe::cmsisdap::tools] Trying device Bus 208 Device 003: ID 1fc9:0143
[2022-07-21T21:40:23Z TRACE probe_rs::probe::cmsisdap::tools] MCU-LINK r0FF CMSIS-DAP V0.078: CMSIS-DAP device with 4 interfaces
[...]

PR #619, opened 2026-04-16
Update to probe-rs v0.31.0

labbott:


sdonnan:
Windows test run:

```
PS E:\stuart\working\humility> git log -1 --oneline
02547623 (HEAD -> probe-rs-v0.31.0, origin/probe-rs-v0.31.0) WIP: hold my beer I want to update probe-rs

PS E:\stuart\working\humility> cargo run --release -- lsusb
    Finished `release` profile [optimized + debuginfo] target(s) in 0.65s
     Running `target\release\humility.exe lsusb`
humility: USB device scan, 2 successful and 6 failed
humility: --- successfully opened devices ---
humility: format: VID:PID:SERIAL, then manufacturer name, then product name
humility: 046d:c628:(serial unknown)    3Dconnexion     SpaceNavigator for Notebooks
humility: 38c6:0001:000000803b5a95305d5cc175fa18799758892acd4c504355    Oxide   CMSIS-DAP v1
humility: --- failures ---
humility: could not access 6 devices:
... snip ...

PS E:\stuart\working\humility> cargo run --release -- probe
    Finished `release` profile [optimized + debuginfo] target(s) in 0.64s
     Running `target\release\humility.exe probe`
humility: attached via CMSIS-DAP V2 IF: 4 DESC: Some(InterfaceAltSetting { interface_number: 4, alternate_setting: 0, num_endpoints: 2, class: 255, subclass: 0, protocol: 0, string_index: Some(9), endpoints: [Endpoint { address: 0x05, direction: Out, transfer_type: Bulk, max_packet_size: 512, packets_per_microframe: 1, interval: 0 }, Endpoint { address: 0x85, direction: In, transfer_type: Bulk, max_packet_size: 512, packets_per_microframe: 1, interval: 0 }] })
humility:        probe => OxLink (CMSIS-DAP), VID 38c6, PID 0001
humility: probe serial => 000000803b5a95305d5cc175fa18799758892acd4c504355
humility:         core => Cortex-M7
humility: manufacturer => STMicroelectronics
humility:         chip => STM32H7, revision 0x2003
humility:       status => executing, debug halt
humility:  debug units => CSTF(x2) CTI(x2) DWT ETM FPB ITM SCS SWO TMC TPIU
humility:         CSTF => 0x5c004000, 0x5c013000
humility:          CTI => 0x5c011000, 0xe0043000
humility:          DWT => 0xe0001000
humility:          ETM => 0xe0041000
humility:          FPB => 0xe0002000
humility:          ITM => 0xe0000000
humility:          SCS => 0xe000e000
humility:          SWO => 0x5c003000
humility:          TMC => 0x5c014000
humility:         TPIU => 0x5c015000
humility:           R0 => 0x24078900
humility:           R1 => 0x24078900
humility:           R2 => 0x0
humility:           R3 => 0x0
humility:           R4 => 0x0
humility:           R5 => 0x0
humility:           R6 => 0x0
humility:           R7 => 0x240788f8
humility:           R8 => 0x0
humility:           R9 => 0x0
humility:          R10 => 0x0
humility:          R11 => 0x0
humility:          R12 => 0x0
humility:           SP => 0x240788f8
humility:           LR => 0x805fd6f
humility:           PC => 0x805fd74
humility:          PSR => 0x61000000
humility:          MSP => 0x240003b0
humility:          PSP => 0x240788f8
humility:          SPR => 0x7000000
```

Seems happy!

cmd/debugmailbox/src/lib.rs
```diff
@@ -269,14 +253,15 @@ fn debugmailboxcmd(context: &mut ExecutionContext) -> Result<()> {
     let subargs = DebugMailboxArgs::try_parse_from(&context.cli.cmd)?;
 
     // Get a list of all available debug probes.
-    let probes = Probe::list_all();
+    let list = probe_rs::probe::list::Lister::new();
+    let probes = list.list_all();
```
mkeeter, line 257: ah, I love the smell of fresh API design in the morning

cmd/rendmp/src/lib.rs
```diff
@@ -166,7 +166,7 @@ use std::io::Write;
 use std::io::prelude::*;
 use std::thread;
 use std::time::{Duration, Instant};
-use strum::VariantNames;
+use strum::VariantNames as _;
```
mkeeter, line 168: spooky

humility-probes-core/src/lib.rs
```diff
@@ -2,7 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use ::probe_rs::{
+use ::probe_rs::Permissions;
+use ::probe_rs::probe::{
```
mkeeter: Vibes: any reason not to write this as one big import, e.g.
```
use ::probe_rs::{Permissions, probe::{ ... }}
```
?

humility-probes-core/src/lib.rs
```diff
@@ -60,43 +62,31 @@ fn get_usb_probe(index: Option<usize>) -> Result<DebugProbeInfo> {
...
         }
     }
 
-    if let Err(DebugProbeError::Usb(Some(ref err))) = res
-        && let Some(rcode) = err.downcast_ref::<rusb::Error>()
-        && *rcode == rusb::Error::Busy
-    {
-        bail!("USB link in use; is OpenOCD or another debugger running?");
-    }
-
```
mkeeter, line 104: It looks like the `Usb` variant still exists, should we keep this block?

humility-probes-core/src/lib.rs
```diff
@@ -117,7 +107,7 @@ pub fn attach_to_probe(
         "usb" => {
             let probe_info = get_usb_probe(index)?;
 
-            let probe = open_probe(&probe_info, speed_khz)?;
+            let probe = probe_info.open()?;
```
mkeeter: Deliberately ignoring `speed_khz` here?

humility-probes-core/src/lib.rs
```diff
@@ -160,20 +150,30 @@ pub fn attach_to_chip(
...
-                Some(chip) => (probe.attach(chip)?, true),
-                None => (probe.attach("armv7m")?, false),
+                Some(chip) => (
+                    probe.attach(chip, Permissions::new().allow_erase_all())?,
+                    true,
+                ),
+                None => (
+                    probe.attach(
+                        "Cortex-M3",
+                        Permissions::new().allow_erase_all(),
```
mkeeter: Why do we need `allow_erase_all` if we're not going to flash?

humility-probes-core/src/lib.rs
```diff
@@ -189,24 +189,36 @@ pub fn attach_to_chip(
...
+                        probe.attach(
+                            chip,
+                            Permissions::new().allow_erase_all(),
+                        )?,
+                        true,
+                    ),
+                    None => (
+                        probe.attach(
+                            "Cortex-M3",
+                            Permissions::new().allow_erase_all(),
```
mkeeter: (same here)

humility-probes-core/src/probe_rs.rs
```diff
@@ -68,6 +71,155 @@ impl ProbeCore {
     }
 }
 
+// XXX borrowed???
```
mkeeter: ...XXX borrowed???

humility-probes-core/src/probe_rs.rs
```diff
@@ -205,85 +357,51 @@ impl Core for ProbeCore {
...
+                progress_bars.get_mut(operation.into()).inc(size);
             }
-
-            flashing::ProgressEvent::FinishedProgramming => {
-                bar.finish_and_clear();
+            ProgressEvent::Failed(operation) => {
+                progress_bars.get_mut(operation.into()).abandon();
+            }
+            ProgressEvent::Finished(operation) => {
+                progress_bars.get_mut(operation.into()).finish();
```
mkeeter: Nit: you could save a few `.into()` calls by making `get_mut` take a `<T: Into<Operation>>` (and moving the `.into()` into that function)

Cargo.toml
```diff
@@ -105,9 +105,12 @@ vsc7448-types = { git = "https://github.com/oxidecomputer/vsc7448.git" }
 
 #
 # We depend on the oxide-stable branch of Oxide's fork of probe-rs to assure
-# that we can float necessary patches on probe-rs.
+# that we can float necessary patches on probe-rs.
```
mkeeter: Nit: linewrap this whole block?

cmd/debugmailbox/src/lib.rs
```diff
@@ -104,9 +105,9 @@ enum DebugMailboxCmd {
 }
 
 fn poll_raw_ap_register(
-    probe: &mut Box<dyn ArmProbeInterface + '_>,
-    ap: &ApAddress,
-    addr: u8,
+    probe: &mut Box<dyn ArmDebugInterface + '_>,
```
mkeeter, line 107: I wondered if we could remove the `Box` here (in favor of a `&mut dyn ArmDebugInterface`), but poking at it shows that it would ripple through a bunch of places, so out of scope for this PR.

humility-probes-core/src/lib.rs
mkeeter: (and in a few places in the diff below)

humility-probes-core/src/lib.rs
labbott: We now get back a raw `std::io::Error` which does not  have `downcast_ref`, only `downcast` (https://doc.rust-lang.org/stable/std/io/struct.Error.html) and it was faster to just pull it out when I was doing the upgrade. It wasn't clear if there was still value in keeping it. Looking again, the flow of this entire function is weird so I'm going to take a pass to see if I can make it more reasonable.

labbott: I did some tweaks and I think I brought it back. I'm still not clear on the value but if we run into more errors that need better messages we can add more.

humility-probes-core/src/lib.rs
labbott: I think I now need an explicit call to `probe.set_speed`

humility-probes-core/src/probe_rs.rs
labbott: Haha this was a note I made that this was based on what probe-rs was doing. I'll clean it up to clarify.

labbott:
This discussion may need to go elsewhere but I'm putting it here for now:

One of the advantages of updating to a newer probe-rs is we can avoid a dependency on `libusb` and just use the native Rust `nusb` library. The catch is this only works with CMSIS-DAP v2 probes. Our out of the box MCULinks are running old firmware and only support CMSIS-DAP v1 which uses the `hidapi` crate which is secretly `libusb` . We've had to do some hacks to make this work (see https://github.com/oxidecomputer/humility/issues/202 and https://github.com/oxidecomputer/humility/issues/466). Some options here:

A) We allow using CMSIS-DAP v1 probes and continue supporting HID via libusb. I do not love this option because 1) We've had problems with hid/libusb in the past and we'll have to continue with our hacks 2) We'll now have to deal with a potential mixture of CMSIS-DAPv1 and CMSIS-DAP v2 probes and two different libraries to debug.

B) We require all MCULinks to be updated to CMSIS DAP v2 firmware. I didn't catch this issue initially because I had already done so. We have reliable instructions on how to do this https://github.com/oxidecomputer/meta/blob/master/engineering/mculink-upgrade.md , the downside is we will need to make sure this happens across our fleet of probes in the lab and manufacturing.

C) We use OxLink which comes with updated CMSIS-DAP firmware and works great.

I expect we will want some combination of B & C depending on how fast we can produce OxLinks . I am not a fan of A.

As far as I can tell this is a concern only for MCULinks not STLinks

sdonnan:
> One of the advantages of updating to a newer probe-rs is we can avoid a dependency on libusb and just use the native Rust nusb library. The catch is this only works with CMSIS-DAP v2 probes. Our out of the box MCULinks are running old firmware and only support CMSIS-DAP v1

We are actively looking into upgrading existing hardware (either by firmware update or hardware replacement). Manufacturing will see the greatest impact, as each loom will have to be replaced or modified. In the near future (July) there are other upgrades planned and it may make sense to lump all these things together.

I think it is important that Humility releases continue to support CMSIS-DAP v1 until we have deployed all the necessary hardware upgrades.

labbott:
I went ahead and just kept the `cmsisdap_v1` feature on permanently for now (with a very big comment saying we'd like to get rid of it in The Future). I went back and forth about making it option but I think that will just cause more problems than it solves.

humility-bin/src/main.rs
```diff
@@ -226,16 +225,25 @@ fn main() -> std::process::ExitCode {
...
 
-    let env = env_logger::Env::default().filter_or("RUST_LOG", log_level);
-
-    env_logger::init_from_env(env);
+    // probe-rs now does tracing
+    tracing_subscriber::fmt()
+        .with_env_filter(
+            EnvFilter::builder().with_default_directive(level).from_env_lossy(),
+        )
+        .init();
```
labbott, line 254: @mkeeter @hawkw is this going to cause more problems when either we try using `slog` or something else?

mkeeter: Nah, `slog` does its own configuration so this should be fine (other than I'll have to resolve a small merge conflict).

However, we may want to keep `env_logger::init_from_env(env)`, since there are other dependencies which still use the `log` crate (I see `lpc55_sign` and `goblin`).

cmd/lsusb/src/lib.rs
```diff
@@ -32,17 +33,19 @@ fn lsusb(_args: LsUsbArgs, context: &mut ExecutionContext) -> Result<()> {
...
+    for dev in devices {
         match list1(&dev) {
             Ok(summary) => successes.push(summary),
             Err(e) => {
                 failures.push((
-                    dev.bus_number(),
-                    dev.address(),
-                    dev.port_number(),
+                    // `bus_id` as an &'str is common
+                    // across all OSes
```
mkeeter: Nit:
```suggestion
                    // `bus_id` as an &'str is common across all OSes
```

cmd/lsusb/src/lib.rs
```diff
@@ -72,7 +75,7 @@ fn lsusb(_args: LsUsbArgs, context: &mut ExecutionContext) -> Result<()> {
     }
 
     if !failures.is_empty() {
-        failures.sort_by_key(|(b, a, p, _)| (*b, *a, *p));
+        failures.sort_by_key(|(b, a, p, _)| (b.clone(), *a, *p));
```
mkeeter: I'm sad that you can't do `b.as_str()` here.

This could be done without any allocations with `sort_by`, e.g.
```rust
        failures.sort_by(|lhs, rhs| {
            (&lhs.0, lhs.1, lhs.2).cmp(&(&rhs.0, rhs.1, rhs.2))
        });
```
(but probably not worth the extra boilerplate)

cmd/lsusb/src/lib.rs
```diff
@@ -104,36 +107,51 @@ fn lsusb(_args: LsUsbArgs, context: &mut ExecutionContext) -> Result<()> {
...
-    let pid = desc.product_id();
+fn list1(dev: &nusb::DeviceInfo) -> Result<(String, String)> {
+    let timeout = Duration::from_millis(100);
+    let vid = dev.vendor_id();
+    let pid = dev.product_id();
+
+    let dev = dev
+        .open()
+        .wait()
+        .map_err(|e| anyhow!("{vid:04x}:{pid:04x}:???\topen failed: {e}"))?;
```
mkeeter: Nit: don't format the error yourself
```suggestion
        .with_context(|| format!("open failed for {vid:04x}:{pid:04x}:???"))?;
```

cmd/lsusb/src/lib.rs
```diff
@@ -104,36 +107,51 @@ fn lsusb(_args: LsUsbArgs, context: &mut ExecutionContext) -> Result<()> {
...
+
+    let languages: Vec<u16> = dev
+        .get_string_descriptor_supported_languages(timeout)
+        .wait()
+        .map(|i| i.collect())
+        .unwrap_or_default();
+
+    let language = languages.first().copied().unwrap_or(US_ENGLISH);
+
+    let man = if let Some(i_manufacturer) =
```
mkeeter: You can condense this with `and_then`, which is shorter and avoids repeating the string in `unwrap_or`:
```rust
    let man = dev_descriptor
        .manufacturer_string_index()
        .and_then(|i| {
            dev.get_string_descriptor(i, language, timeout).wait().ok()
        })
        .unwrap_or_else(|| "(manufacturer unknown)".to_string());
```
This also applies to `prod` and `serial` below.

(I also switched to `unwrap_or_else` to avoid allocating the string if it's not needed, but that's just my personal pedantry)

humility-bin/Cargo.toml
```diff
@@ -18,7 +18,7 @@
 [package]
 name = "humility-bin"
 edition.workspace = true
-version = "0.12.18"
+version = "0.13.0"
```
mkeeter, line 21: 🚀

humility-bin/Cargo.toml
```diff
@@ -107,6 +107,7 @@ scroll = { workspace = true }
 indicatif = { workspace = true }
 colored = { workspace = true }
 indexmap = { workspace = true }
+tracing-subscriber = { workspace = true }
```
mkeeter, line 110: If we're not using `env_logger`, then it can be removed (but I think we do want to keep it, see comment above)

humility-probes-core/src/lib.rs
```diff
@@ -60,44 +64,43 @@ fn get_usb_probe(index: Option<usize>) -> Result<DebugProbeInfo> {
...
-                "Could not find probe {}.\n\
+    let lister = ::probe_rs::probe::list::Lister::new();
+    let mut probe = match lister.open(selector.clone()) {
+        Ok(p) => p,
+        Err(DebugProbeError::ProbeCouldNotBeCreated(
+            ProbeCreationError::NotFound,
+        )) => {
+            if selector.serial_number.is_some() {
+                bail!(
+                    "Could not find probe {}.\n\
```
mkeeter: Nit: fix indentation of the continued text below.  We could also inline `probe_selector` into the format string.

mkeeter:
Two small notes:

The attachment log is now awkwardly wordy (at least on the RoT + OxLink):
```
humility: attached via CMSIS-DAP V2 IF: 4 DESC: Some(InterfaceAltSetting { interface_number: 4, alternate_setting: 0, num_endpoints: 2, class: 255, subclass: 0, protocol: 0, string_index: Some(9), endpoints: [Endpoint { address: 0x05, direction: Out, transfer_type: Bulk, max_packet_size: 512, packets_per_microframe: 1, interval: 0 }, Endpoint { address: 0x85, direction: In, transfer_type: Bulk, max_packet_size: 512, packets_per_microframe: 1, interval: 0 }] })
```

Progress bars are fighting with errors:
Screenshot 2026-06-04 at 11 08 52 AM

(I think we should fix the former, but can probably punt on the latter unless it's easy)

labbott:
> The attachment log is now awkwardly wordy (at least on the RoT + OxLink):

`CMSIS-DAP V2 IF: 4 DESC: Some(InterfaceAltSetting { interface_number: 4, alternate_setting: 0, num_endpoints: 2, class: 255, subclass: 0, protocol: 0, string_index: Some(9), endpoints: [Endpoint { address: 0x05, direction: Out, transfer_type: Bulk, max_packet_size: 512, packets_per_microframe: 1, interval: 0 }, Endpoint { address: 0x85, direction: In, transfer_type: Bulk, max_packet_size: 512, packets_per_microframe: 1, interval: 0 }] })` is a beautiful name for a child.

I'll see if we can print some more specific fields instead.

> Progress bars are fighting with errors:

I'm moderately tempted to add another patch to probe-rs to silence the warning because we were already carrying a similar patch from probe-rs 0.12.0

humility-probes-core/src/lib.rs
labbott: this was a `cargo fmt` special, I think I aligned it

labbott:
`humility: attached via 1fc9:143:Some("NAOFKQDS5OCZ0") MCU-LINK (r0FB) CMSIS-DAP V3.172`

We can play around with this

cmd/lsusb/src/lib.rs
labbott: we're collecting errors so they have to get printed sometime. I moved the formatting to the call site:
```
humility: USB device scan, 2 successful and 7 failed
humility: --- successfully opened devices ---
humility: format: VID:PID:SERIAL, then manufacturer name, then product name
humility: 0403:6001:A1049BI6	FTDI	FT232R USB UART
humility: 1fc9:0143:NAOFKQDS5OCZ0	NXP Semiconductors	MCU-LINK (r0FB) CMSIS-DAP V3.172
humility: --- failures ---
humility: could not access 7 devices:
humility: bus 001, addr 3, port 5: open failed for 046d:0892:???: failed to open device (errno 13)
humility: bus 001, addr 4, port 6: open failed for 3142:0002:???: failed to open device (errno 13)
humility: bus 001, addr 5, port 7: open failed for 045e:082c:???: failed to open device (errno 13)
humility: bus 001, addr 6, port 8: open failed for 1a7c:0191:???: failed to open device (errno 13)
humility: bus 001, addr 7, port 10: open failed for 0bda:5411:???: failed to open device (errno 13)
humility: bus 002, addr 2, port 9: open failed for 0bda:0328:???: failed to open device (errno 13)
humility: bus 002, addr 3, port 10: open failed for 0bda:0411:???: failed to open device (errno 13)
```

humility-bin/src/main.rs
labbott: I don't think we can use both, I tried with both orderings of `env_logger::init_from_env(env)` and the `tracing_subscriber::fmt().init()`

```
labbott@labbott-desktop2:~/humility$ ./target/debug/humility probe

thread 'main' (3793347) panicked at /home/labbott/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/env_logger-0.11.10/src/logger.rs:969:10:
env_logger::init_from_env should not be called after logger initialized: SetLoggerError(())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
labbott@labbott-desktop2:~/humility$ cargo build
   Compiling humility-bin v0.13.0 (/home/labbott/humility/humility-bin)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.06s
labbott@labbott-desktop2:~/humility$ ./target/debug/humility probe

thread 'main' (3793751) panicked at /home/labbott/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tracing-subscriber-0.3.23/src/fmt/mod.rs:520:14:
Unable to install global subscriber: SetLoggerError(())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I'm lightly inclined to leave `env_logger` off for now

mkeeter (approved):

humility-bin/src/main.rs
mkeeter: Interesting, I had assumed they'd be independent!  In that case, I'm fine leaving `env_logger` off (since if it's complaining, that presumably means we've got a global logger installed)

PR #620, opened 2026-04-17
Add `humility ereport` subcommnad

mkeeter:
This PR adds a `humility ereport` subcommand to dump ereports from packrat.

```console
$ humility -d ~/Desktop/hubris.core.ereports ereport dump
humility: attached to dump
task: ereportulator (27)
timestamp: 9441
  k: hw.apollo.undervolt,
  v: 13,
  volts: 0,
  n: 5678,
  bus: "MainBusB"
```

It includes both normal and JSON formatting, e.g.
```json
[
  {
    "task_name": "ereportulator",
    "task_index": 27,
    "timestamp": 9441,
    "recovered": false,
    "contents": {
      "k": "hw.apollo.undervolt",
      "v": 13,
      "volts": 0.0,
      "n": 5678,
      "bus": "MainBusB"
    }
  }
]
```

By default, it prints just ereports which are in the valid portion of the circular buffer; the `--recover` flag also searches for ereports which have been sent but are still available in RAM.

This is written to be library-friendly: `pub fn ereport_dump(..)` does no printing of its own, just returns a `Result<Vec<Ereport>>`.

There are also some incidental improvements to reflection, which I'd like to use more widely in a subsequent PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ members = [
     "cmd/dump",
     "cmd/tofino-eeprom",
     "cmd/discover",
+    "cmd/ereport",
     "cmd/exec",
     "cmd/extract",
     "cmd/flash",
@@ -142,6 +143,7 @@ cmd-discover = { path = "./cmd/discover", package = "humility-cmd-discover" }
 cmd-doc = { path = "./cmd/doc", package = "humility-cmd-doc" }
 cmd-dump = { path = "./cmd/dump", package = "humility-cmd-dump" }
 cmd-tofino-eeprom = { path = "./cmd/tofino-eeprom", package = "humility-cmd-tofino-eeprom" }
+cmd-ereport = { path = "./cmd/ereport", package = "humility-cmd-ereport" }
 cmd-exec = { path = "./cmd/exec", package = "humility-cmd-exec" }
 cmd-extract = { path = "./cmd/extract", package = "humility-cmd-extract" }
 cmd-flash = { path = "./cmd/flash", package = "humility-cmd-flash" }
@@ -195,9 +197,10 @@ anyhow = { version = "1.0.44", features = ["backtrace"] }
 atty = "0.2"
 bitfield = "0.13.2"
 byteorder = "1.3.4"
-cargo_metadata = "0.12.0"
 cargo-readme = "3.3.1"
+cargo_metadata = "0.12.0"
 chrono = "0.4.38"
+ciborium = "0.2.2"
 clap = { version = "3.0.12", features = ["derive", "env"] }
 colored = "2.0.0"
 crc-any = "2.3.5"
@@ -209,6 +212,7 @@ env_logger = "0.9.0"
 fallible-iterator = "0.3.0"
 gimli = "0.32.0"
 goblin = "0.10"
+hex = "0.4.3"
 hubpack = "0.1.1"
 ihex = "3.0"
 indexmap = { version = "1.7", features = ["serde-1"] }
--- a/cmd/ereport/src/lib.rs
+++ b/cmd/ereport/src/lib.rs
@@ -40,12 +40,20 @@ struct Flags {
     /// Print ereports in JSON format
     #[clap(short, long)]
     json: bool,
+
     /// Recover previously-sent ereports from the buffer
+    ///
+    /// The circular buffer may contain previously-sent ereports outside of its
+    /// valid range, i.e. before the `front` index.  When this flag is set, we
+    /// examine those bytes to find valid-looking ereports and print them (with
+    /// `recovered: true`).
     #[clap(short, long)]
     recover: bool,
 }
 
-/// Ereport header (required to agree with SP implementation)
+/// Raw ereport header
+///
+/// This must agree with the SP's implementation in `snitch_core::Store`
 #[derive(Debug, FromBytes)]
 #[repr(C)]
 struct EreportHeader {
@@ -62,6 +70,7 @@ pub struct Ereport {
     task_index: u16,
     timestamp: u64,
     recovered: bool,
+    ena: u64,
     contents: ciborium::Value,
 }
 
@@ -81,12 +90,20 @@ pub fn ereport_dump(
         .with_context(|| format!("could not find `{PACKRAT_BUF_NAME}`"))?;
 
     let buf: Value = humility::reflect::load_variable(hubris, core, buf_ty)?;
-    let storage: Value =
-        buf.field("cell.value.ereport_bufs.storage.storage")?;
+    let outer_storage: Value = buf.field("cell.value.ereport_bufs.storage")?;
+    let inner_storage: Value = outer_storage.field("storage")?;
     let array: Vec<humility_doppel::MaybeUninit<u8>> =
-        storage.field("buffer")?;
-    let front = storage.field::<u32>("front")? as usize;
-    let back = storage.field::<u32>("back")? as usize;
+        inner_storage.field("buffer")?;
+    let front = inner_storage.field::<u32>("front")? as usize;
+    let back = inner_storage.field::<u32>("back")? as usize;
+    let earliest_ena = outer_storage.field::<u64>("earliest_ena")?;
+    let insert_state =
+        outer_storage.field::<humility::reflect::Enum>("insert_state")?;
+    if insert_state.disc() != "Collecting" {
+        humility::warn!(
+            "ereports are being lost; `insert_state` is {insert_state:#?}"
+        );
+    }
 
     // Read data from the circular buffer
     let mut buf_data = Vec::with_capacity(array.len());
@@ -97,6 +114,7 @@ pub fn ereport_dump(
     }
     let mut ereports = vec![];
     let mut buf_data = buf_data.as_slice();
+    let mut current_ena = earliest_ena;
     while !buf_data.is_empty() {
         let header =
             EreportHeader::read_from_prefix(buf_data).ok_or_else(|| {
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -197,9 +199,10 @@ anyhow = { version = "1.0.44", features = ["backtrace"] }
 atty = "0.2"
 bitfield = "0.13.2"
 byteorder = "1.3.4"
-cargo_metadata = "0.12.0"
 cargo-readme = "3.3.1"
+cargo_metadata = "0.12.0"
 chrono = "0.4.38"
+ciborium = "0.2.2"
```
hawkw, line 205: what's the rationale for `ciborium` here? in Hubris we use `minicbor`, and [MGS uses  `serde_cbor`](https://github.com/oxidecomputer/management-gateway-service/blob/e6793bfc4f9a663d7a7b4b3868e5c2c7f8179174/Cargo.toml#L55). this is very much *not* intended as me saying that using `ciborium` is wrong, but instead, me wondering whether we should be using it in MGS as well, instead of `serde_cbor`...

humility-core/src/reflect.rs
```diff
@@ -131,6 +131,27 @@ impl Value {
         }
     }
 
+    /// Looks up a `struct` field with dot-separated members
+    pub fn field<T: Load>(&self, f: &str) -> Result<T> {
+        let mut v = self;
+        for f in f.split(".") {
+            let Value::Struct(s) = v else {
+                bail!("expected a struct when getting field `{f}`");
```
hawkw, line 139: turbo nitpick, feel free to ignore: mayhaps this should include a 
```suggestion
                bail!("expected a struct when getting field `{f}`, but found {v:?} instead");
```

cmd/ereport/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+    verbose: bool,
+    /// Print ereports in JSON format
+    #[clap(short, long)]
+    json: bool,
+    /// Recover previously-sent ereports from the buffer
+    #[clap(short, long)]
+    recover: bool,
+}
+
+/// Reasons a server might cite when using the `REPLY_FAULT` syscall.
```
hawkw, line 51: hm, i don't think that's what this is?

cmd/ereport/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+}
+
+#[derive(Parser, Debug)]
+struct Flags {
+    /// Print verbose info while processing ereports
+    #[clap(short, long, conflicts_with = "json")]
+    verbose: bool,
+    /// Print ereports in JSON format
+    #[clap(short, long)]
+    json: bool,
```
hawkw, line 45: nit, take it or leave it: i don't actually think that it is not actually necessary for these to conflict, if you decided to write the verbose info to `stderr` and the JSON to `stdout`? that way,  i could do `$ humility ereport dump --verbose --json > ereports.json` and get the verbose info to the console...?

cmd/ereport/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+        let task = hubris.lookup_module(HubrisTask::Task(header.task.into()));
+        let contents = ciborium::from_reader(ereport_data)?;
+        ereports.push(Ereport {
+            task_name: task.map(|t| t.name.clone()).ok(),
+            task_index: header.task.into(),
+            timestamp: header.timestamp.into(),
+            recovered: false,
+            contents,
+        });
+        buf_data = next;
```
hawkw, line 122: I would quite like it if the ereport dump also included the ereport's ENA. I think we ought to be able to do that by additionally looking up the [`earliest_ena` field](https://github.com/oxidecomputer/hubris/blob/e105132ac67b0110befbaa93d6f6e2772d7163cd/lib/snitch-core/src/lib.rs#L38-L46), assigning that ENA to the first ereport, and incrementing it by 1 for each subsequent ereport we encounter.

We could also probably do this when recovering, by counting backwards.

cmd/ereport/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+        storage.field("buffer")?;
+    let front = storage.field::<u32>("front")? as usize;
+    let back = storage.field::<u32>("back")? as usize;
+
+    if flags.verbose {
+        println!(
+            "buffer has {} bytes; front: {front}, back: {back}",
+            array.len()
+        );
+    }
```
hawkw, line 94: It might be nice if this code looked up the [`InsertState`](https://github.com/oxidecomputer/hubris/blob/e105132ac67b0110befbaa93d6f6e2772d7163cd/lib/snitch-core/src/lib.rs#L48-L50) and printed whether or not it was currently in the midst of losing data, at least when verbose mode is  set?

cmd/ereport/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+        ciborium::Value::Tag(tag, inner) => {
+            print!("tag({tag}) ");
+            pretty_print_value(inner, indent, false);
+        }
+
+        _ => print!("<unknown>"),
+    }
+}
```
hawkw, line 280: one thing that would be nice about potentially using the same CBOR deserialization library as MGS is that if we did that, we could perhaps share the pretty-printer with `faux-mgs`. not a blocker for this PR, but just a thought.

cmd/ereport/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+            for (i, (k, v)) in entries.iter().enumerate() {
+                print!("{pad}  ");
+                pretty_print_value(k, indent + 1, true);
+                print!(": ");
+                // Special-case handling for the `k`, which is an ereport name
+                // and is printed without quotes
+                let is_ereport_key = match k {
+                    ciborium::Value::Text(t) => t == "k",
+                    _ => false,
+                };
```
hawkw, line 260: if we're going to special-case "k", i think we should just be renaming it to "class", which is what it abbreviates:
```suggestion
                // Special-case handling for the `k`, which is an ereport name
                // and is printed without quotes
                let is_ereport_key = match k {
                    ciborium::Value::Text(t) => t == "class",
                    _ => false,
                };
```

we should probably special-case the "v" field as "version", as well.

Cargo.toml
mkeeter: As far as I can tell, `minicbor` has no equivalent of dynamic objects, e.g. [`ciborium::Value`](https://docs.rs/ciborium/latest/ciborium/enum.Value.html).  You can either deserialize to a specific type, or iterate over tokens.  Pretty-printing by iterating over tokens is a pain, **and** I'd like to store the `Value` in a `struct Ereport` for other consumers of this API.

`serde_cbor` has been [unmaintained for 5 years](https://github.com/pyfisch/cbor/commit/347a3f096d5a1e8e380b9968e74838c9037a5b14), and `ciborium` is recommended as a replacement; MGS should probably switch!

humility-core/src/reflect.rs
mkeeter: Seems reasonable, done!

cmd/ereport/src/lib.rs
mkeeter: Whoops. forgot to push the code after doing a self-review

cmd/ereport/src/lib.rs
mkeeter: I've decided I don't actually like `verbose`, since it just prints an extra line showing buffer extents (which you can get from `readvar` without too much pain).

I do want to keep printing either the decoded info or JSON to `stdout`, though, for consistency with other Humility commands (with log messages going to `stderr`).

cmd/ereport/src/lib.rs
mkeeter: Seems reasonable, although this diff doesn't work 🙃 

I'm adding a remapping pass of `k -> class`, `v -> version` at the end of `fn ereport_dump(..)`, so that both the human-friendly and JSON outputs use those names.

cmd/ereport/src/lib.rs
mkeeter: Sure, done.  I'm logging it with `humility::warn!` so it shows up in red.

Cargo.toml
hawkw: Yup, I think that makes sense --- we should probably go change MGS then!

cmd/ereport/src/lib.rs
hawkw: > Seems reasonable, although this diff doesn't work 🙃

Argh, my bad --- thanks for doing it the correct way! :)

cmd/ereport/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+struct Flags {
+    /// Print ereports in JSON format
+    #[clap(short, long)]
+    json: bool,
+    /// Recover previously-sent ereports from the buffer
+    #[clap(short, long)]
+    recover: bool,
+}
+
+/// Ereport header (required to agree with SP implementation)
```
hawkw, line 48: might be nice if there was a link here to where the SP implementation to which this must agree, or something?

cmd/ereport/src/lib.rs
mkeeter: Whoops, this is now done as well.

cmd/ereport/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+}
+
+#[derive(Parser, Debug)]
+struct Flags {
+    /// Print ereports in JSON format
+    #[clap(short, long)]
+    json: bool,
+    /// Recover previously-sent ereports from the buffer
+    #[clap(short, long)]
+    recover: bool,
```
hawkw, line 45: i feel like the doc for this could maybe be a little more clear about what this means?

cmd/ereport/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+        timestamp: header.timestamp.into(),
+        recovered: true,
+        ena: *ena,
+        contents,
+    })
+}
+
+/// Recursively pretty-print a CBOR value
+fn pretty_print_value(value: &ciborium::Value, indent: usize, is_key: bool) {
+    let pad = "  ".repeat(indent);
```
hawkw, line 252: if you wanted to do this without allocating, you could instead write something like this in your format strings:

```rust
println!("{:indent$}stuff you want indented", "");
```

not actually important, i just think it's nice.

hawkw (approved):

cmd/ereport/src/lib.rs
mkeeter: I can never remember the syntax here (and don't love the extra trailing `""` argument), so I'm fine with a little extra allocation.

cmd/ereport/src/lib.rs
mkeeter: Yup, added

cmd/ereport/src/lib.rs
mkeeter: Sure, added a longer docstring (which `clap` shows when you pass in `--help`)
```
humility-ereport-dump
Print the contents of the ereport buffer

USAGE:
    humility ereport dump [OPTIONS]

OPTIONS:
    -h, --help
            Print help information

    -j, --json
            Print ereports in JSON format

    -r, --recover
            Recover previously-sent ereports from the buffer

            The circular buffer may contain previously-sent ereports outside of
            its valid range, i.e. before the `front` index.  When this flag is
            set, we examine those bytes to find valid-looking ereports and print
            them (with `recovered: true`).
```

PR #621, opened 2026-04-20
remove `humility apptable`

labbott:
We haven't supported this format since 2022


mkeeter (approved):
I approve in principle, though CI has _opinions_

labbott:
> I approve in principle, though CI has _opinions_

There are a bunch of old dumps which are apptable only. I'm going to see if I can prune those without losing too much.

labbott:
My modest proposal is to just delete those dumps since the output would now be "not an apptable" https://github.com/oxidecomputer/humility/pull/622

PR #627, opened 2026-04-23
Move auxflash core implementation into `humility-auxflash`

mkeeter:
This removes one `cmd`-to-`cmd` dependency by decoupling `humility-cmd-flash` from `humility-cmd-auxflash`.

`humility-auxflash` is still unfortunately CLI-flavored (doing its own logging and progress bars), but this is a step in the right direction.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ default-members = ["humility-bin"]
 members = [
     "humility-arch-arm",
     "humility-arch-cortex",
+    "humility-auxflash",
     "humility-bin",
     "humility-cmd",
     "humility-cli",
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -77,7 +78,7 @@ members = [
     "cmd/validate",
     "cmd/vpd",
     "cmd/writeword",
-    "xtask",
+    "xtask", "humility-auxflash",
```
labbott: Keep this in alphabetical order?

mkeeter: I have been betrayed by `cargo new`; fixed!

PR #630, opened 2026-04-23
Move `get_probe_serial` to method on `Cli`

mkeeter:
This removes a `cmd-gdb` → `cmd-openocd` dependency

### DIFF
--- a/humility-cli/src/lib.rs
+++ b/humility-cli/src/lib.rs
@@ -4,7 +4,7 @@
 
 pub mod env;
 
-use anyhow::{Context, Result};
+use anyhow::{Context, Result, bail};
 use clap::{AppSettings, ArgGroup, ArgMatches, Parser};
 use env::Environment;
 use humility::{core::Core, hubris::HubrisArchive, msg, net, warn};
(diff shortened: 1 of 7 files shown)


humility-cli/src/lib.rs
```diff
@@ -114,6 +114,56 @@ pub struct Cli {
...
+                                "Cannot specify probe serial number with both \
+                                 `--probe` and `--serial`"
+                            );
+                        }
+                    }
+                    Ok(Some(cap.get(3).unwrap().as_str().to_string()))
+                } else {
+                    bail!(
+                        "Cannot specify `--probe {}` with `openocd` subcommand \
+                         (must be of the form vid:pid:serial)",
```
hawkw: since this method is now decoupled from `openocd`, might we want to not say "...with openocd subcommand" here, in case this is later used elsewhere?

hawkw (approved):
this looks good overall, with one error message nit

humility-cli/src/lib.rs
mkeeter: good catch, fixed!

PR #631, opened 2026-04-23
Use more powerful reflection functions when possible

mkeeter:
This PR uses improved reflection APIs in a few place.

The relevant APIs are the following:

- `read_variable` (renamed in this PR, was previously `load_variable`) reads a value from a `Core` then uses `Load::from_value` to get a type `T`
- `reflect::Struct::field` accesses a field by name, to eliminate manual destructuring chains

### DIFF
--- a/cmd/sensors/src/lib.rs
+++ b/cmd/sensors/src/lib.rs
@@ -338,15 +338,12 @@ impl SensorReader for HiffySensorReader<'_> {
 struct RamSensorReader {
     /// Position of `LAST_READING` in global RAM
     last_reading: HubrisVariable,
-    last_reading_buf: Vec<u8>,
 
     /// Position of `DATA_VALUE` in global RAM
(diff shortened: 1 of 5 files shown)


cmd/sensors/src/lib.rs
```diff
@@ -338,15 +338,12 @@ impl SensorReader for HiffySensorReader<'_> {
 struct RamSensorReader {
     /// Position of `LAST_READING` in global RAM
     last_reading: HubrisVariable,
-    last_reading_buf: Vec<u8>,
```
mkeeter, line 341: This PR has more allocation churn, because `read_variable` creates an ad-hoc buffer to read into each time.  I think that's fine; we're on the Big Computer, and a sufficiently smart allocator will just give us the same allocation each time.

labbott (approved):

PR #633, opened 2026-04-27
Switch to using hubtools for some hubris archive manipulation

labbott:
When humility/hubris were first developed we didn't have many details of the hubris archive beyond "it's a zip". We've since developed a separate library for hubris manipulation. Let humility use it too.


humility-core/src/hubris.rs
```diff
@@ -638,14 +630,6 @@ pub struct HubrisFlashConfig {
...
 }
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub enum HubrisArchiveDoneness {
-    /// Fully load archive
-    Cook,
-    /// Load archive into memory, but do not otherwise process
-    Raw,
-}
-
```
labbott, line 648: Getting rid of this works fine but it makes the CI tests take longer, probably because they now have to do the full hubris archive load for `extract`. I think I'm going to bring this back and fix `extract` instead.

humility-core/src/hubris.rs
```diff
@@ -1410,13 +1397,14 @@ impl HubrisArchive {
...
             .into_par_iter()
             .map(|i| -> Result<Option<(usize, String, Vec<u8>)>> {
                 // ZipArchive is cheap to clone since the backing is cheap
-                let mut archive = archive.clone();
-                let mut file = archive.by_index(i)?;
-                let path = Path::new(file.name());
+                let archive = hubris.clone();
+
+                let (name, data) = archive.extract_file_by_index(i)?;
+                let path = Path::new(&name);
```
labbott, line 1404: I wrote some of this code a few years ago but had a cryptic commit message about it being too slow. If I had  to guess it was this code which was too slow (maybe before we even had parallel iteration) but we basically match mainline if we let the archive be cloneable https://github.com/oxidecomputer/hubtools/pull/63

mkeeter: Following up on my comments from the hubtools PR, I don't think the cloning is necessary here:
```suggestion
        let mut objects = (0..hubris.file_count()?)
.into_par_iter()
.map(|i| -> Result<Option<(usize, String, Vec<u8>)>> {
                let (name, data) = hubris.extract_file_by_index(i)?;
                let path = Path::new(&name);
```

humility-core/src/hubris.rs
```diff
@@ -1507,21 +1492,24 @@ impl HubrisArchive {
...
             .into_par_iter()
             .map(|i| {
                 // ZipArchive is cheap to clone since the backing is cheap
-                let mut archive = archive.clone();
-                let mut file = archive.by_index(i)?;
-                let path = Path::new(file.name());
+                let archive = RawHubrisArchive::from_vec(hubris.zip.clone())?;
+
+                // ZipArchive is cheap to clone since the backing is cheap
+                let (name, data) = archive.extract_file_by_index(i)?;
```
mkeeter, line 1509: Similarly, 
```suggestion
                let (name, data) = hubris.extract_file_by_index(i)?;
```

humility-core/src/hubris.rs
labbott: Ahhhh I see now that there were mutable shenanigans going on with the `ZipArchive` but that's not the case here.

mkeeter:
Closing in favor of #657 (which bundles in this commit)

PR #636, opened 2026-04-28
Bump humpty, serde, lzss, zerocopy

mkeeter:
(staged on top of #634)

This PR bumps a bunch of crates which want to be updated together.  Most of the textual changes are updating `zerocopy` derives and APIs.

### DIFF
--- a/humility-core/src/hubris.rs
+++ b/humility-core/src/hubris.rs
@@ -32,7 +32,7 @@ use multimap::MultiMap;
 use num_traits::FromPrimitive;
 use rustc_demangle::demangle;
 use scroll::{IOwrite, Pwrite};
-use zerocopy::{AsBytes, FromBytes};
+use zerocopy::{FromBytes, IntoBytes};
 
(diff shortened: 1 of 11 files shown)


humility-core/src/hubris.rs
```diff
@@ -1850,10 +1850,10 @@ impl HubrisArchive {
                             }
                             OXIDE_NT_HUBRIS_TASK => {
                                 match DumpTask::read_from_prefix(note.desc) {
-                                    Some(task) => {
+                                    Ok((task, _)) => {
                                         self.task_dump = Some(task);
                                     }
-                                    None => {
+                                    Err(..) => {
```
labbott: Worth printing the error?

labbott (approved):
Minor comment but LGTM thanks for doing this!

humility-core/src/hubris.rs
mkeeter: Sure, although it can only be one thing (`SizeError`)

PR #646, opened 2026-05-01
Update indicatif

mkeeter:
(staged on top of #645)

`ProgressStyle::template` has become fallible, so I sprinkled a bunch of `?` everywhere.


labbott (approved):
This will save some diffs in the probe-rs upgrade which requires a new `indicatif`

PR #648, opened 2026-05-01
Update to clap 4

mkeeter:
(Staged on #646)

This PR updates `clap` to version 4!

- `parse(try_from_str = parse_int::parse)` becomes `value_parser = parse_int::parse::<u32>`
- Conflicts and dependencies now use the original Rust names (i.e. `requires = "force_dump_agent"` instead of `requires = "force-dump-agent"`)
- I couldn't get the external subcommand to work, so I'm just accumulating `cmd: Vec<String>` in the tail position of the arguments
- `ArgEnum` becomes `ValueEnum`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -155,17 +155,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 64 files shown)


labbott (approved):
Broadly speaking this looks fine. I _think_ we still need the explicit `parse` everywhere to handle hex vs not?

PR #650, opened 2026-05-01
Remove `humility repl` for now

mkeeter:
(staged on #649)

This PR removes `humility repl`, which is not widely used (and isn't any faster than iterating at the CLI, because it reloads the archive on each call).


labbott (approved):
Thanks `repl` for your work

PR #651, opened 2026-05-01
Switch back to upstream capstone

mkeeter:
(staged on #650)

This PR switches us to upstream `capstone-rs`, because https://github.com/capstone-rust/capstone-rs/issues/84 has been fixed!

PR #652, opened 2026-05-04
Revamp how attaching works

mkeeter:
(Staged on #651)

**Note:** A bunch of the LOC changes are in the test suite, so you may want to just look at the first commit when reviewing.  The logic changes are a significant net code reduction (+616/-1144), but a few of the test suite commands now successfully run where they previously exited with an error (+1147/-358), which accounts for the overall LOC increase.

## Overview
The `CommandKind` type has confusing, non-orthogonal behavior.

It combines two ideas:

- Do we need a `HubrisArchive`, and if so, does it need to be fully loaded ("cooked") or do we just need the raw ZIP data?
- Do we need to attach to a system (called a "core")?
  - If so, what kind of attachments are valid?
    - Live system (debugger)
    - Live system (network)
    - Dump (RAM)
    - Archive (flash only)
  - How should we validate any attached system?

These two big ideas ("do we need an archive" and "do we need a core") are combined in unintuitive ways.  For example, `CommandKind::Unattached { archive: Archive::Required }` tries to attach to both a dump and an archive, but does not fail if the archive is absent (see https://github.com/oxidecomputer/humility/pull/564#discussion_r2316200692).

It's also unfortunate that it's defined at the `Command` level, because commands may not have consistent behavior across all of their subcommands.  For example, `humility hiffy -l` just lists Hiffy APIs from an archive, and does not require attaching to a system.  Indeed, if you run `humility hiffy -l` with an archive **and** you have a probe attached to an SP running a different image, it will fail:
```console
$ humility --archive=build-cosmo-b-dev-image-default.zip hiffy -l
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.15s
     Running `target/debug/humility hiffy -l`
humility hiffy failed: image ID in archive ([bd, 1b, fb, 72, 35, 4b, a7, 48]) does not equal ID at 0x8004f50 ([7c, 56, 0, 8, 38, 54, 0, 8])
# this is an attached Grapefruit, why is it interfering with hiffy -l?
```

Some commands work around this by doing an extra parse of the arguments in their `init` function, then return a different `Command` depending on arguments:
https://github.com/oxidecomputer/humility/blob/27d1f1549e58a54538657ab5359be9a5ae856de1/cmd/monorail/src/lib.rs#L1186-L1221

This works, but is annoying to have to do!

I have one more complaint, which is improved but not entirely fixed in this PR:  the `HubrisArchive` is default-constructed, then _actually_ populated based on `archive` or `dump` arguments; if neither is present, then archive is not valid!  Commands have to work around this (with varying degrees of incredulity), e.g.

https://github.com/oxidecomputer/humility/blob/27d1f1549e58a54538657ab5359be9a5ae856de1/cmd/reset/src/lib.rs#L33-L37

## So, what's to be done?

This PR removes `CommandKind` entirely.  It is replaced with a set of functions on `Cli` to build archives and cores, which can be called by commands when they are needed (instead of before the command is invoked).

### Archive handling
- `archive` returns the archive (if available); it **does not** ever return a default-constructed archive
- `try_archive` returns an optional archive
- `raw_archive` returns the bytes of the ZIP file

All of these return a full `HubrisArchive`; it is the caller's responsibility to store it somewhere and hand out references.

### Attachment
- `attach_live` attaches to a live system (debugger or `net`-based)
- `attach_live_or_dump` attaches to a live system or dump
- `attach_archive` attaches to an archive

There are also additional small wrapper functions with suffixes for validation, e.g. `attach_live_booted` checks that the live system has booted.  Each of these functions returns a `Box<dyn Core>` directly, instead of stashing it in the `ExecutionContext`.

## Changes to commands
Each command is updated to construct its archive and/or core on-demand.  It's mostly mechanical porting, e.g. 
```rust
CommandKind::Attached {
    archive: Archive::Required,
    attach: Attach::LiveOnly,
    validate: Validate::Booted,
}
```
becomes
```rust
    let hubris = &context.cli.archive()?;
    let core = &mut *context.cli.attach_live_booted(hubris)?;
```

...but there are a few weird edge cases – for example, `humility probe` now uses `attach_probe` directly.


cmd/dump/src/lib.rs
```diff
@@ -855,10 +855,9 @@ fn dump_agent_status(
 }
 
 fn dumpcmd(context: &mut ExecutionContext) -> Result<()> {
-    let core = &mut **context.core.as_mut().unwrap();
-    let hubris = context.archive.as_ref().unwrap();
-
     let subargs = DumpArgs::try_parse_from(&context.cli.cmd)?;
+    let hubris = &context.cli.archive()?;
+    let core = &mut *context.cli.attach_live_match(hubris)?;
```
mkeeter, line 860: Do we care about the ability to dump from a dump?  Probably not, but it was previously possible.

cmd/hiffy/src/lib.rs
```diff
@@ -233,6 +231,7 @@ fn hiffy(context: &mut ExecutionContext) -> Result<()> {
     // creation failure.  (Note that -L will still create the HiffyContext,
     // even if run on a dump.)
     //
+    let core = &mut *context.cli.attach_live_or_dump_booted(hubris)?;
```
mkeeter: Is there a reason to support dumps here?

cmd/host/src/lib.rs
```diff
@@ -468,8 +468,8 @@ fn host_last_post_code(
 
 fn host(context: &mut ExecutionContext) -> Result<()> {
     let subargs = HostArgs::try_parse_from(&context.cli.cmd)?;
-    let hubris = context.archive.as_ref().unwrap();
-    let core = &mut **context.core.as_mut().unwrap();
+    let hubris = &context.cli.archive()?;
+    let core = &mut *context.cli.attach_live_or_dump_match(hubris)?;
```
mkeeter: `Cosmo` subcommands all call hiffy and can't be used on a dump.

cmd/dump/src/lib.rs
labbott: I feel like if we actually want that we're probably going to want other tooling

cmd/hiffy/src/lib.rs
labbott: Probably not

cmd/map/src/lib.rs
```diff
@@ -64,20 +64,18 @@
...
-    let core = &mut **context.core.as_mut().unwrap();
-    core.op_start()?;
-
-    let hubris = context.archive.as_ref().unwrap();
+    let hubris = &context.cli.archive()?;
 
+    // Use an archive core to easily read from flash
+    let core = &mut *humility::core::attach_archive(hubris)?;
     let regions = hubris.regions(core)?;
-    core.op_done()?;
```
labbott, line 80: I guess these were vestigial `op_start`/`op_done`?

cmd/stmsecure/src/lib.rs
```diff
@@ -295,9 +295,9 @@ fn stmsecure_swapbanks(core: &mut dyn Core) -> Result<()> {
 
 #[rustfmt::skip::macros(format)]
 fn stmsecure(context: &mut ExecutionContext) -> Result<()> {
-    let core = &mut **context.core.as_mut().unwrap();
-
     let subargs = StmSecureArgs::try_parse_from(&context.cli.cmd)?;
+    let hubris = context.cli.try_archive()?;
```
labbott: We never do anything with hubris so this could probably be  removed

labbott (commented):
I think this looks okay I want to take another look in a bit

cmd/map/src/lib.rs
mkeeter: Yeah, and even more vestigal now that we're reading from the archive directly (which doesn't have a core to start and stop).

cmd/hiffy/src/lib.rs
mkeeter: Fixed – we previously needed to build a `HiffyContext` to list functions (which needed a `Core`), but now we can do it from just the `HubrisArchive`.

cmd/stmsecure/src/lib.rs
mkeeter: Tracing through the code, this uses the `chip` from the archive (and nothing else), so I'm fine removing it.

labbott (approved):
LGTM

PR #654, opened 2026-05-05
Fix `udprpc` buffer copy

### LINKED ISSUE #653: `humility monorail status` panic
`humility monorail status` panics when pointed at sidecar reverso with the following error:

```
$ ./humility --ip ... --archive ... monorail status
humility: connecting to XXX%YYY

thread 'main' (1) panicked at humility-hiffy/src/lib.rs:701:38:
range end index 12 out of range for slice of length 11
...
$
```

`humility-hiffy/src/lib.rs:701:38` - https://github.com/oxidecomputer/humility/blob/ebaeedd2b772fe5b97641ed373063a5ca72bd666/humility-hiffy/src/lib.rs#L701

- built `humility` from source at `ebaeedd2b772fe5b97641ed373063a5ca72bd666`
- running _reverso_ sidecar SP at `mkeeter/disable-uplink-connection` - `9b0a4a5460fe385b976ad96a6969f60473fc9066` - https://github.com/oxidecomputer/hubris/commits/mkeeter/disable-uplink-connection/

mkeeter:
In https://github.com/oxidecomputer/humility/pull/608, we changed the `udprpc` backend to truncate the returned buffer based on the actual number of bytes received.  However, we still tried to copy `nreply` bytes out of that buffer.  This fails if the reply type has been serialized (e.g. with Hubpack) and is shorter than the maximal `nreply`.

This PR changes the implementation to extract and copy the (variable-size) reply portion of the buffer, instead of a fixed `nbytes`-sized chunk.

Fixes #653, tested on `niles`

### DIFF
--- a/humility-hiffy/src/lib.rs
+++ b/humility-hiffy/src/lib.rs
@@ -697,9 +697,13 @@ impl<'a> HiffyContext<'a> {
             if code != 0 {
                 return Err(Failure::FunctionError(code));
             }
-            rval[0..nreply as usize]
-                .copy_from_slice(&buf[5..(5 + nreply as usize)]);
+            // The reply may be shorter than `nreply` bytes, if this is an Idol
+            // call that uses serialization for its return type.  `buf` has
+            // already been trimmed based on actual reply length.
+            let reply = &buf[5..];
+            rval[0..][..reply.len()].copy_from_slice(reply);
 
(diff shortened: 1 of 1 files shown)


labbott (approved):
I thought I hit approve yesterday but maybe it got lost?


I think I convinced myself this is correct so LGTM

PR #658, opened 2026-05-06
Make manifest fields non-optional

mkeeter:
(staged on top of #657)

Now that #657 makes it impossible to construct a partially-initialized `HubrisArchive`, there are many fields which are guaranteed to be `Some(..)`.  This PR makes them non-optional, then fixes downstream users.

### DIFF
--- a/humility-net-core/src/lib.rs
+++ b/humility-net-core/src/lib.rs
@@ -96,12 +96,7 @@ impl NetCore {
             hiffy_socket,
             flash: HubrisFlashMap::new(hubris)?,
             ram: None, // filled in below
-            imageid: hubris
-                .imageid
-                .as_ref()
-                .ok_or_else(|| anyhow!("missing image ID"))?
-                .1
-                .clone(),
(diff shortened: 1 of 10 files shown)


humility-net-core/src/lib.rs
```diff
@@ -96,12 +96,7 @@ impl NetCore {
             hiffy_socket,
             flash: HubrisFlashMap::new(hubris)?,
             ram: None, // filled in below
-            imageid: hubris
-                .imageid
-                .as_ref()
-                .ok_or_else(|| anyhow!("missing image ID"))?
-                .1
-                .clone(),
+            imageid: hubris.imageid.1.clone(),
```
labbott: Can this use `hubris.image_id()` to avoid direct field access? (Same comment a few places)

mkeeter: Good call, I've made the field non-`pub` so that everyone has to call `image_id()`.

labbott (approved):

Issue #665, opened 2026-05-13
want humility to be able to speak in refdes


hawkw:
Per a conversation with @adamlouis earlier, manufacturing would really like it if `humility` could use reference designators ("refdes") when referring to components on a board. These designators are typically used to refer to a component when additional testing or rework is necessary, and the mfg folks are currently having to do some unpleasant things like maintaining their own tables of I2C addr/mux pairs to refdes, which I wish they didn't have to do. 

Hubris and its build system already know the refdes of most components, since they are used to refer to them in the IPCC and MGS inventories. The app.toml has a field for annotating a component with its refdes. So, `humility` ought to be using this information as well.

I've started out by making the code for parsing a Hubris config from an archive or core also look at the refdes fields in the TOML, and print them in `humility manifest`. Once that's done, I think we will want to make at least the `humility validate` command and the open-pin check also include refdes in their output. There are probably other places where we should include this as well (`sensors`? `pmbus`? anything else?)

labbott:
Should be fixed by https://github.com/oxidecomputer/humility/pull/667

PR #667, opened 2026-05-14
parse refdes from Hubris manifest

hawkw:
In service of #665, this commit changes Humility to actually parse the refdes field from the Hubris manifest, and print it in `humility manifest` and in `humility manifest`. Subsequent changes changes will add refdeses in other places, such as `humility rendmp`, but I wanted to at least put a pin in this bit for now.


adamlouis:
cc @chris-wiet @Lydia-Sylla

humility-core/src/hubris.rs
```diff
@@ -933,6 +970,7 @@ impl HubrisArchive {
                             device.device.clone(),
                             i,
                         ),
+                        refdes: None, // TODO ELIZA FIXME LOL
```
mkeeter: Do we need to fix this in this PR?

labbott (commented):
I don't want to approve this yet but I support this 👍

humility-core/src/hubris.rs
hawkw: after going back and looking at the code, it appears that we don't currently have refdes for "other sensors" sensors (i.e. the ones that aren't coming from i2c device configs), so it's correct for this to be `None` for the time being:
https://github.com/oxidecomputer/hubris/blob/358159e781a1f9fa6a8e33b6a992c215622c1120/task/sensor-api/build.rs#L24-L31

i will edit the comment to not say what it currently says.

i will note that those sensors do have a `name` field and we should probably update the manifest to also put refdes on them, even if it's just for humility's use for now...

hawkw:
[CI failure](https://github.com/oxidecomputer/humility/actions/runs/26477558815/job/77966892233?pr=667) is just because I apparently forgot to update some of the expectorate output, fixing it now!

humility-core/src/hubris.rs
```diff
@@ -160,9 +161,11 @@ impl HubrisManifest {
         let mut sensors = vec![];
         let mut i2c_devices = vec![];
         let mut i2c_buses = HubrisI2cBusList(vec![]);
+        let mut fmt_meta = ManifestFormatMetadata { max_refdes_len: 0 };
         if let Some(ref config) = config.config {
             if let Some(i2c) = config.i2c.as_ref() {
-                let cfg = HubrisManifestI2cConfig::from_config(i2c)?;
+                let cfg =
+                    HubrisManifestI2cConfig::from_config(i2c, &mut fmt_meta)?;
```
mkeeter: Vibes: I don't like having `from_config` both build the object and mutate a `fmt_meta` reference.

How about setting `max_refdes_len` afterwards, e.g.
```rust
fmt_meta.max_refdes_len = i2c_devices
    .iter()
    .flat_map(|m| &m.refdes)
    .map(|m| m.len())
    .max()
    .unwrap_or(0);
```

cmd/validate/src/lib.rs
```diff
@@ -98,9 +98,18 @@ struct ValidateArgs {
...
+    let refdes_len = hubris
+        .manifest
+        .fmt_meta
+        .max_refdes_len
+        .max(REFDES_HDR.len())
+        .max(NO_REFDES.len());
     println!(
-        "{:2} {:>2} {:2} {:3} {:4} {:13} DESCRIPTION",
+        "{:2} {REFDES_HDR:refdes_len$} {:>2} {:2} {:3} {:4} {:13} DESCRIPTION",
         "ID", "C", "P", "MUX", "ADDR", "DEVICE"
```
mkeeter: Vibes: I don't like mixing inline and out-of-line(?) arguments, because it makes the order harder to figure out at a glance.
```suggestion
        "{:2} {:refdes_len$} {:>2} {:2} {:3} {:4} {:13} DESCRIPTION",
"ID", REFDES_HDR, "C", "P", "MUX", "ADDR", "DEVICE"
```
(also applies elsewhere in this file)

mkeeter (approved):

humility-core/src/hubris.rs
hawkw: I wanted to avoid iterating twice, but I will admit that's maybe a premature optimization when there's usually on the order of 100ish items to iterate... I also thought about having `from_config` return `(HubrisManifestI2cConfig, usize)` but I didn't like the vibes of having it return a random `usize`. I suppose it could also be a field on `HubrisManifestI2cConfig` rather than on `HubrisManifest`. Maybe that's the best option...?

hawkw: (part of why I made it its own struct was that I was kind of imagining we might have max length fields for other things so that the table layout can be made less janky later)

hawkw: Oh, upon looking again, I guess that because `HubrisManifestI2cConfig` is a module-private struct, I feel okay about it just including a `max_refdes_len` which is used to populate a `HubrisManifest` field later.

hawkw: I think you'll prefer 9ace2d2b3b9ef3a5848a33da016be9b6997f7200

cmd/validate/src/lib.rs
hawkw: done in fd00026eeab60835c2b8fe991e78c74eec615847

PR #668, opened 2026-05-15
Make IP address validation lazier

### LINKED ISSUE #609: `--ip` causes early failures even when it doesn't matter
```
$ humility --ip 'fe80::0c1d:8cff:fec0:e207%en7'` --archive=~/oxide/hubris/target/grapefruit-standalone/dist/default/build-grapefruit-standalone-image-default.zip manifest
error: Invalid value "fe80::0c1d:8cff:fec0:e207%en7" for '--ip <IP>': Could not find interface for en7
```

The IP address isn't needed for `humility manifest`, but we parse it proactively when building the `Cli` object (either from the CLI or as an environment variable), and bail if we can't find the interface.

mkeeter:
(Staged on #666)

This PR makes validation of the `--ip` argument lazier: we now store a `Result<ScopedV6Addr, ScopedV6AddrError>`, and only require it to be valid when it's actually used.

Fixes #609

### DIFF
--- a/cmd/gimlet/src/lib.rs
+++ b/cmd/gimlet/src/lib.rs
@@ -105,6 +105,7 @@ fn run(context: &mut ExecutionContext) -> Result<()> {
     let ip = context
         .cli
         .ip
+        .clone()
         // TODO(eliza): I wonder if there's some way to make `clap` enforce that
         // an optional global argument is required, so that it prints a prettier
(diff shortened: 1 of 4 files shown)


labbott (approved):
LGTM, you may want reviews from those who might have strong opinions about `thiserror` shaped things

PR #670, opened 2026-05-19
Introduce `humility-vpd-lib`

labbott:
It's useful to be able to call VPD functions from Rust code

### DIFF
--- a/cmd/vpd/Cargo.toml
+++ b/cmd/vpd/Cargo.toml
@@ -18,8 +18,7 @@ indicatif.workspace = true
 serde.workspace = true
 ron.workspace = true
 
+humility-vpd-lib.workspace = true
 humility-cli.workspace = true
 humility-hexdump.workspace = true
-humility-hiffy.workspace = true
-humility-idol.workspace = true
 humility.workspace = true
--- a/cmd/vpd/src/lib.rs
+++ b/cmd/vpd/src/lib.rs
@@ -84,9 +84,6 @@
 //! humility: successfully wrote 56 bytes of VPD
 //! ```
 //!
-//! You can also use a file as a loopback device via `--loopback`, allowing
-//! you to, e.g., read binary data and format it (i.e., via `--read`).
-//!
 //! To lock a VPD device, use the `--lock` command.  This will lock the VPD
 //! permanently and cannot be undone; subsequent attempts to write to (or
 //! lock) a locked VPD device will result in an error.  The lock status of
@@ -98,22 +95,21 @@
 //!   already locked
 //! - If a device is missing or returns an error when its lock status is
 //!   initially checked, the command returns a *non-zero* exit status without
-//!   locking other devices.  Use `--allow-missing` to disable this behavior.
+//!   locking other devices.
 //! - If all devices are present but some devices fail to lock, other devices
 //!   will be locked and the command will return with a *non-zero* exit status.
 
-use anyhow::{Context, Result, bail};
+use anyhow::{Result, bail};
 use clap::{ArgGroup, Parser};
-use hif::*;
 use humility::core::Core;
 use humility::hubris::*;
 use humility_cli::{ExecutionContext, humility_cmd};
 use humility_hexdump::Dumper;
-use humility_hiffy::*;
-use humility_idol::{self as idol, HubrisIdol};
-use indicatif::{ProgressBar, ProgressStyle};
+use humility_vpd_lib::VpdTarget;
 use std::fs;
-use std::io::{Read, Seek, SeekFrom, Write};
+use std::io::Write;
+use std::path::PathBuf;
+use std::time::Duration;
 
 #[derive(Parser, Debug)]
 #[clap(
@@ -144,7 +140,7 @@ pub struct VpdArgs {
 
     /// write the contents of the specified file into the designated VPD
     #[clap(long, short, value_name = "filename", group = "command")]
-    write: Option<String>,
+    write: Option<PathBuf>,
 
     /// read the contents of the designated VPD (or of all with --list)
     #[clap(long, short, group = "command")]
@@ -183,11 +179,6 @@ pub struct VpdArgs {
     allow_missing: bool,
 }
 
-enum VpdTarget {
-    Device(usize),
-    Loopback(fs::File),
-}
-
 fn vpd_devices(
     hubris: &HubrisArchive,
 ) -> impl Iterator<Item = &HubrisI2cDevice> {
@@ -201,87 +192,45 @@ fn vpd_devices(
 fn list(
     hubris: &HubrisArchive,
     core: &mut dyn Core,
-    subargs: &VpdArgs,
+    timeout: Duration,
+    read: bool,
 ) -> Result<()> {
-    let devices = vpd_devices(hubris).collect::<Vec<_>>();
-    let timeout = std::time::Duration::from_millis(subargs.timeout);
-    let mut context = HiffyContext::new(hubris, core, timeout)?;
-    let read_op = hubris.get_idol_command("Vpd.read")?;
-
-    let locked_op = hubris.get_idol_command("Vpd.is_locked").ok();
-    let results = if let Some(locked_op) = &locked_op {
-        let mut ops = vec![];
-
-        for ndx in 0..devices.len() {
-            let payload = locked_op.payload(&[(
-                "index",
-                idol::IdolArgument::Scalar(ndx as u64),
-            )])?;
-
-            context.idol_call_ops(locked_op, &payload, &mut ops)?;
-        }
-
-        ops.push(Op::Done);
-
-        Some(context.run(core, ops.as_slice(), None)?)
-    } else {
-        println!(
-            "Note: firmware does not support the Vpd.is_locked operation."
-        );
-        println!("Lock status will be missing from the table below.");
-        None
-    };
+    let results = humility_vpd_lib::vpd_list(hubris, core, timeout, read)?;
 
     println!(
         "{:2} {:>2} {:2} {:3} {:4} {:13} {:25} LOCKED",
         "ID", "C", "P", "MUX", "ADDR", "DEVICE", "DESCRIPTION",
     );
 
-    for (ndx, device) in devices.iter().enumerate() {
-        use humility::reflect::Base::Bool;
-        use humility::reflect::Value::Base;
-
-        let mux = match (device.mux, device.segment) {
+    for device in results {
+        let mux = match (device.device.mux, device.device.segment) {
             (Some(m), Some(s)) => format!("{}:{}", m, s),
             (None, None) => "-".to_string(),
             (_, _) => "?:?".to_string(),
         };
 
-        let result = if let (Some(lop), Some(rs)) = (&locked_op, &results) {
-            Some(lop.decode(&rs[ndx]))
-        } else {
-            None
+        let locked = match device.locked {
+            Ok(true) => "locked".to_string(),
+            Ok(false) => "unlocked".to_string(),
+            Err(s) => format!("<{s}>"),
         };
 
         println!(
             "{:2} {:2} {:2} {:3} 0x{:02x} {:13} {:25} {}",
-            ndx,
-            device.controller,
-            device.port.name,
+            device.ndx,
+            device.device.controller,
+            device.device.port.name,
             mux,
-            device.address,
-            device.device,
-            device.description,
-            match result {
-                Some(Ok(Base(Bool(val)))) => match val {
-                    false => "unlocked",
-                    true => "locked",
-                }
-                .to_string(),
-                Some(Ok(r)) => format!("<{r:?}>").to_string(),
-                Some(Err(err)) => format!("<{}>", err).to_string(),
-                None => "(too old)".to_string(),
-            },
+            device.device.address,
+            device.device.device,
+            device.device.description,
+            locked
         );
 
-        if subargs.read {
-            let mut target = VpdTarget::Device(ndx);
-
-            let rval = vpd_slurp(core, &mut context, &read_op, &mut target);
-
+        if let humility_vpd_lib::VpdData::Data(d) = device.data {
             print!(" |\n +--> ");
 
-            match rval {
+            match d {
                 Ok(vpd) => {
                     match tlvc::TlvcReader::begin(&vpd[..]) {
                         Ok(reader) => {
@@ -340,196 +289,27 @@ fn target(hubris: &HubrisArchive, subargs: &VpdArgs) -> Result<VpdTarget> {
         } else {
             bail!("device index {} invalid; --list to list", id)
         }
-    } else if let Some(loopback) = &subargs.loopback {
-        if subargs.write.is_some() {
-            Ok(VpdTarget::Loopback(fs::File::create(loopback)?))
-        } else {
-            Ok(VpdTarget::Loopback(fs::File::open(loopback)?))
-        }
+    } else if subargs.loopback.is_some() {
+        bail!("loopback support is deprecated");
     } else {
         bail!("must specify either device ID or device description");
     }
 }
 
-fn vpd_write(
-    hubris: &HubrisArchive,
-    core: &mut dyn Core,
-    subargs: &VpdArgs,
-) -> Result<()> {
-    let timeout = std::time::Duration::from_millis(subargs.timeout);
-    let mut context = HiffyContext::new(hubris, core, timeout)?;
-    let op = hubris.get_idol_command("Vpd.write")?;
-    let target = target(hubris, subargs)?;
-
-    let bytes = if let Some(ref filename) = subargs.write {
-        let file = fs::File::open(filename)?;
-
-        let p = tlvc_text::load(file).with_context(|| {
-            format!("failed to parse {} as VPD input", filename)
-        })?;
-
-        tlvc_text::pack(&p)
-    } else {
-        vec![0xffu8; 1024]
-    };
-
-    let target = match target {
-        VpdTarget::Device(target) => target,
-        VpdTarget::Loopback(mut file) => {
-            file.write_all(&bytes)?;
-            return Ok(());
-        }
-    };
-
-    let mut all_ops = vec![];
-
-    for (offset, b) in bytes.iter().enumerate() {
-        let mut ops = vec![];
-        let payload = op.payload(&[
-            ("index", idol::IdolArgument::Scalar(target as u64)),
-            ("offset", idol::IdolArgument::Scalar(offset as u64)),
-            ("contents", idol::IdolArgument::Scalar(*b as u64)),
-        ])?;
-
-        context.idol_call_ops(&op, &payload, &mut ops)?;
-        all_ops.push(ops);
-    }
-
-    let nops = (context.text_size() / context.ops_size(&all_ops[0])?) - 1;
-
-    if nops == 0 {
-        bail!("text size is too small for a single write!");
-    }
-
-    let mut offset = 0;
-
-    let bar = ProgressBar::new(bytes.len() as u64);
-
-    bar.set_style(ProgressStyle::default_bar().template(if subargs.erase {
-        "humility: erasing VPD [{bar:30}] {bytes}/{total_bytes}"
-    } else {
-        "humility: writing VPD [{bar:30}] {bytes}/{total_bytes}"
-    })?);
-
-    for chunk in all_ops.chunks(nops) {
-        let mut ops = chunk.iter().flatten().copied().collect::<Vec<Op>>();
-        ops.push(Op::Done);
-
-        let results = context.run(core, ops.as_slice(), None)?;
-
-        for (o, result) in results.iter().enumerate() {
-            op.decode::<()>(result).with_context(|| {
-                format!("failed to write VPD at offset {}", offset + o)
-            })?;
-        }
-
-        offset += results.len();
-
-        bar.set_position(offset as u64);
-    }
-
-    bar.finish_and_clear();
-
-    if subargs.erase {
-        humility::msg!("successfully erased VPD");
-    } else {
-        humility::msg!("successfully wrote {offset} bytes of VPD");
-    }
-
-    Ok(())
-}
-
-fn vpd_read_at(
-    core: &mut dyn Core,
-    context: &mut HiffyContext,
-    op: &idol::IdolOperation,
-    target: &mut VpdTarget,
-    offset: usize,
-) -> Result<Vec<u8>> {
-    let target = match target {
-        VpdTarget::Device(target) => *target,
-        VpdTarget::Loopback(file) => {
-            let mut buffer = vec![];
-            file.seek(SeekFrom::Start(offset as u64))?;
-            file.read_to_end(&mut buffer)?;
-            return Ok(buffer);
-        }
-    };
-
-    let payload = op.payload(&[
-        ("index", idol::IdolArgument::Scalar(target as u64)),
-        ("offset", idol::IdolArgument::Scalar(offset as u64)),
-    ])?;
-
-    let mut ops = vec![];
-
-    context.idol_call_ops(op, &payload, &mut ops)?;
-    ops.push(Op::Done);
-
-    let results = context.run(core, ops.as_slice(), None)?;
-
-    op.decode::<Vec<u8>>(&results[0])
-        .with_context(|| format!("failed to read at offset {offset}"))
-}
-
-fn vpd_slurp(
-    core: &mut dyn Core,
-    context: &mut HiffyContext,
-    op: &idol::IdolOperation,
-    target: &mut VpdTarget,
-) -> Result<Vec<u8>> {
-    //
-    // First, read in enough to read just the header.
-    //
-    let mut vpd = vpd_read_at(core, context, op, target, 0)?;
-
-    let reader = match tlvc::TlvcReader::begin(&vpd[..]) {
-        Ok(reader) => reader,
-        Err(err) => bail!("{:?}", err),
-    };
-
-    //
-    // If this isn't a header, see if it's all 0xff -- in which case we
-    // will suggest that the part is unprogrammed.
-    //
-    let header = match reader.read_header() {
-        Ok(header) => header,
-        Err(err) => match vpd.iter().find(|&b| *b != 0xffu8) {
-            Some(_) => {
-                bail!("bad header: {:x?}", err);
-            }
-            None => {
-                bail!("VPD appears to be unprogrammed");
-            }
-        },
-    };
-
-    //
-    // And now go back and read everything.
-    //
-    let total = header.total_len_in_bytes();
-
-    while vpd.len() < total {
-        vpd.extend(
-            vpd_read_at(core, context, op, target, vpd.len())
-                .with_context(|| format!("failed to read {total} bytes"))?,
-        );
-    }
-
-    Ok(vpd[..total].to_vec())
+enum OutputOption {
+    Tlvc,
+    Raw,
+    Binary(String),
 }
 
 fn vpd_read(
     hubris: &HubrisArchive,
     core: &mut dyn Core,
-    subargs: &VpdArgs,
+    target: VpdTarget,
+    timeout: Duration,
+    output: OutputOption,
 ) -> Result<()> {
-    let timeout = std::time::Duration::from_millis(subargs.timeout);
-    let mut context = HiffyContext::new(hubris, core, timeout)?;
-    let op = hubris.get_idol_command("Vpd.read")?;
-    let mut target = target(hubris, subargs)?;
-
-    let vpd = vpd_slurp(core, &mut context, &op, &mut target)?;
+    let vpd = humility_vpd_lib::vpd_read(hubris, core, target, timeout)?;
 
     //
     // Now we should have the whole thing!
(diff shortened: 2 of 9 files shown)


humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+use std::path::PathBuf;
+use thiserror::Error;
+use tlvc::TlvcReadError;
+
+#[derive(Debug, Error)]
+pub enum VpdError {
+    #[error("hiffy: {0}")]
+    Hiffy(String),
+    #[error("idol: {0}")]
+    Idol(String),
```
labbott: I both really like just taking the `anyhow` error as a string and also think it actually produces terrible output sooooo

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    Loopback,
+    #[error("No VPD entries found")]
+    NoVpd,
+    #[error("Errors in VPD entry prevent locking")]
+    VpdEntry,
+}
+
+pub enum VpdTarget {
+    Device(usize),
+    Loopback(fs::File),
```
labbott: I'm waffling if we should just remove the loopback testing code

cmd/vpd/Cargo.toml
```diff
@@ -18,6 +18,7 @@ indicatif.workspace = true
 serde.workspace = true
 ron.workspace = true
 
+humility-vpd-lib.workspace = true
```
mkeeter, line 21: Can you remove any dependencies (e.g. `humility-hiffy` and `humility-idol`) now that the hard work has moved to the library?

humility-vpd-lib/examples/list.rs
```diff
@@ -0,0 +1,34 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use humility::hubris::{HubrisArchive, RawHubrisArchive};
+
+fn main() {
+    let hubris = std::env::var("HUMILITY_ARCHIVE").unwrap();
+    let probe = std::env::var("HUMILITY_PROBE").unwrap();
+
+    let hubris = RawHubrisArchive::load(&hubris).unwrap();
+    let hubris = HubrisArchive::load(hubris, None).unwrap();
```
mkeeter: This suggests that we're missing a `HubrisArchive::from_archive(path: FilePath)`.  How about adding that function, then removing the `pub` export of `RawHubrisArchive`?

humility-vpd-lib/examples/list.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    let core = &mut *humility_probes_core::attach_to_chip(
+        &probe,
+        Some("STM32H753ZITx"),
```
mkeeter: Can we get the chip from the manifest?

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+use humility_idol::{HubrisIdol, IdolArgument};
+use indicatif::{ProgressBar, ProgressStyle};
+use std::fs;
+use std::io::Write;
+use std::io::{Read, Seek, SeekFrom};
+use std::path::PathBuf;
+use thiserror::Error;
+use tlvc::TlvcReadError;
+
+#[derive(Debug, Error)]
```
mkeeter: Nit: I normally do `derive(Debug, thiserror::Error)` to avoid confusion with the `std::error::Error` trait

humility-vpd-lib/src/lib.rs
mkeeter: If this produces terrible output, what about the more canonical style?
```rust
#[error("idol error")]
Idol(#[source] anyhow::Error),
```

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    Hiffy(String),
+    #[error("idol: {0}")]
+    Idol(String),
+    #[error("I/O error")]
+    Io {
+        #[source]
+        err: std::io::Error,
+    },
+    #[error("tlvc: {0:?}")]
+    Tlvc(TlvcReadError<core::convert::Infallible>),
```
mkeeter, line 40: Opened https://github.com/oxidecomputer/tlvc/issues/11 to make this better

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+}
+
+fn vpd_devices(
+    hubris: &HubrisArchive,
+) -> impl Iterator<Item = &HubrisI2cDevice> {
+    hubris
+        .manifest
+        .i2c_devices
+        .iter()
+        .filter(|device| device.device == "at24csw080")
```
mkeeter: Nit: pull this into a constant?

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+
+/// List all available VPD devices
+pub fn vpd_list(
+    hubris: &HubrisArchive,
+    core: &mut dyn Core,
+    timeout: u32,
+) -> Result<Vec<VpdEntry>, VpdError> {
+    let devices = vpd_devices(hubris).collect::<Vec<_>>();
+    let mut context = HiffyContext::new(hubris, core, timeout)
+        .map_err(|e| VpdError::Hiffy(format!("{e:?}")))?;
```
mkeeter: I'm not sure if you want `:?` or `:#` here.  With context, the difference is multiline versus single line

```
ERROR WITH :?
baz failed

Caused by:
    0: bar failed
    1: oh no

ERROR WITH :#
baz failed: bar failed: oh no
```

([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=546a8f8d96812381e2d352a726134201))

This may be a moot point if you decide to wrap the `anyhow::Error` instead, which I recommended above (because then the caller gets to pick how to format it).

humility-vpd-lib/src/lib.rs
mkeeter: One more advantage of the canonical style: as we switch more functions to return `thiserror`-based errors, it will loudly break (since it will no longer be an `anyhow::Error`), which will poke us to fix it.  If we do string formatting, it's easier for changes to go unnoticed.

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+        .i2c_devices
+        .iter()
+        .filter(|device| device.device == "at24csw080")
+}
+
+/// List all available VPD devices
+pub fn vpd_list(
+    hubris: &HubrisArchive,
+    core: &mut dyn Core,
+    timeout: u32,
```
mkeeter: Make this a `std::time::Duration` so that the API is more obvious?

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    Loopback(fs::File),
+}
+
+/// Represents a single VPD from the hubris archive
+pub struct VpdEntry {
+    /// Index into the vpd devices list. This is expected to be stable
+    /// across hubris releases
+    pub ndx: usize,
+    /// Raw i2c data about the device
+    pub device: HubrisI2cDevice,
```
mkeeter: Vibes: this is a moderately large value type.  When I was poking at `humility validate`-as-a-library, I found it was possible to borrow instead (e.g. add a lifetime to `VpdEntry` and have `pub device: &'a HubrisI2cDevice`); check and see how that goes?

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    if items.is_empty() { Err(VpdError::NoVpd) } else { Ok(items) }
+}
+
+/// Writes the VPD data from the file
+pub fn vpd_write(
+    hubris: &HubrisArchive,
+    core: &mut dyn Core,
+    target: VpdTarget,
+    timeout: u32,
+    write: PathBuf,
```
mkeeter: If you want to make this more flexible, it could be a `&std::path::Path` (which accepts both `Path` and `PathBuf`)

humility-vpd-lib/src/lib.rs
mkeeter: (also true for other APIs in this file)

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    let mut items = vec![];
+
+    let results = context
+        .run(core, ops.as_slice(), None)
+        .map_err(|e| VpdError::Idol(format!("{e:?}")))?;
+
+    for (ndx, device) in devices.into_iter().enumerate() {
+        let locked = context
+            .idol_result::<bool>(&locked_op, &results[ndx])
+            .map_err(|e| format!("{e:?}"));
```
mkeeter: `e:#` if you want a single-line version, which I think you probably do (discussed above)

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    hubris: &HubrisArchive,
+    core: &mut dyn Core,
+    target: VpdTarget,
+    timeout: u32,
+    write: PathBuf,
+) -> Result<(), VpdError> {
+    vpd_erase_write(hubris, core, target, timeout, Some(write))
+}
+
+/// Erases 1k of the VPD (write as `0xff`)
```
mkeeter: This has weird vibes – why 1K?  (also if it's 1024, then 1 KiB would be less ambiguous)

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    let mut ops = vec![];
+    let lock_op = hubris
+        .get_idol_command("Vpd.permanently_lock")
+        .map_err(|e| VpdError::Idol(format!("{e:?}")))?;
+
+    if locking.is_empty() {
+        return Ok(0);
+    }
+
+    if any_missing && !allow_missing {
```
mkeeter: This is a logic change from the older code, which read
```rust
    if locking.is_empty() {
        if locked == devices.len() {
            humility::msg!("all VPDs are already locked");
            return Ok(());
        }

        bail!("no VPDs to lock");
    } else if any_missing && !subargs.allow_missing {
        bail!("some VPDs are missing; use `--allow-missing` to continue")
    }
```

In the new code, if all VPDs are missing this will return `Ok(0)`, rather than bailing with `no VPDs to lock`.

Switch the order of these blocks?  Do we also want to distinguish between the two error cases in the original code?

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+        #[source]
+        err: std::io::Error,
+    },
+    #[error("tlvc: {0:?}")]
+    Tlvc(TlvcReadError<core::convert::Infallible>),
+    #[error("VPD region is unprogrammed")]
+    Unprogrammed,
+    #[error("reflect: {0}")]
+    Reflect(String),
+    #[error("Can't lock a loopback device")]
```
mkeeter: Style: error strings should start with a lowercase character (we can make an exception for "VPD" 😄)

(also applies below)

humility-vpd-lib/src/lib.rs
labbott: I had to go look this up, our eeprom chips are 1K https://github.com/oxidecomputer/hubris/blob/c6dbcaf325b88b9aae51f006986e9bc063bbb6e2/drv/i2c-devices/src/at24csw080.rs#L12-L13 

This deserves better documentation

humility-vpd-lib/src/lib.rs
mkeeter: I was confused why all these functions needed to pass a `&mut VpdTarget`; it's because functions mutate the `fs::File` inside the `Loopback` variant.  That weirdness is another point in favor of just removing it.

humility-vpd-lib/src/lib.rs
labbott: The newer code uses `vpd_list` which returns an error if there are no VPD present so the error gets handled sooner

humility-vpd-lib/src/lib.rs
mkeeter: Or make a single-variant error wrapper thingy?
```rust
#[derive(Debug, thiserror::Error)]
#[error(transparent)]
struct IdolResultError(#[source] anyhow::Error);
```

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    hubris: &HubrisArchive,
+    core: &mut dyn Core,
+    target: VpdTarget,
+    timeout: u32,
+) -> Result<(), VpdError> {
+    vpd_erase_write(hubris, core, target, timeout, None)
+}
+
+/// Permanently lock all VPDs, returns the total count of devices
+/// locked. A return value of 0 indicates that all devices were locked.
```
mkeeter: This has weird vibes, because a return value of "number of locked devices" _also_ means all devices were locked.  I wonder if (from an API perspective) this should be two different functions:

- `fn vpd_lock_all(..) -> Result<(), VpdError>` (either locks all or fails)
- `fn vpd_try_lock_all(..) -> Result<usize, VpdError>` (returns the number of devices locked in this iteration)

Even in that case, I don't love the return type, because a single `usize` is ambiguous; something less ambiguous would be returning
```rust
struct VpdLockCount {
    already_locked: usize,
    newly_locked: usize,
    num_lockable_devices: usize,
}
```

If that were the return type for `vpd_try_lock_all`, then `vpd_lock_all` could call it and then check that `already_locked + newly_locked == num_lockable_devices` when deciding how to return.

humility-vpd-lib/src/lib.rs
mkeeter: (This may be a moot point given my complains about the function signature above)

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+        .map_err(|e| VpdError::Hiffy(format!("{e:?}")))?;
+
+    let op = hubris
+        .get_idol_command("Vpd.permanently_lock")
+        .map_err(|e| VpdError::Idol(format!("{e:?}")))?;
+    // Make sure we can read the VPD
+    vpd_slurp(core, &mut context, hubris, &mut target)?;
+
+    let index = match target {
+        VpdTarget::Device(index) => index,
```
mkeeter: Nit: if it's always invalid to call this function with `VpdTarget::Loopback`, that suggests we should have a more fine-trained type, e.g.
```rust
pub struct VpdTargetDevice(pub usize);

pub enum VpdTarget {
    Device(VpdTargetDevice),
    Loopback(fs::File),
}
```
Then this function could take a `VpdTargetDevice`, making the API harder to misuse.

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+        .idol_call_ops(&op, &payload, &mut ops)
+        .map_err(|e| VpdError::Idol(format!("{e:?}")))?;
+    ops.push(Op::Done);
+
+    let results = context
+        .run(core, ops.as_slice(), None)
+        .map_err(|e| VpdError::Idol(format!("{e:?}")))?;
+
+    context.idol_result::<()>(&op, &results[0]).map_err(|e| {
+        VpdError::Idol(format!("failed to lock {index}: {e:?}"))
```
mkeeter: Vibes: this could be a separate variant instead of baking it into a string right away
```rust
    #[error("failed to lock {index}")]
    LockCallFailed {
        index: usize,
        #[source]
        err: anyhow::Error,
    }
```

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+        .get_idol_command("Vpd.write")
+        .map_err(|e| VpdError::Idol(format!("{e:?}")))?;
+
+    let (bytes, erase) = if let Some(ref filename) = write {
+        let file =
+            fs::File::open(filename).map_err(|err| VpdError::Io { err })?;
+
+        let p = tlvc_text::load(file).map_err(|err| VpdError::Io { err })?;
+
+        (tlvc_text::pack(&p), true)
```
mkeeter: `erase` has the wrong sign here – it should be `false` when reading from a file (and vice versa)

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    if nops == 0 {
+        return Err(VpdError::Idol(
+            "text size is too small for a single write!".to_string(),
+        ));
+    }
+
+    let mut offset = 0;
+
+    // Do we want to make this an argument for optionally not having
+    // a progress bar?
```
mkeeter: We'll eventually want something like `trait HumilityProgress`, but that doesn't need to happen now...

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    let bar = ProgressBar::new(bytes.len() as u64);
+
+    bar.set_style(
+        ProgressStyle::default_bar()
+            .template(if erase {
+                "humility: erasing VPD [{bar:30}] {bytes}/{total_bytes}"
+            } else {
+                "humility: writing VPD [{bar:30}] {bytes}/{total_bytes}"
+            })
+            .map_err(|e| VpdError::Idol(format!("{e:?}")))?,
```
mkeeter: Not the right error type?  (I'm also fine just `unwrap`'ing here, because an error here means that you messed up the progress bar format string)

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+        let mut ops = chunk.iter().flatten().copied().collect::<Vec<Op>>();
+        ops.push(Op::Done);
+
+        let results = context
+            .run(core, ops.as_slice(), None)
+            .map_err(|e| VpdError::Idol(format!("{e:?}")))?;
+
+        for (o, result) in results.iter().enumerate() {
+            context.idol_result::<()>(&op, result).map_err(|e| {
+                VpdError::Idol(format!(
```
mkeeter: Similarly, move the rich error information into a specific variant (instead of formatting)?

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+fn vpd_slurp(
+    core: &mut dyn Core,
+    context: &mut HiffyContext,
+    hubris: &HubrisArchive,
+    target: &mut VpdTarget,
+) -> Result<Vec<u8>, VpdError> {
+    let op = hubris
+        .get_idol_command("Vpd.read")
+        .map_err(|e| VpdError::Idol(format!("{e:?}")))?;
+    //
```
mkeeter: Nit: remove the Bryan Comment Style while you're in here?

labbott:
Open question: we will probably grow more humility-lib crates. Do we want to keep them all at the root for now and make cleanup a problem for Future Us or try some organization now?

humility-vpd-lib/src/lib.rs
labbott: `HiffyContext` still wants a `u32` let me see how painful it would be to clean that up everywhere because I strongly agree with that suggestion

mkeeter: I did it when converting `humility_validate` to a library; it's not too bad (`from_millis` on the front end and `as_millis` on the back end).

mkeeter: Alternatively, we could fix `HiffyContext` 😅

humility-vpd-lib/src/lib.rs
labbott: I don't see manufacturing making use of `--allow-missing` so I think just doing `Result<(), VpdError>` would be cleaner

humility-vpd-lib/src/lib.rs
labbott: I just removed it but kept the `VpdTarget` (maaaybe that should just be a straight `usize`?)

cmd/vpd/Cargo.toml
labbott: yes!

humility-vpd-lib/examples/list.rs
labbott: https://github.com/oxidecomputer/humility/pull/674

humility-vpd-lib/examples/list.rs
labbott: https://github.com/oxidecomputer/humility/pull/675 we can make this one step with an extension trait

humility-vpd-lib/src/lib.rs
labbott: This got removed because I got rid of the loopback feature

humility-vpd-lib/src/lib.rs
labbott: whoops :grimacing:

cmd/vpd/Cargo.toml
labbott: Done!

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+    /// Index into the vpd devices list. This is expected to be stable
+    /// across hubris releases
+    pub ndx: usize,
+    /// Raw i2c data about the device
+    pub device: &'a HubrisI2cDevice,
+    /// Whether the `VpdLock` command has been successfully issued to this
+    /// device.
+    pub locked: Result<bool, String>,
+    /// Reading out the VPD can be very slow and can be omitted
+    pub data: VpdData,
```
labbott, line 109: I don't love this but it takes almost 15 seconds to do a list when reading all the VPD

cmd/vpd/src/lib.rs
```diff
@@ -198,90 +192,48 @@ fn vpd_devices(
         .filter(|device| device.device == "at24csw080")
 }
 
-fn list(
+pub fn list(
```
mkeeter: This doesn't need to be `pub` as far as I can tell

cmd/vpd/src/lib.rs
```diff
@@ -541,188 +324,69 @@ fn vpd_read(
...
+    // These commands don't take a target so check them first
     if subargs.list {
-        list(hubris, core, &subargs)?;
-    } else if subargs.write.is_some() || subargs.erase {
-        vpd_write(hubris, core, &subargs)?;
+        list(hubris, core, timeout, subargs.read)?;
+        return Ok(());
+    } else if subargs.lock_all {
+        if subargs.allow_missing {
+            bail!("--allow-missing is deprecated");
```
mkeeter, line 354: 😎

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,445 @@
...
+/// Represents a single VPD from the hubris archive
+pub struct VpdEntry<'a> {
+    /// Index into the vpd devices list. This is expected to be stable
+    /// across hubris releases
+    pub ndx: usize,
+    /// Raw i2c data about the device
+    pub device: &'a HubrisI2cDevice,
+    /// Whether the `VpdLock` command has been successfully issued to this
+    /// device.
+    pub locked: Result<bool, String>,
```
mkeeter: Nit: don't bake error info into a string?
```suggestion
    pub locked: Result<bool, anyhow::Error>,
```

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,445 @@
...
+use humility::core::Core;
+use humility::hubris::{HubrisArchive, HubrisI2cDevice};
+use humility_hiffy::HiffyContext;
+use humility_idol::{HubrisIdol, IdolArgument, IdolDecodeError};
+use indicatif::{ProgressBar, ProgressStyle};
+use std::fs;
+use std::path::Path;
+use tlvc::TlvcReadError;
+
+#[derive(Debug, thiserror::Error)]
```
mkeeter, line 23: It's annoying that the `#[error("..")]` annotations don't show up as docstrings.  Over in #687, I added docstrings to each variant as well so that they show up in `cargo docs` (and added a `#![warn(missing_docs)]` annotation); up to you if you want to do that here.

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,445 @@
...
+        let data = if read_data {
+            VpdData::Data(vpd_slurp(core, &mut context, hubris, &mut target))
+        } else {
+            VpdData::NotRead
+        };
+
+        items.push(VpdEntry { ndx, locked, data, device });
+    }
+
+    if items.is_empty() { Err(VpdError::NoVpd) } else { Ok(items) }
```
mkeeter, line 175: Vibes: if there are no VPDs in the manifest, an empty list seems like a valid response to me (rather than returning an error).

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,445 @@
...
+    for d in devices {
+        match (d.locked, d.data) {
+            (Ok(true), _) => {
+                // Do nothing, this is already locked
+            }
+            (Ok(false), VpdData::Data(Ok(_))) => {
+                locking.push(d.ndx);
+            }
+            (Err(_), _) | (_, VpdData::Data(Err(_))) => {
+                any_missing = true;
```
mkeeter: Now that we're always bailing on `any_missing`, want to just return the error here?

mkeeter: We could also make the error more specific (e.g. the device index and/or inner error), but I'm fine punting on that.

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,445 @@
...
+    let lock_op = hubris
+        .get_idol_command("Vpd.permanently_lock")
+        .map_err(VpdError::Idol)?;
+
+    if any_missing {
+        return Err(VpdError::VpdEntry);
+    }
+
+    // We aren't missing any VPD and don't have any VPD to lock.
+    // vpd_list returns an error on an empty VPD list so there's
```
mkeeter: I think that `vpd_list` always returns a list that's the same size as "number of VPDs in the manifest", so this comment is a little confusing.

Like I said above, I think an empty list should be valid to return from `vpd_list`, and I'm fine translating that into `AlreadyLocked` here, because every VPD is (tautologically) locked.

mkeeter: _Maybe_ we should return a `VpdLockStatus::NoVpds` variant as a way to make the caller acknowledge that they did a silly thing (called `lock_all` on a manifest with no VPDs)?

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    Ok(VpdLockStatus::Count(locking.len()))
+}
+
+/// Permanently lock a single VPD
+pub fn vpd_lock(
+    hubris: &HubrisArchive,
+    core: &mut dyn Core,
+    mut target: VpdTarget,
```
mkeeter: Does this still need to be `mut`?

mkeeter (approved):
A few small comments / questions, but this broadly looks good to me.

humility-vpd-lib/src/lib.rs
labbott: I went back and forth on this. I feel like if you're calling something to list VPDs you are expecting there to be some entries so an empty list would be a surprise and an error is much more obvious  vs attempting to iterate over an empty list.

humility-vpd-lib/src/lib.rs
labbott: I still think the error from `vpd_list` is the better choice but I'll tweak the comment to make it clearer

labbott: This comment wasn't adding much so I just removed it

humility-vpd-lib/src/lib.rs
connorl309: As somebody wholly unaffiliated with Oxide and its hardware, but as somebody that works on enterprise server firmware and curious about Oxide's work, I wonder if this is simply an artifact of EEPROMs containing VPD predominantly being interfaced with via I2C. I am aware that Oxide tries to not (or exclusively doesn't?) use a BMC in its systems; in this case, why did hardware design opt to only use I2C EEPROMs? Is this purely because vendor components (DIMMs immediately come to mind) exclusively are designed for the I2C interface?

humility-vpd-lib/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum VpdError {
+    /// Failure from hiffy
+    #[error("hiffy error")]
+    Hiffy(#[source] anyhow::Error),
+    /// Failure from idol
+    #[error("idol error")]
+    Idol(#[source] anyhow::Error),
+    /// I/O problem
+    #[error("I/O error")]
```
hawkw: personally, i would prefer to use 
```suggestion
    #[error(transparent)]
```
here. i don't think that formatting the top-level error as "I/O error" and only providing the more detailed description when formatting the whole cause chain actually gets us anything here. and i think that the formatted description of the various IO errors are pretty obviously IO errors.

labbott: Doesn't like this
```
error: transparent variant can't contain #[source]
  --> humility-vpd-lib/src/lib.rs:35:9
   |
35 |         #[source]
   |         ^^^^^^^^^

error: could not compile `humility-vpd-lib` (lib) due to 1 previous error
```

I was following the guide about error types which says to use `source` everywhere

PR #672, opened 2026-05-20
Switch to using `Duration` with `HiffyContext`

labbott:
This makes it much clearer for eventually doing humility as a library. There's not a nice way to have clap automatically parse a duration but we can at least change all the arguments to be a `u64` to easily pass to Duration.

### DIFF
--- a/cmd/auxflash/src/lib.rs
+++ b/cmd/auxflash/src/lib.rs
@@ -23,9 +23,9 @@ struct AuxFlashArgs {
     /// sets timeout
     #[clap(
         long, short = 'T', default_value_t = 15000, value_name = "timeout_ms",
-        value_parser = parse_int::parse::<u32>
+        value_parser = parse_int::parse::<u64>
     )]
-    timeout: u32,
+    timeout: u64,
 
     #[clap(subcommand)]
     cmd: AuxFlashCommand,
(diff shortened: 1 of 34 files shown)


cmd/auxflash/src/lib.rs
```diff
@@ -101,7 +101,8 @@ fn auxflash(context: &mut ExecutionContext) -> Result<()> {
     let subargs = AuxFlashArgs::try_parse_from(&context.cli.cmd)?;
     let hubris = &context.cli.archive()?;
     let core = &mut *context.cli.attach_live_booted(hubris)?;
-    let mut worker = AuxFlashHandler::new(hubris, core, subargs.timeout)?;
+    let timeout = std::time::Duration::from_millis(subargs.timeout);
```
hawkw, line 104: i suppose the alternative to this could be to write a wrapper type which has a `FromStr` implementation that parses an integer and then `Duration::from_millis`es it, but...meh. This seems fine.

cmd/console-proxy/src/posix.rs
```diff
@@ -38,15 +38,16 @@ impl<'a> UartConsoleHandler<'a> {
     pub fn new(
         hubris: &'a HubrisArchive,
         core: &'a mut dyn Core,
-        hiffy_timeout: u32,
-        poll_interval: u32,
+        hiffy_timeout: u64,
+        poll_interval: u64,
```
hawkw: personally I feel like maybe we ought to accept these as `Duration`s here and make the caller do that, but maybe it's annoyinger that way?

hawkw (approved):

cmd/auxflash/src/lib.rs
mkeeter: I had the same thought, and the same "...meh" reaction

mkeeter (approved):

PR #674, opened 2026-05-21
Add `HubrisArchive::load_from_path`

labbott:
For library functions this is a nice shortcut if we don't need to directly manipulate the `RawHubrisArchive`


humility-core/src/hubris.rs
```diff
@@ -1336,6 +1336,13 @@ impl HubrisArchive {
         inlined
     }
 
+    pub fn load_from_path<P: AsRef<Path> + std::fmt::Debug + Copy>(
```
labbott, line 1400: The parameter was borrowed from hubtools, certainly open to suggestions

mkeeter: I'm the person who wrote them in `hubtools`, but cannot fathom why I added `Copy` as a bound here.  I opened https://github.com/oxidecomputer/hubtools/issues/66

mkeeter (approved):

PR #676, opened 2026-05-21
Clean up errors in `humility-arch-arm`

mkeeter:
This is a tiny PR which gives `humility-arch-arm` fine-grained error types.

### DIFF
--- a/humility-arch-arm/src/lib.rs
+++ b/humility-arch-arm/src/lib.rs
@@ -210,40 +210,6 @@ fn instr_operands(cs: &Capstone, instr: &capstone::Insn) -> Vec<ARMRegister> {
     rval
 }
 
-#[derive(Debug, thiserror::Error)]
-pub enum InstructionError {
-    #[error("multiple source registers")]
-    MultipleSourceRegisters,
-    #[error("multiple target registers")]
-    MultipleTargetRegisters,
-}
-
-fn instr_source_target(
-    cs: &Capstone,
-    instr: &capstone::Insn,
-) -> Result<(Option<ARMRegister>, Option<ARMRegister>), InstructionError> {
-    let detail = cs.insn_detail(instr).unwrap();
-
-    let mut source: Option<ARMRegister> = None;
-    let mut target: Option<ARMRegister> = None;
-
-    for op in detail.regs_read() {
-        if source.is_some() {
-            return Err(InstructionError::MultipleSourceRegisters);
-        }
-        source = Some((*op).into());
-    }
-
-    for op in detail.regs_write() {
-        if target.is_some() {
-            return Err(InstructionError::MultipleTargetRegisters);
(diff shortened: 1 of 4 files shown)


humility-arch-arm/src/lib.rs
```diff
@@ -211,25 +210,33 @@ fn instr_operands(cs: &Capstone, instr: &capstone::Insn) -> Vec<ARMRegister> {
     rval
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum InstructionError {
```
jamesmunns: Do we want to make this `#[non_exhaustive]` to avoid semver breakage of humility-as-a-lib?

This makes things a little tedious in some places, but we might want to start caring about the impact of semver breaking changes if we expect to start having external consumers.

humility-arch-arm/src/lib.rs
```diff
@@ -211,25 +210,33 @@ fn instr_operands(cs: &Capstone, instr: &capstone::Insn) -> Vec<ARMRegister> {
...
+    #[error("multiple source registers")]
+    MultipleSourceRegisters,
```
jamesmunns: Might be worth including doc comments on the enum, and on each variant, at least answering "what does this error mean, and what should a user do about it if they get it". If I got these errors, I'd probably have to go ping you in chat to figure out what I was doing wrong.

If we expect users to start handling these errors (instead of just bailing), we should probably give them some useful feedback!

jamesmunns (approved):
Overall the change looks reasonable, we might want to get in the habit of being a bit more complete in documentation, or at least improve-as-we-go.

humility-arch-arm/src/lib.rs
```diff
@@ -211,25 +210,33 @@ fn instr_operands(cs: &Capstone, instr: &capstone::Insn) -> Vec<ARMRegister> {
     rval
 }
 
+#[derive(Debug, thiserror::Error)]
```
jamesmunns: Oh, we might also want to hit some common derives here as well, PartialEq, Clone, and Hash are usually big ones.

Clone is just generally useful, PartialEq allows for easier asserting (incl in tests), Hash is also one of those "eventually someone wants to stick it in a map" kind of things.

humility-arch-arm/src/lib.rs
mkeeter: I don't think so – I'd like to minimize tedium, and am fine playing fast-and-loose with semver right now (since the only users of the library are us).

mkeeter:
@jamesmunns 

After actually looking at the error, I think it's best to just delete it: the error occurs if a `mov` instruction has multiple source or destination registers, which is Not Allowed, so I'm going fail via panicking instead.  I've inlined `instr_source_target` into the one place it was used, to make it more obvious.

PR #677, opened 2026-05-21
Construct `ExecutionContext` in `humility-bin`

mkeeter:
(staged on #666)

`ExecutionContext::new` is a weird function: it _usually_ builds an `ExecutionContext`, but sometimes bails out unceremoniously with `std::process::exit`.  This is because it's also handling a final set of subcommands-as-cli-flags, e.g. `--list-targets`.

In this PR, we move all of that logic into `humility-bin`, which is already designed to exit with a `std::process::ExitCode` (instead of calling `exit(..)` directly).

Logic should be the same, but this moves `--list-targets` handling earlier (which should fix #597).


humility-bin/src/main.rs
```diff
@@ -35,21 +38,200 @@ fn main() -> std::process::ExitCode {
...
+        && cli.probe.is_none()
+        && cli.ip.is_none()
+        && cli.target.is_none()
+    {
+        use std::env;
+
+        if let Ok(e) = env::var("HUMILITY_PROBE") {
+            cli.probe = Some(e);
+        } else if let Ok(e) = env::var("HUMILITY_IP") {
+            cli.ip = Some(e.parse().unwrap_or_else(|x| match x {}));
```
jamesmunns, line 124: Wait how is parse returning Infallible here?

humility-bin/src/main.rs
```diff
@@ -35,21 +38,200 @@ fn main() -> std::process::ExitCode {
...
+    // line -- but that isn't actually what we want here: if an option is
+    // set on the command line, we want to ignore the fact that a
+    // different mutually exclusive option may be in the environment.  So
+    // we only look at the environment if none of these is set (Clap
+    // assures that at most one is set on the command line), and we
+    // examine the environment in an order of precence: probe, ip, target,
+    // dump.  If the user has specified no command line options but
+    // multiple of these in the environment, they will get the first in
+    // this ordering that we find.
+    if cli.dump.is_none()
```
jamesmunns, line 114: Leaving a note, but it's probably not worth fixating on because this isn't new code:

It *feels* like there should be a nicer way to specify this, or at least, the semantics of what we've chosen as "the acceptable set of configuration combinations" *feels* overly complicated. Especially the repeated "is_some", matches, and asserts. I feel like I need a flowchart to follow this now :D

I have no calibration for "necessary complexity" vs "incidental complexity" here, but just leaving a note as an uninformed newcomer that "whew, this feels like a lot". It might be worth having a couple of functions for each acceptable source/combo of options to make it easier to hang documentation off of, so folks know what is/isn't allowed.

humility-bin/src/main.rs
```diff
@@ -35,21 +38,200 @@ fn main() -> std::process::ExitCode {
...
+                    msg!("failed to match environment: {err:?}");
+                    return std::process::ExitCode::FAILURE;
+                }
+            };
+
+            //
+            // Cannot specify a dump/probe/IP address and also an
+            // environment and target
+            //
+            assert!(cli.dump.is_none());
```
jamesmunns, line 144: Maybe turn these asserts into `ExitCode::FAILURE`s too, to be polite?

humility-bin/src/main.rs
```diff
@@ -35,21 +38,200 @@ fn main() -> std::process::ExitCode {
...
+    // flags, rather than subcommands, so we'll special-case them here.
+    if cli.version {
+        eprintln!(
+            "{} {} {}",
+            env!("CARGO_BIN_NAME"),
+            env!("CARGO_PKG_VERSION"),
+            PROBES_ENABLED
+        );
+        if cmd.is_some() {
+            humility::warn!("ignoring subcommand after printing version");
```
jamesmunns: Already in scope

```suggestion
            warn!("ignoring subcommand after printing version");
```

humility-bin/src/main.rs
```diff
@@ -35,21 +38,200 @@ fn main() -> std::process::ExitCode {
...
     };
-    let OuterCli { cli, cmd } = outer_cli;
-    if let Some(s) = version(&cli) {
-        eprintln!("{s}");
+    let OuterCli { mut cli, cmd } = outer_cli;
+
+    // The --version (-V) and --list-targets operations are implemented as
+    // flags, rather than subcommands, so we'll special-case them here.
+    if cli.version {
+        eprintln!(
```
jamesmunns: When do we use `humility::warn!`, `humility::msg!`, and `eprintln!`?

humility-bin/src/main.rs
```diff
@@ -35,21 +38,200 @@ fn main() -> std::process::ExitCode {
...
+                    match description {
+                        Some(d) => d,
+                        None => "-",
+                    }
+                );
+            }
+        }
+
+        if cmd.is_some() {
+            humility::warn!("ignoring subcommand after listing targets");
```
jamesmunns:
```suggestion
            warn!("ignoring subcommand after listing targets");
```

humility-bin/src/main.rs
```diff
@@ -35,21 +38,200 @@ fn main() -> std::process::ExitCode {
...
+            if cli.archive.is_some() {
+                let msg = if m.value_source("archive")
+                    == Some(clap::parser::ValueSource::CommandLine)
+                {
+                    "archive on command-line"
+                } else {
+                    "archive in environment variable"
+                };
+
+                humility::warn!(
```
jamesmunns:
```suggestion
                warn!(
```

jamesmunns (commented):
A lot of pedantic nits, feel free to take them or leave them :)

humility-bin/src/main.rs
mkeeter: Over in https://github.com/oxidecomputer/humility/pull/668, we made IP parsing lazier – the parse captures a `Result<..>` which is only checked if we're actually using the IP address.

humility-bin/src/main.rs
mkeeter: Mutual exclusion between these options is enforced by `clap`; I'll add a comment.

humility-bin/src/main.rs
mkeeter: Right now, it's purely based on vibes!  We'll want to standardize on `log`, but that's further down the line.

humility-bin/src/main.rs
mkeeter: Good catch, it wasn't in scope in the previous file that I copied from

humility-bin/src/main.rs
labbott: As noted in the rework RFD, i would like to get rid of environments and just have other tooling set `HUMILITY_PROBE` and `HUMILITY_ARCHIVE` as appropriate. Environments solved an actual problem for us but I do not think the belong in a humility binary.

labbott (approved):
Part of me really wants to have a separate binary for dealing with humility dumps simply so the logic here isn't so complicated. I have convinced myself this is fine but it should probably be spot checked against our lab machines for good measure.

humility-bin/src/main.rs
mkeeter: Yeah, I don't love this but it hasn't gotten worse than before 😄

mkeeter:
I tested this out on lab machines (with various environment / target combinations), and everything works fine.

PR #679, opened 2026-05-26
Remove `unsafe` from `humility qspi`

mkeeter:


cmd/qspi/src/lib.rs
```diff
@@ -225,9 +226,8 @@ fn optional_nbytes<'a>(
                 Ok(results) => match &results[0] {
                     Ok(buf) => {
                         if mem::size_of::<DeviceIdData>() == buf.len() {
-                            let did: DeviceIdData = unsafe {
-                                std::ptr::read(buf.as_ptr() as *const _)
-                            };
+                            let did =
```
jamesmunns: nit: Maybe just get rid of the `size_of` check, and handle the [`CastError`](https://docs.rs/zerocopy/latest/zerocopy/error/type.CastError.html) (which includes "wrong size") through anyhow to avoid an unsightly `unwrap` on [`ref_from_bytes`](https://docs.rs/zerocopy/latest/zerocopy/trait.FromBytes.html#method.ref_from_bytes)?

jamesmunns (approved):
One nit, but overall LGTM.

cmd/qspi/src/lib.rs
mkeeter: Sure, done!

PR #680, opened 2026-05-26
Move hiffy / Idol decoding to a single place

mkeeter:
Previously, we had two different functions to go from a `Result<Vec<u8>, IpcError>` to a decoded type: `idol_result` and `hiffy_decode`.  This PR removes them both in favor of a `decode` function on `IdolOperation`:

```rust
// Old (two different ways of decoding, somewhat verbose):
hiffy_decode::<u32>(hubris, &op, r)?
context.idol_result::<u32>(&op, &r)?;
// New (one way to decode, shorter):
op.decode::<u32>(&r)?;
```

The `IdolOperation` already has a handle to the `&HubrisArchive`, so we can remove that argument from `hiffy_call`.

The only significant change is in error design: we now have a strongly-typed `IdolDecodeError`, which separates "you're holding it wrong" (`IdolDecodeError::Failed`) from "Idol returned an error" (`IdolDecodeError::Idol`).  The latter is strongly-typed all the way down; the former wraps `anyhow::Error` in a few places.

### DIFF
--- a/humility-idol/src/lib.rs
+++ b/humility-idol/src/lib.rs
@@ -23,7 +23,48 @@ pub struct IdolOperation<'a> {
     pub code: u16,
     pub args: Option<&'a HubrisStruct>,
     pub ok: HubrisGoff,
-    pub error: IdolError<'a>,
+    pub error: IdolErrorType<'a>,
+}
+
+/// Error returned from [`IdolOperation::decode`]
+///
+/// This error is either an error reported by Idol ([`IdolError`]) or an error
+/// that occured in the process of decoding ([`IdolDecodeFailed`])
+#[derive(Debug, thiserror::Error)]
+pub enum IdolDecodeError {
+    /// Error reported from Idol
+    #[error(transparent)]
+    Idol(#[from] IdolError),
+
+    /// Decoding the result failed in Humility
+    #[error(transparent)]
+    Failed(#[from] IdolDecodeFailed),
+}
+
+/// Idol decode errors happening outside of Idol
+#[derive(Debug, thiserror::Error)]
+pub enum IdolDecodeFailed {
+    #[error("failed to load zerocopy value")]
+    BytecastFailed(#[source] anyhow::Error),
+    #[error("failed to deserialize value")]
+    DeserializeValueFailed(#[source] anyhow::Error),
+    #[error("failed to load value with reflection")]
+    LoadFailed(#[source] anyhow::Error),
+    #[error("operation has no error type, but decode was called with an error")]
+    NoErrorType,
+}
+
+/// Errors reported by Idol
+#[derive(Debug, thiserror::Error)]
+pub enum IdolError {
+    #[error("<server died; its new ID is {0}>")]
+    ServerDied(u32),
+    #[error("<complex error: {0}>")]
+    ComplexError(String),
+    #[error("<unknown variant: {0}>")]
+    UnknownErrorVariant(u32),
+    #[error("{0}")]
+    Named(String),
 }
 
 #[derive(Debug)]
@@ -42,7 +83,7 @@ where
 }
 
 #[derive(Debug)]
-pub enum IdolError<'a> {
+pub enum IdolErrorType<'a> {
     CLike(&'a HubrisEnum),
     Complex(&'a HubrisEnum),
     None,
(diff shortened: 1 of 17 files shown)


humility-idol/src/lib.rs
```diff
@@ -23,7 +23,48 @@ pub struct IdolOperation<'a> {
...
+/// that occured in the process of decoding ([`IdolDecodeFailed`])
+#[derive(Debug, thiserror::Error)]
+pub enum IdolDecodeError {
+    /// Error reported from Idol
+    #[error(transparent)]
+    Idol(#[from] IdolError),
+
+    /// Decoding the result failed in Humility
+    #[error(transparent)]
+    Failed(#[from] IdolDecodeFailed),
```
hawkw: how do you feel about naming this something like 
```suggestion
    /// Decoding the result failed in Humility
    #[error(transparent)]
    DecodeFailed(#[from] IdolDecodeFailed),
```
instead? `Failed` feels a bit unclear to me --- it could just as easily mean "the idol call failed", which is the opposite of what this one means...

humility-idol/src/lib.rs
```diff
@@ -248,6 +289,66 @@ impl<'a> IdolOperation<'a> {
...
     }
+
+    /// Decodes a value returned from calling the given Idol operation
+    pub fn decode<T: humility::reflect::Load>(
+        &self,
+        val: &Result<Vec<u8>, IpcError>,
+    ) -> Result<T, IdolDecodeError> {
+        match val {
+            Ok(val) => {
+                let ty = self.hubris.lookup_type(self.ok).unwrap();
```
hawkw: hm, since we are returning errors in all the other "you're holding it wrong" cases, i kind of wonder if we might want to return another error variant here to say "the type you tried to decode the thing to couldn't be found"? maybe not, I dunno.

humility-idol/src/lib.rs
```diff
@@ -248,6 +289,66 @@ impl<'a> IdolOperation<'a> {
...
+                    }
+                    ::idol::syntax::Encoding::Ssmarshal
+                    | ::idol::syntax::Encoding::Hubpack => {
+                        humility::reflect::deserialize_value(
+                            self.hubris,
+                            val,
+                            ty,
+                        )
+                        .map_err(IdolDecodeFailed::DeserializeValueFailed)?
+                        .0
```
hawkw: i wonder a bit if we might want to have separate error variants for `Ssmarshal` and `hubpack` here so that we can tell you which encoding we were trying to use? or, does the error from `humility::reflect::deserialize_value` already tell you that?

humility-idol/src/lib.rs
```diff
@@ -248,6 +289,66 @@ impl<'a> IdolOperation<'a> {
...
+                IdolErrorType::Complex(error) => {
+                    IdolError::ComplexError(error.name.to_string())
+                }
+                IdolErrorType::None => {
+                    return IdolDecodeFailed::NoErrorType.into();
+                }
+            },
+            IpcError::ServerDied(c) => IdolError::ServerDied(c),
+        };
+        IdolDecodeError::Idol(out)
```
hawkw: to be honest, i found the use of `out` here a little hard to follow, i only noticed that after seeing that in the `IdleErrorType::None` case, there was an explicit return`. Part of me wants to say, why don't we just make all the paths that return an `IdolError` explicitly wrap that in `IdolDecodeError::Idol` for clarity, even if it's much uglier-looking? i dunno...

hawkw (commented):
overall, i like this change! i had a few smallish suggestions.

humility-idol/src/lib.rs
mkeeter: I decided to just call `self.hubris.lookup_type` in the constructor and store the value in the `IdolOperation`, making this a moot point.

humility-idol/src/lib.rs
mkeeter: Seems reasonable, done!

humility-idol/src/lib.rs
mkeeter: I decided to instead glomp everything in a `DeserializeFailed` variant, then include the actual Idol encoding.

humility-idol/src/lib.rs
mkeeter: Done (slightly differently, by just calling `.into()` everywhere)

hawkw (approved):
Thanks for addressing the view feedback, this looks great!

labbott (commented):
LGTM I'll let others take a pass before I approve

PR #682, opened 2026-05-26
Move `hiffy_call` into a method on `HiffyContext`

mkeeter:
(Staged on #681)

This is a small simplification PR to make `hiffy_call` into a method on `HiffyContext`, since that's the type which clearly "owns" the behavior.

### DIFF
--- a/humility-hiffy/src/lib.rs
+++ b/humility-hiffy/src/lib.rs
@@ -1368,6 +1368,77 @@ impl<'a> HiffyContext<'a> {
     pub fn scratch_size(&self) -> usize {
         self.scratch_size
     }
+
+    /// Executes a Hiffy call, returning the output value (or an error)
+    pub fn call<T: humility::reflect::Load>(
+        &mut self,
+        core: &mut dyn Core,
+        op: &idol::IdolOperation,
+        args: &[(&str, idol::IdolArgument)],
+        lease_write: Option<&[u8]>,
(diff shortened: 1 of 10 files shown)


humility-hiffy/src/lib.rs
```diff
@@ -1368,6 +1368,75 @@ impl<'a> HiffyContext<'a> {
     pub fn scratch_size(&self) -> usize {
         self.scratch_size
     }
+
+    /// Executes a Hiffy call, returning the output value (or an error)
+    pub fn call<T: humility::reflect::Load>(
```
hawkw, line 1373: i kinda wonder if this ought to still be named 
```suggestion
    pub fn hiffy_call<T: humility::reflect::Load>(
```
since all the call sites for it basically look like `context.call(...)` which doesn't make it clear that it's hiffy. i realize that the type is called `HiffyContext`, but the references to

labbott (commented):
LGTM I have no opinion on naming here

hawkw (approved):
i also don't feel strongly about the naming thing

humility-hiffy/src/lib.rs
mkeeter: >  but the references to

RIP Eliza, eaten by a grue mid-comment 🪦

(I'm going to keep this as `call`, since like you said it's a method on a `HiffyContext`)

PR #683, opened 2026-05-26
Remove `strerror` in favor of `decode_error`

mkeeter:
(staged on #682)

This is a small PR to consolidate `IpcError` decoding into a single place (`IdolOperation::decode_error`).

I don't try to preserve the exact error formatting; if you have opinions on what information would be useful, see #680 for where the new error type is introduced.  I'd be surprised if anyone was parsing CLI output for exact strings here, but cc @adamlouis just in case.

### DIFF
--- a/humility-dump-agent/src/hiffy.rs
+++ b/humility-dump-agent/src/hiffy.rs
@@ -2,12 +2,12 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 use crate::DumpAgent;
-use anyhow::{Result, bail};
+use anyhow::{Context, Result, bail};
 use core::mem::size_of;
 use hif::*;
 use humility::{core::Core, hubris::HubrisArchive};
 use humility_hiffy::{HiffyContext, IpcError};
-use humility_idol::{self as idol, HubrisIdol};
+use humility_idol::{self as idol, HubrisIdol, IdolDecodeError, IdolError};
 use humpty::{DumpAreaHeader, DumpSegment, DumpSegmentHeader};
 use std::time::Duration;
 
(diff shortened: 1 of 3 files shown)


labbott (approved):

humility-dump-agent/src/hiffy.rs
```diff
@@ -162,7 +164,7 @@ impl DumpAgent for HiffyDumpAgent<'_> {
         let results = self.run(ops.as_slice())?;
 
         if let Err(err) = results[rindex] {
-            bail!("failed to take dump: {}", op.strerror(err))
+            return Err(op.decode_error(err)).context("failed to take dump");
         }
```
hawkw, line 168: nit, take it or leave it: now that this is no longer using `bail`, i feel like it could be changed to something like
```rust
results[rindex].map_err(|err| op.decode_error(err).context("failed to take dump")?;
```

also: i was...quite struck...by the potential alternative interpretation of the string "failed to take dump". if this is not immediately obvious to you, you're probably better off that way!

humility-dump-agent/src/hiffy.rs
```diff
@@ -97,11 +98,12 @@ impl DumpAgent for HiffyDumpAgent<'_> {
...
-                    "failed to add segment at address {base:#x} for length \
-                    {size}: {}",
-                    op.strerror(*err),
-                );
+                return Err(op.decode_error(*err)).with_context(|| {
+                    format!(
+                        "failed to add segment at address {base:#x} \
+                         for length {size}"
+                    )
+                });
```
hawkw, line 106: take it or leave it: this could `?` now that it no longer uses `bail!`?

hawkw (approved):

humility-dump-agent/src/hiffy.rs
mkeeter: sooooo I considered writing it in that form, but vaguely decided against it for two reasons:

- `rustfmt` wraps onto multiple lines, so it's not any shorter
- Vague vibes: I feel like `?` as a way to return an error – when you're not using the actual value – is a little less obvious than an explicit `return`.  It's fine in a chain where the end result is used, but seems to sneaky when used as a one-character conditional `return`.

(I have no further comment on the text of the string 🙈)

humility-dump-agent/src/hiffy.rs
mkeeter: Same here, I'd rather be explicit versus `result.map_err(...)?`

PR #684, opened 2026-05-27
Fix parsing of dashboard timeout arg

evan-oxide:
This was causing `humility dashboard` to panic with "Mismatch between definition and access of `timeout`. Could not downcast to u64, need to downcast to u32"


labbott (approved):
good catch!

PR #686, opened 2026-05-27
calculate max lengths for nicer tabular output, add refdes to `sensors list`

hawkw:
Building on some of the changes in #667, this commit changes the manifest-loading code to track the maximum lengths of more fields, so that we can use them to dynamically determine the character lengths of columns in various tabular outputs. This way, they are always long enough to prevent something from being weird-looking.

I also added a `refdes` column to `humility sensors list` while I was messing with stuff.


humility-bin/tests/cmd/manifest/manifest.control_plane_agent.overflow.0.stdout
```diff
@@ -45,219 +45,219 @@
...
-                 75  2 B  1:1 0x6a -      m2_hp_only    M.2 A NVMe Basic Management Command
-                 76  2 B  1:2 0x6a -      m2_hp_only    M.2 B NVMe Basic Management Command
+                 ID C  P  MUX ADDR REFDES DEVICE     DESCRIPTION
+                  0  2 F  -   0x48 J194   tmp117     Southwest temperature sensor
+                  1  2 F  -   0x49 J195   tmp117     South temperature sensor
+                  2  2 F  -   0x4a J196   tmp117     Southeast temperature sensor
+                  3  2 F  -   0x70 U336   pca9545    U.2 ABCD mux
+                  4  2 F  -   0x71 U339   pca9545    U.2 EFGH mux
+                  5  2 F  -   0x72 U337   pca9545    U.2 IJ/FRUID mux
+                  6  2 F  1:1 0x50 J206   at24csw080 U.2 Sharkfin A VPD
```
mkeeter, line 55: Could we switch to a minimum of two spaces between device and description?  Since the descriptions have spaces themselves, having a single space makes it harder to visually separate
```
at24csw080 U.2 Sharkfin A VPD
```

mkeeter: (Same comment applies basically everywhere)

hawkw: sure, can do!

PR #688, opened 2026-06-01
Convert to `slog` as our logging library

mkeeter:
Per RFD 659, this converts us to using `slog` everywhere.

This PR is mostly mechanical!  `slog` expects a logger object to be threaded through function calls.  Here, we lazily construct it in the `Cli`, then grab it from the `ExecutionContext`.  It's mostly passed as an argument, although in places where we build an `XyzContext` (e.g. `HiffyContext`), we clone a copy of the logger and store it locally to reduce boilerplate.

(cloning a `slog::Logger` is expected to be cheap)

Note that we still use `env_logger` for configuration in the `humility-bin` package, so that we can get verbose `probe-rs` logs (and maybe other libraries as well, but that's the loudest one).


mkeeter:
One thing that I'd like opinions on: should we re-export the `slog` types from `humility` (perhaps via the now-deleted `humility-log` crate)?  I could imagine exporting the various logging macros and `Logger` type, so that people using downstream crates wouldn't need a `slog` dependency of their own.

mkeeter:
> One thing that I'd like opinions on: should we re-export the `slog` types from `humility` (perhaps via the now-deleted `humility-log` crate)?  I could imagine exporting the various logging macros and `Logger` type, so that people using downstream crates wouldn't need a `slog` dependency of their own.

I've decided that I like this, but it's in a standalone commit at the top of the stack if anyone strongly disagrees.

jamesmunns (approved):
Let's go

hawkw (approved):
sure fine whatever

PR #689, opened 2026-06-02
Only `load` with a probe

mkeeter:
`Core::load` is only implemented for `ProbeCore`; all other cores just `bail!`.  This PR removes it from the trait and moves it to a function on `ProbeCore`.

We then change the return type of unambiguous attachment functions to be a specific core type, instead of a `Box<dyn Core>`.  In cases where we need the type-erased box, it's easy to convert further down the line, but I think that library functions should defer type-erasing until it's actually needed.

(This is a building block for a `humility-flash` library crate)

### DIFF
--- a/humility-probes-core/src/probe_rs.rs
+++ b/humility-probes-core/src/probe_rs.rs
@@ -204,7 +204,45 @@ impl Core for ProbeCore {
         Ok(())
     }
 
-    fn load(&mut self, path: &Path) -> Result<()> {
+    fn reset(&mut self) -> Result<()> {
+        let mut core = self.session.core(0)?;
+        core.reset()?;
+        self.halted = false;
+        Ok(())
+    }
+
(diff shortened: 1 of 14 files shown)


humility-probes-core/src/probe_rs.rs
```diff
@@ -204,7 +204,45 @@ impl Core for ProbeCore {
         Ok(())
     }
 
-    fn load(&mut self, path: &Path) -> Result<()> {
+    fn reset(&mut self) -> Result<()> {
```
mkeeter, line 207: Github is butchering the diff here; these functions haven't changed, but `load` got moved out of the `impl Core` block.

labbott (approved):
Haha yes no more `Box`

mkeeter:
@labbott this is going to fight with #619, so I'm fine waiting for that to merge

mkeeter:
(Laura and I chatted and decided to merge this first, since she wants to test her stuff in a mfg context)

PR #690, opened 2026-06-03
Load ELF data directly using `probe_rs`

mkeeter:
(staged on #689)

Now that we are using `probe-rs` for everything, we don't need to convert to IHEX / SREC.  This lets us remove a bunch of dependencies.  I also switched `ProbeCore::load` to return a strong error type, while I was poking at it.

This is a building block for a `humility flash` library.

Bidding a fond farewell to the best docstring in the codebase, describing `elf_chunks`:

> While it may sound like the impetus for an OSHA investigation at the North Pole...


mkeeter:
I flashed both the SP and RoT on Grapefruit, and everything looks fine.

cmd/rebootleby/Cargo.toml
```diff
@@ -11,9 +11,5 @@ humility-cortex = { workspace = true }
 humility-cli = { workspace = true }
 clap = { workspace = true }
 anyhow = { workspace = true }
-parse_int = { workspace = true }
```
mkeeter, line 14: These were unused dependencies that I happened to notice while pruning.

humility-probes-core/src/probe_rs.rs
```diff
@@ -261,17 +277,19 @@ impl ProbeCore {
...
 
         let bar = ProgressBar::new(0);
         let erase_style = ProgressStyle::default_bar()
-            .template("humility: erasing [{bar:30}] {bytes}/{total_bytes}")?;
+            .template("humility: erasing [{bar:30}] {bytes}/{total_bytes}")
+            .unwrap();
         let flash_style = ProgressStyle::default_bar()
-            .template("humility: flashing [{bar:30}] {bytes}/{total_bytes}")?;
+            .template("humility: flashing [{bar:30}] {bytes}/{total_bytes}")
+            .unwrap();
```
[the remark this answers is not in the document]
mkeeter: This is deliberate; the progress bar can only fail if it's given an invalid template string, and we won't do that (or will catch it quickly, because it's a loud failure mode).

humility-probes-core/src/probe_rs.rs
```diff
@@ -318,14 +336,9 @@ impl ProbeCore {
...
-            &mut self.session,
-            path,
-            flashing::Format::Hex,
-            options,
-        ) {
-            bail!("Flash loading failed {:?}", e);
-        };
+        let mut loader = self.session.target().flash_loader();
+        loader.load_elf_data(&mut std::io::Cursor::new(elf_data))?;
+        loader.commit(&mut self.session, options)?;
```
labbott, line 317: This is  going to change ever so slightly when we do the `probe-rs` upgrade https://docs.rs/probe-rs/latest/probe_rs/flashing/struct.FlashLoader.html#method.load_image I think it should still match up okay and we can just use the default `ElfOptions` https://docs.rs/probe-rs/latest/probe_rs/flashing/struct.ElfOptions.html

labbott (commented):
This makes me a little nervous but I think we're probably okay because our elf is very simple

```
$ arm-none-eabi-readelf --segments img/final.elf

Elf file type is REL (Relocatable file)
Entry point 0x10181
There is 1 program header, starting at offset 52

Program Headers:
  Type           Offset   VirtAddr   PhysAddr   FileSiz MemSiz  Flg Align
  LOAD           0x000054 0x00010000 0x00010000 0x339e8 0x339e8 R   0x20

 Section to Segment mapping:
  Segment Sections...
   00     .sec1 
```

```
$ arm-none-eabi-readelf --segments img/final.elf
 
Elf file type is REL (Relocatable file)
Entry point 0x80002e9
There is 1 program header, starting at offset 52

Program Headers:
  Type           Offset   VirtAddr   PhysAddr   FileSiz MemSiz  Flg Align
  LOAD           0x000054 0x08000000 0x08000000 0x88100 0x88100 R   0x20

 Section to Segment mapping:
  Segment Sections...
   00     .sec1 
```

I guess my biggest concern would be if there's any rounding changes. We may also want to take a closer look at the updated `DownloadOptions` in new `probe-rs` https://docs.rs/probe-rs/latest/probe_rs/flashing/struct.DownloadOptions.html to make sure we're getting the behavior we want.

mkeeter:
@labbott Can you say more about what rounding changes would look like, e.g. are you worried about a different erase / write granularity?

labbott:
> @labbott Can you say more about what rounding changes would look like, e.g. are you worried about a different erase / write granularity?

Currently, we're explicitly breaking up the ELF into 32 byte chunks and using that to generate the `ihex`. I'm concerned if there could be different behavior for the last unaligned block. e.g we had an image of 512+1 bytes in length, that gets converted to  (512/32)  + 1 chunks in ihex. Block write size is 512. I would expect that that last byte would get handled the same way for both images and that last block should look the same in all cases but I'd want to be sure.

mkeeter:
> Currently, we're explicitly breaking up the ELF into 32 byte chunks and using that to generate the `ihex`. I'm concerned if there could be different behavior for the last unaligned block. e.g we had an image of 512+1 bytes in length, that gets converted to  (512/32)  + 1 chunks in ihex. Block write size is 512. I would expect that that last byte would get handled the same way for both images and that last block should look the same in all cases but I'd want to be sure.

I dug into the `probe-rs` code and am now feeling more confident.

All data loading happens through a `FlashLoader` object.  `load_elf_data` finds `PT_LOAD` sections and grabs their data; the previous `ihex` path finds `DATA` records and behaves similarly.

In both cases, data is accumulated by calling `FlashLoader::add_data`.  The important part is [here](https://github.com/oxidecomputer/probe-rs/blob/078eafb444c4e03fa7d41c7273264d00564210ce/probe-rs/src/flashing/builder.rs#L203-L207): if data is contiguous, then it's accumulated into the same buffer (instead of a separate allocation).  The chunks produced by `elf_chunks` are contiguous within an individual `PT_LOAD` section, so they should be merged back into a single allocation; this means that the `ihex` and ELF paths should look the same by the time the `FlashLoader` is done accumulating data.

(and presumably this won't change with the `probe-rs` bump, but I can recheck when rebasing)

labbott (approved):
Thanks for digging into that. I think we should be fine then.

PR #693, opened 2026-06-04
Build helios package in CI

citrus-it:
Build a helios package for humility in CI.
This will be picked up by [grabber](https://github.com/oxidecomputer/lengths-of-wire/tree/main/grabber), that runs on the package server; the resulting package will automatically be published to the repository.


mkeeter (approved):
lgtm (as an illumos packaging expert)

PR #694, opened 2026-06-04
Make `-p archive` fallback explicit

mkeeter:
(staged on #692)

Previously, `attach_probe` would attach to either a probe (as expected) or an archive if the probe name was "archive" (surprise!).

Though this is surprising, it's load-bearing for a few integration tests – `humility readvar -p archive ...` works fine for variables which are present in the archive itself, e.g. kernel config tables, and would otherwise try (and fail) to connect to a probe.

This PR moves that logic to a dedicated `attach_probe_with_archive_fallback` function.

(marking this as a draft because I'm also going to try changing this in `readvar` instead)

### DIFF
--- a/cmd/probe/src/lib.rs
+++ b/cmd/probe/src/lib.rs
@@ -88,6 +88,7 @@
 
 use anyhow::{Result, bail};
 use clap::Parser;
+use humility::core::Core;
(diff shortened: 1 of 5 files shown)


mkeeter:
Closing in favor of #695

PR #695, opened 2026-06-04
Remove `-p archive` fallback

mkeeter:
(staged on #692)

Previously, if you specified `--probe=archive`, Humility would attach to the archive instead of the probe, even in subcommands which wanted a live system.  This was useful in one case: `readvar` could read variables from flash.

This PR removes the fallback: `attach_probe` now returns a concrete `ProbeCore`, without the `--probe=archive` special case.  As the only user of the old behavior, `humility readvar` grows a new `--archive` flag to compensate.

(This is my new preferred alternative to #694)

### DIFF
--- a/cmd/probe/src/lib.rs
+++ b/cmd/probe/src/lib.rs
@@ -88,6 +88,7 @@
 
 use anyhow::{Result, bail};
 use clap::Parser;
+use humility::core::Core;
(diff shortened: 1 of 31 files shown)


labbott (approved):
Honestly I think I forgot you even could do `-p archive`

PR #698, opened 2026-06-10
Add library for `humility dump`

labbott:
This is broken up into two commits

1) Refactor `humility dump` to make some of the logic easier to follow
2) Add the dumping library

Because of the way `humility dump` currently works it's tricky to have all the library functions replace the same functions in the binary. Almost tempted to say we should start from scratch with `humility dump-ng` but that's probably a bad idea.

### DIFF
--- a/cmd/dump/src/lib.rs
+++ b/cmd/dump/src/lib.rs
@@ -408,7 +408,8 @@ fn print_dump_breakdown(breakdown: &DumpBreakdown, log: &Logger) {
     print_val("data expansion", breakdown.inverted);
 }
 
-fn dump_via_agent(
+/// Simulation and testing features
+fn simulate_dump_via_agent(
     hubris: &HubrisArchive,
     core: &mut dyn Core,
     subargs: &DumpArgs,
@@ -552,135 +553,55 @@ fn dump_via_agent(
 
         core.run()?;
         info!(log, "core resumed");
-    } else {
+    } else if subargs.emulate_dumper {
         let segments = hubris.dump_segments(core, None, false)?;
         let mut agent = get_dump_agent(hubris, core, subargs, log)?;
         let header = agent.read_dump_header()?;
 
-        if !subargs.force_read && subargs.extract.is_none() {
-            if header.dumper != humpty::DUMPER_NONE
-                && !subargs.initialize_dump_agent
-                && !subargs.force_overwrite
-                && task.is_none()
-            {
-                bail!(
-                    "there appears to already be one or more dumps in situ; \
+        if header.dumper != humpty::DUMPER_NONE
+            && !subargs.force_overwrite
+            && task.is_none()
+        {
+            bail!(
+                "there appears to already be one or more dumps in situ; \
                     list them with --list and extract them with --extract_all"
-                )
-            }
-
-            if task.is_none() || subargs.initialize_dump_agent {
-                info!(log, "initializing dump agent state");
-                agent.initialize_dump()?;
-            }
-
-            if subargs.initialize_dump_agent {
-                return Ok(());
-            }
-
-            if task.is_none() {
-                info!(log, "initializing segments");
-                agent.initialize_segments(&segments)?;
-            }
+            )
         }
 
-        if subargs.emulate_dumper {
-            agent.core().halt()?;
-            info!(log, "core halted");
-
-            if let Some(ref stock) = subargs.stock_dumpfile {
-                hubris.dump(agent.core(), task, Some(stock), None, log)?;
-            }
-
-            let base = header.address;
-            let total = segments.iter().fold(0, |ttl, (_, size)| ttl + size);
-
-            let address = if task.is_some() {
-                match emulate_task_dump_prep(agent.core(), &segments, base) {
-                    Err(e) => {
-                        agent.core().run()?;
-                        info!(log, "core resumed after failure");
-                        return Err(e);
-                    }
-                    Ok(address) => {
-                        assert!(area.is_none());
-                        area = Some(DumpArea::ByAddress(address));
-                        address
-                    }
-                }
-            } else {
-                base
-            };
-
-            emulate_dump(agent.core(), task, address, total, log)?;
-            agent.core().run()?;
-            info!(log, "core resumed");
-        } else if !subargs.force_read && subargs.extract.is_none() {
-            if subargs.force_manual_initiation {
-                agent.core().halt()?;
-                info!(log, "leaving core halted");
-                let base = header.address;
-                info!(
-                    log,
-                    "unplug probe and manually \
-                    initiate dump from address {:#x}",
-                    base
-                );
-                info!(
-                    log,
-                    "e.g., \"humility hiffy --call \
-                    Dumper.dump -a address={:#x}\"",
-                    base
-                );
-                return Ok(());
-            }
-
-            //
-            // We are about to disappear for -- as the kids say -- a minute.
-            // Set our timeout to be a literal minute so we don't prematurely
-            // give up.
-            //
-            agent.core().set_timeout(std::time::Duration::new(60, 0))?;
-
-            //
-            // Tell the thing to take a dump.
-            //
-            if let Err(err) = agent.take_dump() {
-                //
-                // If that fails, it may be because we ran out of space.  Check
-                // our dump headers; if all of them are consumed, assume
-                // that we ran out of space -- and if any of them are consumed,
-                // process whatever we find (some dump is better than none!) and
-                // warn accordingly.
-                //
-                if let Ok(all) = agent.read_dump_headers(true) {
-                    let c = all
-                        .iter()
-                        .filter(|&&(h, _)| h.dumper != humpty::DUMPER_NONE)
-                        .count();
-
-                    if c == all.len() {
-                        warn!(
-                            log,
-                            "dump has indicated failure ({err:?}), but this is \
-                            likely due to space exhaustion; \
-                            dump will be extracted but may be incomplete!"
-                        );
-                    } else if c != 0 {
-                        warn!(
-                            log,
-                            "dump has indicated failure ({err:?}), but some dump \
-                            contents appear to have been written; \
-                            dump will be extracted but may be incomplete!"
-                        );
-                    } else {
-                        return Err(err);
-                    }
-                } else {
-                    return Err(err);
-                }
-            }
+        if task.is_none() {
+            info!(log, "initializing segments");
+            agent.initialize_segments(&segments)?;
         }
+        agent.core().halt()?;
+        info!(log, "core halted");
+
+        if let Some(ref stock) = subargs.stock_dumpfile {
+            hubris.dump(agent.core(), task, Some(stock), None, log)?;
+        }
+
+        let base = header.address;
+        let total = segments.iter().fold(0, |ttl, (_, size)| ttl + size);
+
+        let address = if task.is_some() {
+            match emulate_task_dump_prep(agent.core(), &segments, base) {
+                Err(e) => {
+                    agent.core().run()?;
+                    info!(log, "core resumed after failure");
+                    return Err(e);
+                }
+                Ok(address) => {
+                    assert!(area.is_none());
+                    area = Some(DumpArea::ByAddress(address));
+                    address
+                }
+            }
+        } else {
+            base
+        };
+
+        emulate_dump(agent.core(), task, address, total, log)?;
+        agent.core().run()?;
+        info!(log, "core resumed");
 
         //
         // If we're here, we have a dump in situ -- time to pull it.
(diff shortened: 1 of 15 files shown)


humility-dump-lib/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "humility-dump-lib"
```
mkeeter: Nit: I would expect this to just be `humility-dump` for consistency with other library crates (same with folder name).

humility-dump-lib/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+version = "0.1.0"
+edition.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
mkeeter: probably don't need this
```suggestion
```

humility-dump-lib/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+use slog::Logger;
+use std::path::PathBuf;
+use std::time::Instant;
+
+/// Error returned from dumping
+#[derive(Debug, thiserror::Error)]
+pub enum DumpError {
+    /// Tried to take a dump when there were pending dumps. The
+    /// dumps should be extracted first before trying to take a dump.
+    #[error("Cannot take a dump while one is already in progress")]
```
mkeeter: Nit: error `Display` strings should be lower case, per [C-GOOD-ERR](https://rust-lang.github.io/api-guidelines/interoperability.html#error-types-are-meaningful-and-well-behaved-c-good-err)

humility-dump/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+            .unwrap_or(false)
+    {
+        let imageid = hubris.image_id();
+
+        Ok(Box::new(
+            UdpDumpAgent::new(core, imageid, log)
+                .map_err(DumpError::UdpAgent)?,
+        ))
+    } else {
+        // XXX
```
mkeeter: XXX?

humility-dump-lib/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+    let areas = task_areas(&headers);
+    for (area, (task, headers)) in &areas {
+        let task_name =
+            match hubris.lookup_module(HubrisTask::Task(task.id.into())) {
+                Ok(module) => match headers[0].contents {
+                    humpty::DUMP_CONTENTS_SINGLETASK => module.name.to_owned(),
+                    humpty::DUMP_CONTENTS_TASKREGION => {
+                        format!("{}.region", module.name.to_owned())
+                    }
+                    c => panic!("unknown contents type: {c}"),
```
mkeeter: Return an error instead here?

humility-dump-lib/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    hubris
+        .dump(&mut out, None, Some(&dumpfile), started, log)
+        .map_err(DumpError::HubrisDump)?;
+
+    Ok(())
+}
+
+/// Take and collect a dump via the agent. If a file is not
+/// specified the file will be saved accoridng to the default
```
mkeeter:
```suggestion
/// specified the file will be saved according to the default
```

humility-dump-lib/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+/// humility logic (hubris.core.0, hubris.core.1 etc.)
+pub fn take_system_dump_via_agent(
+    hubris: &HubrisArchive,
+    core: &mut dyn Core,
+    dumpfile: Option<&PathBuf>,
```
mkeeter: I don't love having the naming logic in the library (I would be happier if this took a `std::io::Write` handle), but that's annoying to fix right now since it's actually in `HubrisArchive::dump`.

humility-dump-lib/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+    pub time: u64,
+}
+
+/// Available dump types
+pub enum DumpType {
+    /// No dumps
+    None,
+    /// A whole system dump and its size
+    System(u32),
+    /// Individual hubris tasks
```
mkeeter:
```suggestion
    /// Individual Hubris tasks
```

humility-dump/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+    use_hiffy: bool,
+    core: &mut dyn Core,
+    log: &Logger,
+) -> Result<DumpType, DumpError> {
+    let mut agent = get_dump_agent(hubris, core, use_hiffy, log)?;
+
+    let headers =
+        agent.read_dump_headers(false).map_err(DumpError::DumpHeaders)?;
+
+    if headers.is_empty() || headers[0].0.dumper == humpty::DUMPER_NONE {
```
mkeeter, line 409: Nit: because we pass `raw: false` to `read_dump_headers`, is it even possible to get `DUMPER_NONE`?

mkeeter: ("This is how it worked before and I don't want to deal with it in this PR" is a valid response here)

humility-dump-lib/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+        let size = headers.iter().fold(0, |ttl, h| ttl + h.written);
+
+        let task_name =
+            match hubris.lookup_module(HubrisTask::Task(task.id.into())) {
+                Ok(module) => match headers[0].contents {
+                    humpty::DUMP_CONTENTS_SINGLETASK => module.name.to_owned(),
+                    humpty::DUMP_CONTENTS_TASKREGION => {
+                        format!("{} [region]", module.name.to_owned())
+                    }
+                    c => panic!("unknown contents type: {c}"),
```
mkeeter: Return error instead of panicking?

humility-dump/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+                    }
+                    c => panic!("unknown contents type: {c}"),
+                },
+                _ => "<unknown>".to_owned(),
+            };
+
+        tasks.push(TaskDumpData {
+            area: *area,
+            size,
+            task_name,
```
mkeeter, line 446: Do we want a richer task name here in the API, instead of baking it into a string immediately?

humility-dump-lib/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+                    humpty::DUMP_CONTENTS_SINGLETASK => module.name.to_owned(),
+                    humpty::DUMP_CONTENTS_TASKREGION => {
+                        format!("{}.region", module.name.to_owned())
+                    }
+                    c => panic!("unknown contents type: {c}"),
+                },
+                _ => "<unknown>".to_owned(),
+            };
+
+        let dumpfile = (0..)
```
mkeeter: You can skip this and just pass `None` to `hubris.dump(..)`; it will then do this same logic internally.

humility-dump-lib/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+                return Err(DumpError::DumpSpaceExhaustion(err));
+            } else if c != 0 {
+                return Err(DumpError::IncompleteDump(err));
+            } else {
+                return Err(DumpError::TakeDump(err));
+            }
+        } else {
+            return Err(DumpError::TakeDump(err));
+        }
+    }
```
mkeeter: Nit: because this is a transformation on `err` and always returns, I would write it as follows
```suggestion
    agent.take_dump().map_err(|err| {
        if let Ok(all) = agent.read_dump_headers(true) {
            let c = all
                .iter()
                .filter(|&&(h, _)| h.dumper != humpty::DUMPER_NONE)
                .count();
            if c == all.len() {
                DumpError::DumpSpaceExhaustion(err)
            } else if c != 0 {
                DumpError::IncompleteDump(err)
            } else {
                DumpError::TakeDump(err)
            }
        } else {
            DumpError::TakeDump(err)
        }
    })?;
```

humility-dump-lib/src/lib.rs
mkeeter: Hmmmm, this wouldn't handle `DUMP_CONTENTS_TASKREGION` separately.  I'm not sure if we ever actually want to save a `TaskRegion` dump to file, though?

humility-dump-lib/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+    /// Time the task was dumped
+    pub time: u64,
+}
+
+/// Available dump types
+pub enum DumpType {
+    /// No dumps
+    None,
+    /// A whole system dump and its size
+    System(u32),
```
mkeeter: Nit: make this a named `System { size: u32 }`, to make it a little more obvious?

humility-dump-lib/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+        .dump(&mut out, None, Some(&dumpfile), started, log)
+        .map_err(DumpError::HubrisDump)?;
+
+    Ok(())
+}
+
+/// Take and collect a dump via the agent. If a file is not
+/// specified the file will be saved accoridng to the default
+/// humility logic (hubris.core.0, hubris.core.1 etc.)
+pub fn take_system_dump_via_agent(
```
mkeeter: Rename to `take_system_dump`?  I'm not sure what `via_agent` adds at this point.

humility-dump-lib/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+        hubris
+            .dump(&mut out, task.0, Some(&dumpfile), started, log)
+            .map_err(DumpError::HubrisDump)?;
+    }
+
+    Ok(())
+}
+
+/// Extract an existing system dump via the agent
+pub fn extract_system_dump_via_agent(
```
mkeeter: Vibes: how about splitting this into a public and private function, e.g.
```rust
/// Extract an existing system dump
pub fn extract_system_dump(
    hubris: &HubrisArchive,
    core: &mut dyn Core,
    dumpfile: PathBuf,
    use_hiffy: bool,
    log: &Logger,
) -> Result<(), DumpError> {
    let started = Instant::now();
    let mut agent = get_dump_agent(hubris, core, use_hiffy, log)?;
    extract_system_dump_via_agent(
        hubris,
        &mut *agent,
        Some(&dumpfile),
        started,
        log,
    )
}

fn extract_system_dump_via_agent(
    hubris: &HubrisArchive,
    mut agent: &mut dyn DumpAgent,
    dumpfile: Option<&std::path::Path>,
    started: std::time::Instant,
    log: &Logger,
) -> Result<(), DumpError> {
    let mut out = DumpAgentCore::new(
        HubrisFlashMap::new(hubris).map_err(DumpError::FlashMap)?,
    );

    let _ = agent
        .read_dump(None, &mut out, false, log)
        .map_err(DumpError::ReadDump)?;

    hubris
        .dump(&mut out, None, dumpfile, Some(started), log)
        .map_err(DumpError::HubrisDump)?;

    Ok(())
}
```

The advantage of this approach is that `take_system_dump` below can call `extract_system_dump_via_agent`, removing some duplicate logic.

humility-dump-lib/Cargo.toml
labbott: I did `humility-vpd-lib` before but I'm also not attached to the `-lib` suffix if others prefer it be dropped :shrug:

humility-dump/src/lib.rs
labbott: I think this was a reminder that I grabbed a timeout instead of having it passed in. I'm also looking at the code and timeout doesn't get used anywhere except `humility-net-core` which is the exact opposite case here! I might take a pass to fix this up.

humility-dump/src/lib.rs
labbott: Yeah I was copy/pasting the existing logic. I think you are correct but I'm lightly wary to remove it without more testing. I might just add another comment.

humility-dump/src/lib.rs
labbott: can you say more? It sounds like you may be proposing returning a enum ?

humility-dump-lib/Cargo.toml
mkeeter: Yeah, dropping the `lib` is my preference (sorry I didn't notice in `humility-vpd-lib`)

humility-dump/src/lib.rs
mkeeter: Yeah, replacing `task_name: String` with `task_name: TaskName`, with
```rust
pub enum TaskDumpName {
    SingleTask(String),
    TaskRegion(String),
    Unknown,
}

impl std::fmt::Display for TaskDumpName {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            Self::SingleTask(s) => s.fmt(f),
            Self::TaskRegion(s) => write!(f, "{s} [region]"),
            Self::Unknown => "<unknown>".fmt(f),
        }
    }
}
```

humility-dump-lib/src/lib.rs
labbott: I had a thought about differentiating between dump via agent vs calling `hubris.dump` directly. `hubris.dump` doesn't really need a library except for maybe a wrapper to call `core.halt`. I'll rename.

humility-dump-lib/src/lib.rs
labbott: I took a pass at pulling this out. It wasn't too bad.

humility-dump/src/lib.rs
labbott: I'd still like to fix this but I don't know if I want to depend on this so I cleaned it up with a comment for now.

cmd/dump/src/lib.rs
```diff
@@ -452,7 +453,11 @@ fn dump_via_agent(
         info!(log, "core halted");
 
         if let Some(ref stock) = subargs.stock_dumpfile {
-            hubris.dump(core, task, Some(stock), None, log)?;
+            let mut file = std::fs::OpenOptions::new()
+                .write(true)
+                .create_new(true)
+                .open(stock)?;
```
mkeeter, line 459: Nit: this no longer prints `dumping to {filename}`; I'm ambivalent on whether that matters

mkeeter: (also true elsewhere)

humility-dump/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+    agent.initialize_dump().map_err(DumpError::InitializeDump)?;
+
+    hubris
+        .dump(&mut out, None, file, Some(started), log)
+        .map_err(DumpError::HubrisDump)?;
+
+    Ok(())
+}
+
+/// Take and collect a dump via the agent. If a file is not
```
mkeeter: Docstring is now outdated, since we pass in a `std::io::Write`

humility-dump/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+    // Not all these are fatal at the moment. Just log a warning for a
+    // few of these.
+    match r {
+        Ok(_) => (),
+        Err(DumpError::DumpSpaceExhaustion(err)) => {
+            warn!(
+                log,
+                "dump has indicated failure ({err:#}), but this is \
+                    likely due to space exhaustion; \
+                    dump will be extracted but may be incomplete!"
```
mkeeter: Nit: indentation
```suggestion
                likely due to space exhaustion; \
                dump will be extracted but may be incomplete!"
```

humility-dump/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+    System {
+        /// Size of the system dump
+        size: u32,
+    },
+    /// Individual Hubris tasks
+    Tasks(Vec<TaskDumpData>),
+}
+
+/// List all available dumps
+pub fn list_dumps_via_agent(
```
mkeeter: Nit: rename to `list_dumps`, because there's no `DumpAgent` in the signature?

mkeeter (approved):

mkeeter:
Looks good, approved with a few nits

cmd/dump/src/lib.rs
labbott: Yeah that was one of the things I didn't love about the refactor. I think it's less important for the places where you are passing in a file name but I added it back

PR #699, opened 2026-06-10
Move `humility flash` to a library

mkeeter:
This PR moves the bulk of `humility flash` into a library crate.  There are two main entry points:

- `get_image_state` tells us whether the current image matches or not
- `program_image` flashes an image (and auxflash)

As always, things are fully documented and use strong error types.

The logic is broadly the same as before, but has been tightened in a few places.  For example, in the old implementation, _any_ error while checking image state (e.g. a probe error) would lead to Humility trying to reflash the image; our new implementation distinguishes between "image matches / doesn't match" and Humility-level errors.

There's one small API change: `HubrisArchive::chip` now returns a `Result<String>` (instead of an `Option<String>`).  In 2026, we should never be trying to flash / reset an archive which does not include the chip name (and this isn't used in post-mortem debugging, so it's fine to be less backwards-compatible).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
     "humility-core",
     "humility-doppel",
     "humility-dump-agent",
+    "humility-flash",
     "humility-hexdump",
     "humility-hiffy",
     "humility-i2c",
@@ -80,7 +81,7 @@ members = [
     "cmd/validate",
     "cmd/vpd",
     "cmd/writeword",
-    "xtask", "humility-log",
+    "xtask",
 ]
 resolver = "2"
 
@@ -133,6 +134,7 @@ humility-caboose = { path = "./humility-caboose" }
 humility-cli = { path = "./humility-cli", default-features = false }
 humility-doppel = { path = "./humility-doppel" }
 humility-dump-agent = { path = "./humility-dump-agent" }
+humility-flash = { path = "./humility-flash" }
 humility-hexdump = { path = "./humility-hexdump" }
 humility-hiffy = { path = "./humility-hiffy" }
 humility-i2c = { path = "./humility-i2c" }
--- a/cmd/flash/src/lib.rs
+++ b/cmd/flash/src/lib.rs
@@ -19,7 +19,7 @@
 //! will be programmed after the image is written.  See RFD 311 for more
 //! information about auxiliary flash management.
 
-use anyhow::{Context, Result, bail};
+use anyhow::{Result, bail};
 use clap::Parser;
 use humility::{
     core::Core,
@@ -27,7 +27,7 @@ use humility::{
     log::{Logger, info},
 };
 use humility_cli::{ExecutionContext, humility_cmd};
-use humility_flash::{get_image_state, program_auxflash};
+use humility_flash::{ProgramAuxflashSuccess, get_image_state, program_image};
 
 #[derive(Parser, Debug)]
 #[clap(name = "flash", about = env!("CARGO_PKG_DESCRIPTION"))]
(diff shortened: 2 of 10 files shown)


cmd/flash/src/lib.rs
```diff
@@ -36,19 +36,6 @@ pub struct FlashArgs {
     #[clap(long, short = 'F')]
     force: bool,
 
-    /// if using OpenOCD, do not actually flash, but show commands and retain
```
mkeeter, line 39: These arguments were only used for OpenOCD, which is no longer used.

cmd/flash/src/lib.rs
```diff
@@ -70,156 +57,98 @@ pub struct FlashArgs {
...
 
-/// Validates the image and auxiliary flash against our subcommand
-///
-/// If `subargs.check` is true, returns `Ok(())` on a clean check and `Err(..)`
-/// otherwise; the core is left running.
-///
-/// If `subargs.check` is false, returns `Ok(())` if the check _fails_ (meaning
-/// we should reflash), and `Err(..)` if all checks pass (meaning we should
-/// _not_ reflash).  The core is left halted if we should reflash and is running
-/// otherwise.
```
mkeeter, line 81: RIP to the most confusing function semantics in Humility

humility-core/src/hubris.rs
```diff
@@ -1124,15 +1124,6 @@ pub struct HubrisFlashMeta {
...
 }
 
-//
-// Flash information pulled from the archive
-//
-pub struct HubrisFlashConfig {
-    pub metadata: HubrisFlashMeta,
-    pub elf: Vec<u8>,
-    pub chip: Option<String>,
-}
```
mkeeter, line 1134: This is no longer used as a single object; I've exposed helper functions to get both the `elf` and `chip` data individually.

Cargo.toml
```diff
@@ -78,7 +79,7 @@ members = [
     "cmd/validate",
     "cmd/vpd",
     "cmd/writeword",
-    "xtask", "humility-log",
```
mkeeter, line 83: Dunno what was going on here, but I fixed it.

humility-probes-core/src/lib.rs
```diff
@@ -16,7 +16,7 @@ use anyhow::{Result, anyhow, bail};
 mod probe_rs;
 mod unattached;
 
-pub use probe_rs::ProbeCore;
+pub use probe_rs::{LoadError, ProbeCore};
```
mkeeter, line 21: This is newly public because `ProgramImageError::LoadFailed` wraps it.

cmd/flash/src/lib.rs
labbott: I purposely didn't want to break command line feature when I removed openocd but it's been so long since anyone really used it it's probably fine

humility-flash/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+    // error.  (It will hopefully be pretty clear to the user that a
+    // half-flashed part is going to be in an ill-defined state!)
+    //
+    let elf = hubris
+        .load_flash_elf()
+        .map_err(ProgramImageError::CouldNotReadElfData)?;
+    core.load(&elf).map_err(ProgramImageError::LoadFailed)?;
+
+    //
+    // On Gimlet Rev B, the BOOT0 pin is unstrapped -- and during a flash,
```
labbott, line 324: are there still gimlet-b's around we care about?

labbott (commented):
Want to take another pass but generally LGTM

cmd/flash/src/lib.rs
mkeeter: Yeah, I figured that anyone is using those flags they should be able to figure it out right quick.

humility-flash/src/lib.rs
mkeeter: It's still built in Hubris, so I think we should keep it around.

(I was also trying to minimize behavior changes while moving logic into the library crate)

labbott (approved):

Issue #479, opened 2024-04-18
`humility lsusb` output swaps VID/PID


hawkw:
The `humility lsusb` command claims that probe IDs are output in the format `VID:PID:SERIAL`, which is the format that a Humility environment JSON file or `humility -p` expects probes to be listed. However, the actual output appears to be backwards. For example, when I run `humility lsusb` with two ST-Link probes connected to my system, I get this output:

```console
eliza@theseus $ humility lsusb
humility: USB device scan, 2 successful and 32 failed
humility: --- successfully opened devices ---
humility: format: VID:PID:SERIAL, then manufacturer name, then product name
humility: 374e:0483:0030003C3431511237393330	STMicroelectronics	STLINK-V3
humility: 3754:0483:000B00154D46501520383832	STMicroelectronics	STLINK-V3
humility: --- failures ---
humility: could not access 32 devices:
...
```

On the other hand, the Linux `lsusb(8)` utility disagrees with `humility lsusb`, if I list devices by VID 0483, I see both ST-Links:
```console
eliza@theseus $ lsusb -d 0483:
Bus 005 Device 034: ID 0483:3754 STMicroelectronics STLINK-V3
Bus 005 Device 044: ID 0483:374e STMicroelectronics STLINK-V3
```
which seems correct to me --- I would expect both ST-Links to have the same VID but different PIDs, rather than differing VIDs.

I'll also note that if the `humility lsusb` output is copied directly to the `--probe` command-line option, Humility doesn't like it, but if I swap the order of the two IDs, it works:
```console
eliza@theseus $ humility --probe 374e:0483:0030003C3431511237393330 -a ~/Code/oxide/hubris/target/gimletlet/dist/default/build-gimletlet-image-default.zip tasks
humility tasks failed: Probe could not be created

Caused by:
    Probe was not found.

eliza@theseus $ humility --probe 0483:374e:0030003C3431511237393330 -a ~/Code/oxide/hubris/target/gimletlet/dist/default/build-gimletlet-image-default.zip tasks
humility: attached to 0483:374e:0030003C3431511237393330 via ST-Link V3
system time = 19248
ID TASK                       GEN PRI STATE
 0 jefe                         0   0 recv, notif: fault timer(T+52)
 1 sys                          0   1 recv, notif: exti-wildcard-irq(irq6/irq7/irq8/irq9/irq10/irq23/irq40)
 2 i2c_driver                   0   2 recv
 3 user_leds                    0   5 recv, notif: timer
[...]

PR #700, opened 2026-06-11
Move more `Core` functions to `ProbeCore`

mkeeter:
`vid_pid`, `step`, and `write_reg` are only available on `ProbeCore`.  This PR removes them from the `Core` trait, adjusting signatures where necessary to force the concrete type.


mkeeter:
I had to do a bit of extra wrangling to get the probeless build working; it now skips subcommands which are probe-only, both in the `Cargo.toml` (optional dependencies) and in codegen (ugly hacks).

humility-bin/build.rs
```diff
@@ -32,19 +32,42 @@ pub enum Subcommand {{
...
+                    .into_iter()
+                    .collect(),
+                // We also look at which commands are enabled by the `probes`
+                // feature on the `humility-bin` crate, then remove them.
+                metadata
+                    .workspace_members
+                    .iter()
+                    .find(|p| metadata[p].name == "humility-bin")
+                    .map(|p| metadata[p].clone())
+                    .unwrap()
```
hawkw: turbo-nit: should this maybe be an `expect` so that, if the build fails, we get an error message that actually tells us what happened?

humility-bin/build.rs
```diff
@@ -32,19 +32,42 @@ pub enum Subcommand {{
...
+                    .features["probes"]
+                    .iter()
+                    .flat_map(|p| {
+                        p.strip_prefix("cmd-")
+                            .map(|p| format!("humility-cmd-{p}"))
+                    })
+                    .collect::<BTreeSet<_>>(),
+            )
```
hawkw: style nit, take it or leave it: i find throwing all of this in one giant tuple a little hard to follow; what do you think about doing something more like this:
```suggestion
            // These are everything that can potentially pull in libusb
            let banned_deps = ["probe-rs", "humility-probes-core", "rusb"]
                .into_iter()
                .collect();
                // We also look at which commands are enabled by the `probes`
                // feature on the `humility-bin` crate, then remove them.
            let disabled_cmds = metadata
                .workspace_members
                .iter()
                .find(|p| metadata[p].name == "humility-bin")
                .map(|p| metadata[p].clone())
                .unwrap()
                .features["probes"]
                .iter()
                .flat_map(|p| {
                    p.strip_prefix("cmd-")
                        .map(|p| format!("humility-cmd-{p}"))
                })
                .collect::<BTreeSet<_>>();
            (banned_deps, disabled_cmds)
```
this is a little more repetitive but i feel like it makes it slightly more obvious which expression corresponds to which variable?

labbott:
> I had to do a bit of extra wrangling to get the probeless build working; it now skips subcommands which are probe-only, both in the `Cargo.toml` (optional dependencies) and in codegen (ugly hacks).

I had hoped to be able to drop this after the probe-rs update but so long as we need to keep CMSIS DAP v1 around we will still need these hacks :(

humility-bin/build.rs
mkeeter: Yeah, the tuple kinda grew out of control; fixed

humility-bin/build.rs
mkeeter: doctor doctor it hurts when I rename `humility-bin`

humility-probes-core/src/probe_rs.rs
```diff
@@ -364,4 +340,28 @@ impl ProbeCore {
         self.halted = true;
         Ok(())
     }
+
+    pub fn vid_pid(&self) -> Option<(u16, u16)> {
+        Some((self.vendor_id, self.product_id))
+    }
```
labbott: I think we can drop the `Option` here now

labbott (approved):
minor fixup but LGTM otherwise

labbott:
I also think we can move `info` to `ProbeCore` but that seems like a follow up task

humility-cli/src/lib.rs
```diff
@@ -267,7 +262,16 @@ impl Cli {
...
         } else {
-            self.attach_probe(hubris).map(|b| Box::new(b) as Box<dyn Core>)
+            #[cfg(feature = "probes")]
+            {
+                self.attach_probe(hubris).map(|b| Box::new(b) as Box<dyn Core>)
+            }
+            #[cfg(not(feature = "probes"))]
+            {
+                Err(anyhow!(
+                    "probes feature is missing; ip address or dump is required"
```
hawkw: similarly, maybe:
```suggestion
                    "humility was compiled without the \"probes\" feature; \
                     an IP address or dump is required"
```

humility-cli/src/lib.rs
```diff
@@ -182,7 +182,16 @@ impl Cli {
...
-            self.attach_probe(hubris).map(|b| Box::new(b) as Box<dyn Core>)
+            #[cfg(feature = "probes")]
+            {
+                self.attach_probe(hubris).map(|b| Box::new(b) as Box<dyn Core>)
+            }
+            #[cfg(not(feature = "probes"))]
+            {
+                Err(anyhow!(
+                    "probes feature is missing; ip address is required"
+                ))
```
hawkw, line 194: turbo nit: maybe:
```suggestion
                Err(anyhow!(
                    "humility was compiled without the \"probes\" feature; \
                     an IP address is required"
                ))
```
or something? i think "probes feature is missing" is a little bit unclear for the user who may not be aware that this is specifically referring to a compile-time feature flag rather than some CLI arg you forgot to provide or something...

humility-probes-core/src/probe_rs.rs
hawkw: This is maybe a suggestion better saved for a follow-up change, since I'm sure it will break some stuff, but I really wish that this was 
```rust
pub struct VidPid {
    pub vid: u16,
    pub pid: u16,
}
```
rather than a tuple, so it's impossible to get them backwards (since we've had [at least one bug where we did that](https://github.com/oxidecomputer/humility/issues/479)).

hawkw (approved):
looks good to me overall, with some minor nits

humility-probes-core/src/probe_rs.rs
mkeeter: This is easy enough; I also added a `Display` implementation that prints `vid:pid` with the expected formatting.

humility-cli/src/lib.rs
mkeeter: Sure, done

humility-cli/src/lib.rs
mkeeter: Yup, done

humility-probes-core/src/probe_rs.rs
mkeeter: Good catch, done!

mkeeter:
> I also think we can move `info` to `ProbeCore` but that seems like a follow up task

That was easy enough to do that I just did it here as well

PR #702, opened 2026-06-17
Use `hubtools` function to get file metadata without extraction

mkeeter:
https://github.com/oxidecomputer/hubtools/pull/65 allows us to get file metadata without extracting the file.

When loading the archive, this can be significant (because the auxiliary flash image is chunky).  On my machine, this yields a 16% improvement in test suite runtime (`cargo test --release`): from 31.4s to 26.1s.

It also lets us hide the `zip` internals in `humility extract`.

### DIFF
--- a/cmd/extract/src/lib.rs
+++ b/cmd/extract/src/lib.rs
@@ -134,8 +134,7 @@ use clap::Parser;
 use humility_cli::{ExecutionContext, humility_cmd};
 use humility_log::info;
 use std::fs::File;
-use std::io::Cursor;
-use std::io::{self, Read, Write};
+use std::io::{self, Write};
 
 #[derive(Parser, Debug)]
 #[clap(name = "extract", about = env!("CARGO_PKG_DESCRIPTION"))]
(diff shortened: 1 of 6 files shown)


cmd/extract/src/lib.rs
```diff
@@ -157,32 +156,24 @@ fn extract(subargs: ExtractArgs, context: &mut ExecutionContext) -> Result<()> {
...
-        // file size without uncompressing.
-        let cursor = Cursor::new(archive.zip);
-        let mut archive = zip::ZipArchive::new(cursor)?;
-
         println!("{:>12} NAME", "SIZE");
 
-        for i in 0..archive.len() {
-            let file = archive.by_index(i)?;
-            println!("{:12} {}", file.size(), file.name());
+        for i in 0..archive.file_count()? {
```
jamesmunns, line 161: It seems like we might want a file iterator for this, not sure if this is something we do or if it's a dependency thing yet, will check.

edit: nope, it's a `zip::ZipArchive` thing, and doing it by index is actually how they recommend doing it [in their example](https://docs.rs/zip/latest/zip/read/struct.ZipArchive.html). Boo.

jamesmunns (approved):
Looks reasonable to me! One note, but it's a complaint about `zip`'s API, not anything you did.

evan-oxide (approved):

PR #703, opened 2026-06-18
Fix PMBus backend selector

mkeeter:
(There's a new big block comment that explains what's going on, so you may want to go read that)

Previously, we would always select the `IdolWorker` PMBus agent when network connected.  This is fine for the `udprpc` backend, where the HIF program is executed on the host computer and so the additional overhead of Idol IPC doesn't matter.  

However, the introduction of the `hiffy`+`net` backend means that we would (1) select the `IdolWorker` agent, because the system is network-connected then (2) use the `hiffy` backend, executing the larger programs on the SP.  This is not ideal!

This PR teaches the `pmbus` crate about the different `HiffyContext` backends, so that it can make smarter choices.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2934,6 +2934,8 @@ dependencies = [
  "humility-idol",
  "idol",
  "postcard",
+ "strum 0.28.0",
+ "strum_macros 0.28.0",
  "thiserror 2.0.18",
  "zerocopy",
 ]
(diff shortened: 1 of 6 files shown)


jamesmunns (approved):
Made a suggestion for a different way to organize this in https://github.com/oxidecomputer/humility/pull/704, feel free to take it or leave it :)

I think the current change/comment is reasonable to understand as-is, though my personal style pref is #704.

jamesmunns:
Just noting I'm still happy with the changes after the tweaks, still +1 to merge :)

PR #708, opened 2026-06-29
Replace all in-memory cores with `InMemoryCore`

mkeeter:
This PR unifies the disparate `DumpAgentCore`, `DryCore`*, `ArchiveCore`, and `DumpCore` into a single `InMemoryCore` type.  Once this PR lands, we will be down to the irreducible set of `ProbeCore`, `InMemoryCore`, and `NetCore`.

Each of these cores stored data in memory, all in slightly different ways, and with different degrees of rigor – for example, some of them could not handle reads across multiple contiguous memory spans.  The new `InMemoryCore` uses the newly-published [`datamap`](https://docs.rs/datamap/latest/datamap/) crate, which is a rigorous wrapper around a `BTreeMap<u32, Vec<u8>>` (non-overlapping, efficient reads, etc).

This change adds an extra `memcpy` in a few places.  Previously, some of the cores stored data in a single linear array and then indexed into it with a `BTreeMap<u32, (usize, usize)>` (`address → (offset, size)`), instead of owning the data in the map.  I think the consistency of a single implementation is worth it; this is not going to be a limiting factor in typical Humility usage.

*the `DryCore` has been mostly forgotten, but is a way to store dumps without a matching archive then rehydrate them later


humility-core/src/mem.rs
```diff
@@ -2,44 +2,130 @@
...
-use std::collections::{BTreeMap, HashMap};
-use std::fs::File;
-use std::io::Read;
-
-pub struct DumpCore {
-    contents: Vec<u8>,
-    regions: BTreeMap<u32, (u32, usize)>,
+use std::{collections::HashMap, fs::File, io::Read};
+
+/// A core stored entirely in host memory
```
labbott: nit: 'host memory' is slightly ambiguous (is this the same host as `host-panic`?) Maybe "A core stored in the humility host memory"

humility-core/src/mem.rs
```diff
@@ -2,44 +2,130 @@
...
+    ///
+    /// When reading, this is checked before `mem`
+    flash: Option<HubrisDataMap>,
+
+    /// Memory contents, loaded from other sources
+    ///
+    /// Note that when a dump is loaded, `mem` includes both flash and RAM
+    /// (instead of storing data separately in `flash` and `mem`).  This only
+    /// affects error messages.
+    mem: Option<HubrisDataMap>,
```
labbott, line 30: Is this better modeled as an enum vs `Option` explosion?

humility-core/src/mem.rs
```diff
@@ -2,44 +2,130 @@
...
 }
 
-impl DumpCore {
-    pub(crate) fn new(dump: &str) -> Result<DumpCore> {
-        let mut file = File::open(dump)?;
-        let mut regions = BTreeMap::new();
+impl Core for InMemoryCore {
+    fn is_dump(&self) -> bool {
+        true
+    }
```
labbott: rename to `is_memory_core`

humility-core/src/mem.rs
```diff
@@ -2,44 +2,130 @@
...
+        Ok(())
+    }
+
+    fn run(&mut self) -> Result<()> {
+        Ok(())
+    }
+}
+
+impl InMemoryCore {
+    pub fn from_dump(dump: &str) -> Result<Self> {
```
labbott: changing this to `PathBuf` would be nice but I can do that in a follow up pass

humility-core/src/mem.rs
```diff
@@ -2,44 +2,130 @@
...
+    }
+
+    fn is_archive(&self) -> bool {
+        // An archive dump only has the flash data available
+        self.flash.is_some() && self.mem.is_none() && self.registers.is_none()
+    }
+
+    fn is_net(&self) -> bool {
+        false
+    }
```
labbott, line 48: All these `is_X` functions make me wonder if we're missing an enum for a core type somewhere (or maybe it's not that bad?)

labbott (commented):
mostly nits

humility-core/src/mem.rs
mkeeter: Agreed – later down the line, we may want to turn the `dyn Core` on its head, replacing it with an `enum Core` with our three core types (net, probe, in-memory).  This would be better from a type-safety perspective, because stuff like `send` would only exist on the `NetCore`, but would be a bunch of work.

humility-core/src/mem.rs
mkeeter: Gonna switch to

> A core stored entirely in the memory of the computer running Humility

humility-core/src/mem.rs
mkeeter: I don't think so – we end up using many possible combinations, so it would be awkward to have `InMemoryCore::FlashAndMemButNoRegisters`, etc.

Two things that we _could_ change if folks feel strongly:

- We could merge `mem` and `flash`, since they should be non-overlapping.  Some of the previous cores kept them separate to give finer-grained error messages, but that also could (theoretically) lead to confusion if the same data range is present in both; combining them would bail with an error when inserting overlapping data.  On the other hand, there are a few places which use `Core::is_archive`, so we'd have to revisit those call sites.
- The fields also don't _need_ to be `Options`, since an empty map is equivalent; again, the `Option` is for better error messages, to distinguish "we don't have any flash data" versus "we don't have _that_ flash data".  I guess we could treat empty maps as a special case, though, which would make the `Option` unnecessary.

humility-core/src/mem.rs
mkeeter: Done, this wasn't too bad but had a few ripple effects.

labbott (approved):

mkeeter:
I tested this with `humility dump` using both a probe and network connection, and `humility tasks` over a network connection.  All of the dumps look fine.  Unit tests also exercise both `DumpCore` and `ArchiveCore`, so I'm not too worried here.

hubris.core.2
evan-oxide, line 1: Was this added by accident?

mkeeter: nothing to see here

PR #710, opened 2026-07-02
Rework `host` commands

jamesmunns:
This aims to support https://github.com/oxidecomputer/hubris/pull/2518, but also remedy https://github.com/oxidecomputer/hubris/issues/2586.

### DIFF
--- a/cmd/host/src/lib.rs
+++ b/cmd/host/src/lib.rs
@@ -163,6 +163,14 @@ static HOST_STATE_BUF_NAME_2: &str =
 
 static PACKRAT_BUF_NAME: &str = "task_packrat::main::BUFS";
 
+// packrat field names
+const PACKRAT_LAST_PANIC_PAYLOAD: &str =
+    "cell.value.host_info.host_panic_payload";
+const PACKRAT_LAST_PANIC_STATE: &str = "cell.value.host_info.host_panic_state";
+const PACKRAT_BOOT_FAIL_PAYLOAD: &str =
+    "cell.value.host_info.host_panic_payload";
+const PACKRAT_BOOT_FAIL_STATE: &str = "cell.value.host_info.host_panic_state";
+
 /// Mirror type of the internal buf struct in `host_sp_comms`. Must be kept in
 /// (partial) sync with that structure (fields that are present need to match,
 /// other fields can be ignored).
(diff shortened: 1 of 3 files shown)


cmd/host/src/lib.rs
```diff
@@ -229,48 +235,174 @@ fn print_escaped_ascii(mut bytes: &[u8]) {
...
-    let d = read_uqvar(hubris, core, SEPARATE_LAST_HOST_PANIC_NAME)?;
-    if let Some(d) = d {
-        return print_panic(d, log);
+    // Work through the different places "last panic" info could be hiding
+    let sources: [fn(&HubrisArchive, &mut dyn Core) -> _; _] = [
+        host_last_panic_packrat,
+        host_last_panic_spcomms_old,
+        |h, c| host_last_panic_spcomms_new(h, c, HOST_STATE_BUF_NAME_1),
+        |h, c| host_last_panic_spcomms_new(h, c, HOST_STATE_BUF_NAME_2),
+    ];
```
labbott, line 375: Is this in order of newest to oldest?

cmd/host/src/lib.rs
```diff
@@ -229,48 +235,174 @@ fn print_escaped_ascii(mut bytes: &[u8]) {
...
+) -> Result<Option<Vec<u8>>> {
+    let buf = read_qualified_state_buf(hubris, core, base_buf)?;
+    let maybe_bf = buf.map(|b| b.last_boot_fail);
+    Ok(maybe_bf)
+}
+
+fn host_boot_fail_packrat(
+    hubris: &HubrisArchive,
+    core: &mut dyn Core,
+) -> Result<Option<Vec<u8>>> {
```
labbott: I keep wishing this could share more code with `host_last_panic` but I think they are just enough different that would be annoying

jamesmunns: I can take a swing at it.

cmd/host/src/lib.rs
jamesmunns: The old code did basically:

* `host_last_panic_spcomms_old`
* `host_last_panic_spcomms_new(h, c, HOST_STATE_BUF_NAME_1)`

I made the executive decision to try packrat first, then the new-new spcomms last. Open to re-order this, it shouldn't be order dependent.

cmd/host/src/lib.rs
jamesmunns: Done in https://github.com/oxidecomputer/humility/pull/710/commits/bf0ecf22e765a07b651ef4d13716e1b78aba15c9

humility-bin/tests/cmd/host-panic/host-panic.control_plane_agent.overflow.0.stderr
```diff
@@ -1,2 +1,2 @@
 humility: attached to dump
-humility host failed: read of 24964 bytes failed: could not find addr 0x24021ff4 in memory
+humility host failed: read of 8264 bytes failed: could not find addr 0x24004380 in memory
```
labbott, line 2: I'm curious why this changed. I'm guessing this was the order of how things were tried and what the final one was?

labbott (approved):
LGTM. As a follow on it would be good to grab a dump with a panic/boot error in the new format and have that in CI.

(I don't remember when I started doing it but I pruned a bunch of very very old hubris cores that were adding complexity. Adding more cores that are actually in use would be good to do!)

humility-bin/tests/cmd/host-panic/host-panic.control_plane_agent.overflow.0.stderr
jamesmunns: I actually don't understand this change! Maybe we could look at this together during the watercooler or some time today?

jamesmunns: Okay @labbott I think this is just it looking for the new, smaller, but also not present data in an old image. The old code would pull the ~24k big blob, whereas the new code is going to look for the newer and smaller data first. I'm going to go ahead and merge this.

PR #711, opened 2026-07-07
Attempt a basic PC range check as part of `hubris.validate`

labbott:
`validate` on a hubris archive will verify the `image_id` present in an archive matches what's in flash. This has an unfortunate failure mode for the RoT where we partition the flash manually. It's possible to have the image ID check succeed on the archive but actually be running the opposite image. This causes all kinds of confusion. Add a basic spot check that the PC is in at least one module range. This could cause potential failures if the PC is actually running in a range that does not correspond to a module. Based on other work, the chances of this happening are very limited and should be transient.


labbott:
```
labbott@labbott-desktop2:~/humility$ ./target/debug/humility -a ~/hubris/target/oxide-rot-1-selfsigned/dist/a/build-oxide-rot-1-selfsigned-image-a.zip tasks
humility: attached via 1fc9:143:Some("U2VLC0GBANWF1") MCU-LINK (r0FB) CMSIS-DAP V3.172
system time = 200385
ID TASK                       GEN PRI STATE    
 0 jefe                         0   0 recv, notif: fault timer(T+15)
 1 hiffy                        0   6 notif: timer(T+18)
 2 idle                         0   7 RUNNING
 3 update_server                0   3 recv
 4 syscon_driver                0   2 recv
 5 gpio_driver                  0   3 recv
 6 sprot                        0   6 notif: spi-irq(irq59)
 7 swd                          0   4 recv, notif: timer sp_reset-irq(irq4) jtag_detect-irq(irq5)
 8 dumper                       0   5 recv
 9 attest                       0   3 recv
labbott@labbott-desktop2:~/humility$ ./target/debug/humility -a ~/hubris/target/oxide-rot-1-selfsigned/dist/b/build-oxide-rot-1-selfsigned-image-b.zip tasks
humility: attached via 1fc9:143:Some("U2VLC0GBANWF1") MCU-LINK (r0FB) CMSIS-DAP V3.172
humility tasks failed: PC at 0x1ef34 is not part of any module. This is likely an incorrect archive.
```

humility-core/src/hubris.rs
```diff
@@ -2103,11 +2103,19 @@ impl HubrisArchive {
...
+        if self.task_dump().is_some() {
             return Ok(());
+        } else {
+            core.halt()?;
+            if let Ok(pc) = core.read_reg(ARMRegister::PC) {
+                if self.instr_mod(pc).is_none() {
+                    bail!("PC at 0x{pc:x} is not part of any module. This is \
+                            likely an incorrect archive.");
+                }
+            }
```
hawkw: i believe clippy would like you to have written something more like:
```suggestion
            if let Ok(pc) = core.read_reg(ARMRegister::PC)
                && self.instr_mod(pc).is_none() {
                    bail!("PC at 0x{pc:x} is not part of any module. \
                        This is likely an incorrect archive.");
                }
```

humility-core/src/hubris.rs
```diff
@@ -2103,11 +2103,19 @@ impl HubrisArchive {
...
         }
 
-        if criteria == HubrisValidate::ArchiveMatch {
+        if self.task_dump().is_some() {
             return Ok(());
+        } else {
+            core.halt()?;
+            if let Ok(pc) = core.read_reg(ARMRegister::PC) {
+                if self.instr_mod(pc).is_none() {
+                    bail!("PC at 0x{pc:x} is not part of any module. This is \
```
hawkw: unimportant turbo nitpick: if we write
```suggestion
                    bail!("PC at {pc:#x} is not part of any module. This is \
```
that tells the formatter to emit the `0x` prefix itself. which...changes absolutely nothing here so it doesn't matter...

hawkw (approved):

humility-core/src/hubris.rs
evan-oxide: I think the way this interacts with flashing is not ideal. If you try to re-flash an A image that's already been flashed to the A slot while the RoT is running a B image:
- the old behavior was to fail with "archive appears to be already flashed on attached device; use -F ("--force") to force re-flash"
- this PR's behavior is to re-flash the A image
- the ideal behavior is probably to warn the user that they probably meant to flash the other image, then fail with the  "archive appears to be already flashed" error.

labbott: yeah you're right that's confusing behavior. I think that would point to needing to fix validate to return a more specific error

evan-oxide: Or don't use `validate` for the pre-flash check. In my PR, I considered a bunch of different return types then gave up trying to make the same `validate` function work for flashing and for both of the `HubrisValidate` options.

humility-core/src/hubris.rs
```diff
@@ -2103,11 +2103,19 @@ impl HubrisArchive {
                 );
         }
 
-        if criteria == HubrisValidate::ArchiveMatch {
+        if self.task_dump().is_some() {
             return Ok(());
+        } else {
+            core.halt()?;
```
evan-oxide, line 2097: Should we do `core.run()?` again after checking the PC?

labbott: Absolutely yes this got lost in some refactoring 😬

humility-core/src/hubris.rs
```diff
@@ -2170,6 +2163,34 @@ impl HubrisArchive {
...
+            format!("failed to read image ID at 0x{:x}; board mismatch?", addr)
+        })?;
+        Ok(id)
+    }
+
+    /// The core must be halted before this is called
+    pub fn is_pc_within_archive(
+        &self,
+        core: &mut dyn crate::core::Core
+    ) -> Result<(bool, u32)> {
```
labbott: I don't quite follow this return type. Can you explain it with a comment?

humility-flash/src/lib.rs
```diff
@@ -92,11 +93,30 @@ pub fn get_image_state(
...
+        Ok(id) if id != hubris.image_id() => {
+            return Ok(ImageStateResult::DoesNotMatch(
+                ImageStateMismatch::FlashArchiveMismatch(anyhow!(
+                    "image ID in archive ({:x?}) does not equal \
+                         image ID of target ({:x?})",
+                    hubris.image_id(),
+                    id,
+                )),
+            ));
+        }
```
labbott, line 131: I think these should be two separate error types: one to represent an error reading the image ID and a separate one to indicate an actual mismatch. This also means we can avoid having to create an extra error message with `anyhow` and have all the information in the `ImageStateMismatch` enum.

labbott (commented):
Thanks for doing this

humility-flash/src/lib.rs
evan-oxide: Done. That's a nice side effect of not calling `validate` anymore.

humility-core/src/hubris.rs
evan-oxide: Done. Yeah it's kinda weird. I wanted a helper function for doing the check, but also didn't want to have to remove the PC value from your error message. It will be useful for debugging when this error gets triggered by some other strange situation someday.

lzrd:
Test this corner case: corrupt/erase both Hubris images, reset the RoT. Use humility to dump the registers. It should be in a bootleby loop. (check bootleby source).
The other harder to hit case is in pre-kernel where ROM routines are being used to check image signatures.
There should be an override flag "--i-know-what-i-am-doing" or some such.

humility-core/src/hubris.rs
labbott: No need to change it now but it's also worth considering something like

```
enum PcResult {
     InArchive,
     NotInArchive { pc: u32 },
}
```

humility-core/src/hubris.rs
```diff
@@ -2080,34 +2080,27 @@ impl HubrisArchive {
...
             bail!(
                     "image ID in archive ({:x?}) does not equal \
                     ID at 0x{:x} ({:x?})",
                     self.imageid.1, self.imageid.0, id,
                 );
         }
 
-        if criteria == HubrisValidate::ArchiveMatch {
+        if self.task_dump().is_some() {
             return Ok(());
```
evan-oxide, line 2095: @labbott What does it mean for an archive to be "a dump from a single task"? I'm wondering if it's intentional/correct that if you pass in `HubrisValidate::Booted` in that case, it returns early here and skips the boot check.

jamesmunns: I think it's possible to take a dump from only a single task's memory range (e.g. a "single task dump"), but not an "entire system" dump. Laura can confirm though :)

labbott: James is correct. We have the notion of single task dumps vs a wholes system dump. You can run many of the same commands like e.g. `humility tasks` and `humility ringbuf` and things should work as expected. I don't think we capture the boot information in a single task dump which is why it gets skipped.

evan-oxide: Thanks, I added a comment about it.

evan-oxide:
> Test this corner case: corrupt/erase both Hubris images, reset the RoT. Use humility to dump the registers. It should be in a bootleby loop. (check bootleby source). The other harder to hit case is in pre-kernel where ROM routines are being used to check image signatures. There should be an override flag "--i-know-what-i-am-doing" or some such.

I corrupted both the A and B images so that it would stay in the bootloader. `humility registers` succeeded since it doesn't check for an archive mismatch anyways. `humility tasks` failed as expected.

It's looking like we won't need a `--skip-pc-check` flag, but let's see what folks think of the changes to the test cases.

humility-bin/tests/cmd/counters-ipc-filtered/counters-ipc-filtered.flash-ram-mismatch.0.stderr
evan-oxide, line 1: I split up the test-case fixes into 3 commits. The first is straightforward changes like this file, where just the error message changed. But the other two are tests where the command used to succeed (or appear to succeed?) but now fails the PC check. Does anyone disagree with any of these changes and think that we should allow someone to, for example, run `humility readvar` while in the bootloader? If so, that might be a reason to add a `--skip-pc-check` flag.

humility-core/src/hubris.rs
evan-oxide: I do like that, changed.

humility-bin/tests/cmd/counters-ipc-filtered/counters-ipc-filtered.flash-ram-mismatch.0.stderr
labbott: I'm inclined to say that the change is fine and that it's okay to give failures if we're in the bootloader. For the SP in particular we should never expect to end up in the bootloader so any attempt to read things from hubris RAM is likely to be questionable. 

The RoT may be a more interesting case because we do expect to briefly go through a non-hubris bootloader but that code is designed to be so minimal I think rejecting it is okay.  Really I think if future us actually finds a situation where something like `--skip-pc-check` is valuable/solves a real problem I think we can add it then.

humility-flash/src/lib.rs
```diff
@@ -92,11 +104,32 @@ pub fn get_image_state(
...
 
+    if let Ok(PcResult::NotInArchive { pc }) = hubris.is_pc_within_archive(core)
+    {
+        warn!(
+            log,
+            "PC at 0x{pc:x} is not part of any module. Maybe you're \
+                    flashing an A image while a B image is running, or vice \
+                    versa?"
+        );
+    }
```
labbott, line 115: we drop the `Err` case here. That should only happen if we can't read the PC and we're probably going to have other errors that would get caught below so I think it's fine.

labbott (commented):
LGTM. This is actually originally my PR so I can't actually approve it!

humility-flash/src/lib.rs
evan-oxide: Yeah my thought was for it to just log nothing unless it's sure the PC location is bad. Would you prefer to have a "failed to check PC location" warning too?

Issue #709, opened 2026-07-01
Warn when using the wrong RoT A/B image


evan-oxide:
It's easy to accidentally flash the RoT with an A image when its boot preference is set to B, or vice versa. This leads to great confusion when your new image does not run after the RoT resets. Similarly, other commands like `humility tasks` can end up using an archive that doesn't match what's actually running. It would be helpful for humility to print a warning message if it detects these situations. This could probably be done as part of the ID check, by checking if the running PC matches the range in the archive.

Someday we could turn the warning into a hard error or have humility automatically detect which archive to use, but let's start with just the warning.

evan-oxide:
Closed by https://github.com/oxidecomputer/humility/pull/711

PR #712, opened 2026-07-07
Use BOOTED_IMAGE to warn when using wrong A/B image

evan-oxide:
The companion hubris PR is https://github.com/oxidecomputer/hubris/pull/2591. 

Both this and Laura's PR https://github.com/oxidecomputer/humility/pull/711 are trying to address issue https://github.com/oxidecomputer/humility/issues/709 - we can choose to merge one or combine both approaches in some way.

This PR's approach is to make the hubris kernel record the image ID and name in the `BOOTED_IMAGE` variable so that humility can tell which A/B image is actually running on the RoT. Humility now uses that image ID to check whether an archive matches what's running on the target, and uses the image name to warn when flashing an image to the inactive slot.

**New behavior:**

- If you run `humility tasks` etc with an A archive and a target running a B image, it will:
  - fail with an ID mismatch error, if it can read BOOTED_IMAGE
  - log a "can't read BOOTED_IMAGE" warning and fall back to reading the ID from FLASH, if it can't

- If you're flashing an A image and a target running a B image, it will:
  - log a "you're probably doing this by accident" warning, if it can read BOOTED_IMAGE
  - be silent, if it can't

Note that it never blocks you from flashing and doesn't change the logic of whether an archive needs to be re-flashed.

**Tradeoffs**

- This approach is very unlikely to have false-positive errors/warnings, but can easily have false negatives if the address of BOOTED_IMAGE drifts between versions (or the image is too old, or the kernel never booted). Hopefully it will work often enough to be useful. Or should we try to make the linker give BOOTED_IMAGE a fixed address?
- It uses an extra 24 bytes of RAM.

**Testing:**

Lightly tested on a grapefruit's RoT and SP. If folks like this approach then I'll keep testing more combos and edge cases. Also CI tests will fail because I haven't yet added the line `humility: can't detect booted image, reading image ID from FLASH (BOOTED_IMAGE not in archive's symbol table)` to 415 unit tests.

### DIFF
--- a/humility-core/src/hubris.rs
+++ b/humility-core/src/hubris.rs
@@ -2071,36 +2071,39 @@ impl HubrisArchive {
         &self,
         core: &mut dyn crate::core::Core,
         criteria: HubrisValidate,
+        log: &Logger,
     ) -> Result<()> {
-        let ntasks = self.ntasks();
         if core.is_net() || core.is_archive() {
             return Ok(());
         }
 
-        // To validate that what we're running on the target matches what
-        // we have in the archive, we are going to check the image ID, an
-        // identifer created for this purpose.
-        let addr = self.imageid.0;
-        let nbytes = self.imageid.1.len();
-        assert!(nbytes > 0);
+        // To validate that what we're running on the target matches what we
+        // have in the archive, we are going to check the image ID, an identifer
+        // created for this purpose. First try to read the ID of the image that
+        // actually booted, as recorded by the kernel. If that fails then fall
+        // back to reading the ID from FLASH, which can be misleading on targets
+        // with A/B images. For example, if the archive is an A image, then
+        // `read_image_id_from_flash` will return the ID of whichever A image
+        // was last flashed regardless of whether the target is currently
+        // running an A or B image.
+        let id = match self.booted_image(core) {
+            Ok((id, _)) => id,
+            Err(err) => {
+                info!(
+                    log,
+                    "can't detect booted image, reading image ID from FLASH ({})",
+                    err
+                );
+                self.read_image_id_from_flash(core)?
+            }
+        };
 
-        let mut id = vec![0; nbytes];
-        core.read_8(addr, &mut id[0..nbytes]).with_context(|| {
-            format!("failed to read image ID at 0x{:x}; board mismatch?", addr)
-        })?;
-
-        let deltas = id
-            .iter()
-            .zip(self.imageid.1.iter())
-            .filter(|&(lhs, rhs)| lhs != rhs)
-            .count();
-
-        if deltas > 0 || id.len() != self.imageid.1.len() {
+        if id != self.imageid.1 {
             bail!(
                     "image ID in archive ({:x?}) does not equal \
-                    ID at 0x{:x} ({:x?})",
-                    self.imageid.1, self.imageid.0, id,
-                );
+                    image ID of target ({:x?})",
+                    self.imageid.1, id,
+            );
         }
 
         if criteria == HubrisValidate::ArchiveMatch {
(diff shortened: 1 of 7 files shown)


humility-core/src/hubris.rs
```diff
@@ -2071,36 +2071,39 @@ impl HubrisArchive {
...
+                    log,
+                    "can't detect booted image, reading image ID from FLASH ({})",
+                    err
+                );
+                self.read_image_id_from_flash(core)?
+            }
+        };
 
-        if deltas > 0 || id.len() != self.imageid.1.len() {
+        if id != self.imageid.1 {
```
evan-oxide, line 2101: I can't think of any reason why the original code would need to do that manual comparison instead of just using `!=`. Hope I'm not missing something.

humility-core/src/hubris.rs
evan-oxide, line 2122: We could start using `BOOTED_IMAGE` for this too. But maybe it's better to use `CURRENT_TASK_PTR` because that's set at the end of kernel startup, while `BOOTED_IMAGE` is set at the beginning of kernel startup.

humility-core/src/hubris.rs
```diff
@@ -2071,36 +2071,39 @@ impl HubrisArchive {
...
+        // running an A or B image.
+        let id = match self.booted_image(core) {
+            Ok((id, _)) => id,
+            Err(err) => {
+                info!(
+                    log,
+                    "can't detect booted image, reading image ID from FLASH ({})",
+                    err
+                );
+                self.read_image_id_from_flash(core)?
```
evan-oxide, line 2097: It might be worth checking the PC location here as part of the fallback when we can't read BOOTED_IMAGE. In addition to requiring the flash ID to match, we could either fail or warn if the PC suggests this is the wrong image slot.

labbott:
I think this is more complete than just the PC check which is a bit hacky. 

It also occurs to me this is reimplementing the bootleby decision log for the RoT https://github.com/oxidecomputer/bootleby/blob/bdf56dd950b934360df596ed5b2d8b8813c92168/src/bin/bootleby.rs#L201-L207 , https://github.com/oxidecomputer/bootleby/blob/bdf56dd950b934360df596ed5b2d8b8813c92168/memory.x#L39 (see https://github.com/oxidecomputer/hubris/issues/2066 as well). Didn't think of this until I finished reviewing this work!

We don't have a problem with the SP so part of me says that we should just read the decision log in humility. Do you see any value in having the same information for the SP even though everything is just "default"?

evan-oxide:
> We don't have a problem with the SP so part of me says that we should just read the decision log in humility. Do you see any value in having the same information for the SP even though everything is just "default"?

Oh cool, it would be nice to reuse an existing feature. Questions / concerns:

1) Is it valuable for humility's validation process to be target agnostic and not have to check a bootleby-specific thing when it sees that the target is an RoT? Or is that not a big deal?

2) It looks like the log doesn't include which slot bootleby actually picked, so humility would need to duplicate the override logic, which is not too bad. I think it does mean that if bootleby added a new higher-priority override to the log then humility could guess the wrong slot - oh wait there's a version byte that protects against that.

3) Humility needs to know how bootleby's slot boolean maps to an actual archive. It could check `image_name == "a"`, but hopefully there's a nicer way that doesn't require hardcoding a name string.

But if none of those seem like a big deal, then I think it would be fine to use the bootleby log instead of `BOOTED_IMAGE`. And the fact that it has a fixed address in RAM is great.

evan-oxide:
I think we're going to go with the program counter approach instead. Work on that is happening over here: [humility#711 Attempt a basic PC range check as part of `hubris.validate`](https://github.com/oxidecomputer/humility/pull/711)

PR #715, opened 2026-07-14
Bump probe-rs to pick up humility#705 fix

labbott:
This adds some retries on the USB reading.


labbott:
Fixes https://github.com/oxidecomputer/humility/issues/705

PR #718, opened 2026-07-22
Bump our probe-rs fork

labbott:
This brings in a newer hidapi-rs and serialport. There were compilation problems for some cases when using the older version.


jamesmunns (approved):
Looks good to me!

PR #720, opened 2026-08-04
Avoid creating transparent wrappers instead of peeling them off

jamesmunns:
Instead of creating wrappers and then peeling them off, we can instead just not create them in the first place.

Removed reflect tests since they now would require a HubrisArchive, but new core tests still pass the same.


jamesmunns:
@lzrd do you want me to close this, if you're not interested?

lzrd:
Always nice to see more lines deleted than added.

lzrd (approved):
LGTM

PR #721, opened 2026-08-06
Set a default probe speed

labbott:
Newer probe-rs changed the `read_8` backend to actually use 8-bit reads instead of just relying on `read_32`. This introduces a noticable performance regression. Using a faster probe speed improves the performance to what we expect. With `read_32` the probe itself was not the bottleneck whereas repeated reads with `read_8` require a probe to go at a faster speed.


jamesmunns (approved):

adamlouis:
Empirically, the changes here are fast enough to avoid full duty cycle fans on `argon` which was our main concern.

The runtime of `humility flash --verify` is reduced from 55s to 31s - though is still above the 15s in `0.12.6`.

`0.12.6` - `14.984s`
`0.13.4` - `55.284s`
`probe_go_fast` - `31.345s`

See - https://github.com/oxidecomputer/facade/issues/727#issuecomment-5209160547

adamlouis (approved):

labbott:
Verified that 8mhz probe speed works on OxLink, old MCULink and STLink so it should be fine for most use cases.

PR #725, opened 2026-08-19
Fix halting behavior preventing `humility probe` from working

labbott:
@adamlouis flagged this in a facade change that we were getting errors from `humility probe` "The core has to be halted for the operation, but was not." . This introduces two change: a way to get the previous CPU state from `run` and `halt` and a fix in `validate`


evan-oxide (approved):
LGTM, sorry about that

humility-core/src/core.rs
```diff
@@ -11,6 +11,14 @@ use humility_log::{Logger, info};
 use std::time::Duration;
 use thiserror::Error;
 
+/// State of the CPU before the halt/run operation
+pub enum PreviousCpuState {
```
jamesmunns, line 15: Should we mark this `#[must_use]` to catch places where we don't check this but should?

labbott: I tried this and I think it causes more problems than it's solving. Most other places in the code are very deliberate about wanting halt/run so the logic seems to result in `let _ = core.run()?`. Maybe this is a sign we need a `fn run_while_halted(&self, fn ...)` for parts of the code? That also seems like follow up work.

PR #726, opened 2026-08-21
Bump humility version for humility#725

labbott:


labbott:
Should have thought about that before I hit merge on humility#725

jamesmunns (approved):

PR #728, opened 2026-09-03
Rename packrat field names

### LINKED ISSUE #727: `humility host last-panic` fails with `member missing from struct: last_boot_fail`
```
$ humility --version
humility 0.13.7
```

```
build-cosmo-b-lab-image-default-v1.79.0.zip
```

```
$ pfexec humility host last-panic
humility: attached to xxxx:yyyy:ZZZZZZZZZ via CMSIS-DAP V1
humility host failed: loading reflected value of type humility_doppel::ClaimOnceCell<humility_cmd_host::HostStateBuf>

Caused by:
    member missing from struct: last_boot_fail
```

jamesmunns:
Closes #727 

In https://github.com/oxidecomputer/hubris/pull/2518/changes/8011935be039537ab6fa6023e100ead3bb431fe9 I changed the name of these fields after the humility PR landed, and didn't go back to change them.

This addresses that. We don't need to support the old names as that never landed in the main hubris tree.


labbott (approved):

cmd/host/src/lib.rs
```diff
@@ -164,12 +164,10 @@ static HOST_STATE_BUF_NAME_2: &str =
...
-const PACKRAT_LAST_PANIC_PAYLOAD: &str =
-    "cell.value.host_info.host_panic_payload";
-const PACKRAT_LAST_PANIC_STATE: &str = "cell.value.host_info.host_panic_state";
-const PACKRAT_BOOT_FAIL_PAYLOAD: &str =
-    "cell.value.host_info.host_panic_payload";
-const PACKRAT_BOOT_FAIL_STATE: &str = "cell.value.host_info.host_panic_state";
+const PACKRAT_LAST_PANIC_PAYLOAD: &str = "cell.value.host_info.panic_payload";
+const PACKRAT_LAST_PANIC_STATE: &str = "cell.value.host_info.panic_state";
+const PACKRAT_BOOT_FAIL_PAYLOAD: &str = "cell.value.host_info.panic_payload";
+const PACKRAT_BOOT_FAIL_STATE: &str = "cell.value.host_info.panic_state";
```
adamlouis, line 170: @jamesmunns are there tests we can add to catch issues like these moving forward?

jamesmunns: Yes, I'll add an issue to add this to our regression test suite. We have tests for older versions, but I didn't add one as we hadn't landed the hubris change yet when the humility PR did (which also fixed some other issues).

jamesmunns: Opened https://github.com/oxidecomputer/humility/issues/729

adamlouis:
@jamesmunns This works on the affected systems - thanks for the fix!

